PR #2, opened 2022-07-27
New geospatial

bamader:
This changeset introduces the new SDK geospatial functionality for both raw data and fhir formatted data in accordance with the SDK PRD and SDD. GeocodeResult has been defined as a dataclass, and an abstract base class GeocodeClient has been created for each of raw and FHIR-formatted data, from which vendor-specific implementations will derive. Users are now able to geocode from strings and dictionaries (in the case of raw data) and from resources and bundles (in the case of FHIR data). Unit tests are written for every function in all new modules.

Notes:
* because `fhir` was a taken name in the directory namespace, I've temporarily called the fhir directory `fhir_wrappers`; this can change once we move everything over to the new repo
* I know the new repo is created but wasn't sure if we were ready to begin pushing code to it yet (since there's so many dependencies and things to set up in the new repo), hence this PR being opened here


src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
```diff
@@ -0,0 +1,99 @@
...
+
+        :param resource: The resource whose addresses should be geocoded
+        :param overwrite: Whether to save the geocoding information over
+          the raw data, or to create a copy of the given data and write
+          over that instead. Defaults to True (write over given data).
+        """
+        if not overwrite:
+            resource = copy.deepcopy(resource)
+
+        rtype = resource.get("resourceType", "")
```
KennethSkylight, line 53: Could we change `rtype` to just `resource_type`? Makes it more readable

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
```diff
@@ -0,0 +1,99 @@
...
+            ClientBuilder(creds)
+            .with_licenses(["us-standard-cloud"])
+            .build_us_street_api_client()
+        )
+
+    @property
+    def client(self) -> us_street.Client:
+        return self.__client
+
+    def geocode_resource(self, resource: dict, overwrite=True) -> dict:
```
KennethSkylight, line 38: Not a request for change just a small rant. I get annoyed when we have to name functions with words that are verbs and nouns at the same time lol.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
```diff
@@ -0,0 +1,99 @@
...
+          over that instead. Defaults to True (write over given data).
+        """
+        if not overwrite:
+            resource = copy.deepcopy(resource)
+
+        rtype = resource.get("resourceType", "")
+        if rtype == "Patient":
+            self._geocode_patient_resource(resource)
+
+        return resource
```
KennethSkylight, line 57: I think we should include what a function returns in the docstrings. Data type and what is it. Thoughts?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class FhirGeocodeClient(ABC):
+    """
+    A basic abstract class representing a vendor-agnostic geocoder client
+    designed as a wrapper to process FHIR-based data (including resources
+    and bundles). Requires implementing classes to define methods to
+    geocode from both bundles and resources. Treats the underlying client
+    in vendor-specific implementations as a special internal property
+    which can be accessed using .client().
```
spencer-skylight: Rather than highlighing the use of the client (which is mostly meant to be internal), maybe replace this with:
```
Callers should use the interface methods (e.g. geocode_resource) to interact with the vendor-specific client property.
```

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    @property
+    @abstractmethod
+    def client(self):
```
spencer-skylight: I think we should comment each abstract class, not so much for SDK users, but for developer users who want to implement the methods and want to know what they should do.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
spencer-skylight: I agree that we should document what is returned. Currently our wider solution doesn't explicitly call out return value in the comments. I think this decision was made deliberately, with the expectation that the return value should be specified in the comments, and the return type is in the method signature.

I'd propose being more explicit in the comments about this returning a dict with the `resource` param enhanced with lat and long.  Maybe also discuss with the wider team why we are doing things this way currently and get a consensus around explicitly documenting return value or not.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
```diff
@@ -0,0 +1,64 @@
...
+    @abstractmethod
+    def geocode_resource(self):
+        pass
+
+    @abstractmethod
+    def geocode_bundle(self):
+        pass
+
+
+def _get_one_line_address(address: dict) -> str:
```
spencer-skylight: I'm a little torn on how to handle these, but what do you think about adding these to the abstract class definition itself?  I think that would allow extending classes to call them directly without needing to import an _-function explicitly.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
```diff
@@ -0,0 +1,64 @@
+from abc import ABC, abstractmethod
```
spencer-skylight: I'm thinking we could move this to the `__init__.py` class, which would change import from:
```
import phdi.fhir.geospatial.geospatial
import phdi.fhir.geospatial.smarty
```
to
```
import phdi.fhir.geospatial
import phdi.fhir.geospatial.smarty
```

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
bamader: Sure thing!

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
bamader: LOL I totally get it, I'm a language nerd myself so I can def see how that would be annoying. I'll try to think of another name but unfortunately I think `geocode_resource` is the most concise and precise summation of what the function does 😂

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
```diff
@@ -0,0 +1,99 @@
...
+
+from smartystreets_python_sdk import StaticCredentials, ClientBuilder
+from smartystreets_python_sdk import us_street
+from smartystreets_python_sdk.us_street.lookup import Lookup
+
+from typing import List
+from copy import copy
+
+
+class SmartyGeocodeClient(FhirGeocodeClient):
```
spencer-skylight: I think this should be called SmartyFhirGeocodeClient to differentiate it from the one in phdi.geospatial.smarty.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
bamader: I'm actually completely in agreement with you. When I took on a big codebase refactor a two months ago, having returns and return types in doc strings was one of the things I standardized. However, I then went out on vacation and came back to find them removed from all the docstrings, so some kind of decision had taken place while I was gone that docstrings didn't need returns or return data types. Maybe @spencer-skylight can provide more clarity around this, because I personally think a docstring _should_ have return information. Since we're in control of the SDK, I guess we can always add them back, too?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
```diff
@@ -0,0 +1,99 @@
...
+    formatted data using the SmartyStreets API.
+    Requires an authorization ID as well as an authentication token
+    in order to build a street lookup client.
+    """
+
+    def __init__(self, auth_id, auth_token):
+        self.auth_id = auth_id
+        self.auth_token = auth_token
+        creds = StaticCredentials(auth_id, auth_token)
+        self.__client = (
```
spencer-skylight: I had thought we'd have the FhirGeocodeClient contain an instance of GeocodeClient rather than the vendor-specific implementation's client. That way you have access to GeocodeClient.geocode_from_string and geocode_from_dict.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
```diff
@@ -0,0 +1,99 @@
...
+        :param bundle: A bundle of fhir resources
+        :param overwrite: Whether to overwrite the address data in the given
+        bundle's resources (True), or whether to create a copy of the bundle
+        and overwrite that instead (False). Defaults to True.
+        """
+        if not overwrite:
+            bundle = copy.deepcopy(bundle)
+
+        for resource in bundle.get("entry", []):
+            _ = self.geocode_resource(resource.get("resource", {}), overwrite=True)
```
spencer-skylight: Since the hierarchy of a bundle is `bundle -> entry (array) -> resource`, could we slightly refactor to rename `resource` to `entry`?

src/lib/phdi-building-blocks/phdi_building_blocks/geospatial/smarty.py
```diff
@@ -0,0 +1,108 @@
...
+        Geocodes a dictionary-formatted address using SmartyStreets. The given
+        dictionary should conform to standard nomenclature around address fields,
+        including:
+
+            street: the number and street address
+            street2: additional street level information (if needed)
+            apartment: apartment or suite number (if needed)
+            city: city to geocode
+            state: state to geocode
+            zip: the postal code to use
```
spencer-skylight: Since the aims for conversion to the SDK include equity and accessibility, I think we should support as many address fields as we can... Could we rename zip -> postal code (foreign addresses and maybe territories might not have "zip"?), and add other things that might be supported by fhir and/or geocode vendor services (district, country)?

src/lib/phdi-building-blocks/phdi_building_blocks/geospatial/smarty.py
```diff
@@ -0,0 +1,108 @@
...
+    in order to build a street lookup client.
+    """
+
+    def __init__(self, auth_id, auth_token):
+        self.auth_id = auth_id
+        self.auth_token = auth_token
+        creds = StaticCredentials(auth_id, auth_token)
+        self.__client = (
+            ClientBuilder(creds)
+            .with_licenses(["us-standard-cloud"])
```
spencer-skylight: The license should probably be a parameter, and I get a little nervous that we are only supporting a small subset of configuration that could be applied to a Smarty client (just a general rumination, not actionable yet).

spencer-skylight (changes requested):
Great work - this is exciting! I had a few comments to review, and also wanted to suggest trying to move the old `fhir.py` to phdi/fhir/__init__.py while we work on migrating to the new package structure.  I think the two are equivalent, and we will thank ourselves for not having to change everything from phdi.fhir_wrapper to phdi.fhir at the end of the refactor work.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
bamader: Sure, I can do this

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
bamader: I think that makes sense for both of these functions, since they exist to help facilitate processing FHIR data. I'll go ahead and fold them into the abstract class.

bamader: I think they'll have to become static methods but that actually makes a lot of sense given the way we're using them--they don't care about class state at all and only operate on parameters in a utilitarian fashion so they can be bound to the class rather than specific objects.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
bamader: Oh yeah, this is a good call

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
bamader: Should we do this for both sets of abstract classes? I.e. the fhir wrapper and the raw data one?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/smarty.py
bamader: Ah I see. That makes sense to me, there's not really a reason to re-create client logic here.

bamader: Also lets me refactor a bit of duplicated logic out

src/lib/phdi-building-blocks/phdi_building_blocks/geospatial/smarty.py
bamader: Sure, I can add as many of those as I can

bamader: Turns out smarty only supports one more parameter (though it's an important one for Puerto Rican addresses) so we'll have to be conscious of that when implementing new vendor options

src/lib/phdi-building-blocks/phdi_building_blocks/geospatial/smarty.py
bamader: That's fair. I'll make the license parameterized, but for now I think let's err on the side of presenting that small configuration. If a user doesn't have a preexisting geocoding vendor license, they're going to default to smarty, so I say let's try to make it as simple on them as possible.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir_wrappers/geospatial/geospatial.py
spencer-skylight: Yeah, I think so, and we can use the same methodology as we develop the other modules as well.

bamader: That makes sense. And I like the way that cleans up both the directory structure and the imports.

phdi/geospatial/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+          2. makes it accessible through the use of .client()
+        This instance variable will hold whatever vendor-specific client
+        is instantiated by implementing classes so that they can perform
+        geocoding without referencing the underlying client that connects
+        to the vendor service.
+        """
+        pass
+
+    @abstractmethod
+    def geocode_from_str(self, address: str) -> Union[GeocodeResult, None]:
```
spencer-skylight, line 50: Can of worms warning... Would there ever be a need to return more than one result, like if there was an ambiguous match?

phdi/geospatial/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+@dataclass
+class GeocodeResult:
+    """
+    A basic dataclass representing a successful geocoding response.
+    """
+
+    street: List[str]
+    city: str
+    state: str
+    zipcode: str
```
spencer-skylight, line 15: Update to postalcode?

phdi/geospatial/smarty.py
```diff
@@ -0,0 +1,118 @@
...
+        lookup.zipcode = address.get("postal_code", "")
+        lookup.urbanization = address.get("urbanization", "")
+        lookup.match = "strict"
+
+        # Geocode and return
+        self.__client.send_lookup(lookup)
+        return _parse_smarty_result(lookup)
+
+
+def _parse_smarty_result(lookup):
```
spencer-skylight, line 88: First off, I really like how you've modularlized this code! :rocket:  

And a nit: would this make more sense as a staticmethod?

spencer-skylight (changes requested):

phdi/geospatial/smarty.py
bamader: Oooo good catch, yeah I definitely think it would. I'll fold it into the smarty client.

phdi/geospatial/__init__.py
bamader: Whoops good catch

phdi/fhir/geospatial/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    designed as a wrapper to process FHIR-based data (including resources
+    and bundles). Requires implementing classes to define methods to
+    geocode from both bundles and resources. Callers should use the
+    provided interface functions (e.g. geocode_resource) to interact with
+    the underlying vendor-specific client property.
+    """
+
+    @property
+    @abstractmethod
+    def geocode_client(self):
```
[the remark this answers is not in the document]
bamader: To clarify, is your statement about getting rid of the abstract property geocode_client, or the whole abstract class for a GeocodeClient?

Re: whether all services have a client: my understanding is that this is actually a fairly common practice for most geolocating services. Things like the Census API and MapBox provide clients, as far as I know, and we saw the same pattern just with accessing an Azure storage container (we set up a client for connection purposes). Insofar as that's true, I do think having the abstract makes some sense until/unless we see that most vendors in question aren't providing a specific client property, though this isn't a strongly held conviction on my part.

Re: your second bullet on vendor-specific clients having different programming interfaces: actually having this set up as an abstract property in the base class is necessary for implementing classes to actually access and use it in a uniform way. Based on my reading and understanding, this is the standard way to create "instance variables" for abstract python classes (which aren't a thing on their own because there's no constructor). Now, I do see what your point is ostensibly saying, which is that there's no reason a vendor-specific class couldn't just define its own extra instance variable if it needs it, and if that is the point you're making, then I could probably be convinced to switch.

phdi/geospatial/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    @abstractmethod
+    def geocode_from_str(self, address: str) -> Union[GeocodeResult, None]:
+        """
+        Function that uses the implementing client to perform geocoding
+        on the provided address, which is formatted as a string.
+        """
+        pass
+
+    @abstractmethod
+    def geocode_from_dict(self, address: dict) -> Union[GeocodeResult, None]:
```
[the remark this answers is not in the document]
bamader: Gotcha, I'll move the dictionary definition over to the abstract class. My worry with making it too FHIR-like is that if a user wants to use something FHIR-like, odds are they'll just import the fhir/geospatial module since we provide a perfect wrapper for a resource. If a user is making the conscious choice to geocode by generic dict rather than FHIR, I hesitate to impose too much in the way of an extension or nested entry structure just to access the address. And I think going really generic like this will help alleviate one of the concerns Jeremy voiced that users shouldn't have to provide too many parameters / should be able to supply whatever fields they have to geocode by address.

On the flip side, if your point is about using a FHIR-like structure in the sense that we can see what other geospatial fields FHIR supports and make those keys in the generic dict, that's something I could definitely get on board with, in which case just let me know!

phdi/geospatial/__init__.py
bamader: Ugh I thought a lot about this question when I was implementing it. What do we do if the geocoder returns multiple possible candidate matches... I personally settled on returning the first one for a couple of reasons:

1. it removes the hurdle of a user having to do filtering or post-processing to get a result they may just believe is simple. Trying to account for multiple matches means we'd have to do a lot more array formatting, which--while not necessarily bad--would make it more cumbersome to work with because the return data would always be _lists_ of GeocodeResults rather than just the desired result.
2. From what I found, SmartyStreets at least sorts the matches in the order of greatest likelihood, meaning that the first result is the one the geocoder considers "the best." I felt that for a dev release in particular, going with the single most likely result was good enough. I would hesitate to take on the additional burden of #1 above without having some user research or experimentation confirm for us that multiple matches is a problem users have.

spencer-skylight: Sounds good.  I thought through this more, and think it's OK for a few reasons:
* This is a good and simple approach that STLTs might want to adopt at least as a starting point
* We can extend the function (i.e. return_all_results parameter) or create new functions later to return more data if it's deemed necessary

spencer-skylight (commented):
One last remaining comment in phdi.geospatial.GeocodeClient for the geocode_from_dict function documentation - should contain the input format.  And related implementation in phdi.geospatial.smarty.SmartyGeocodeClient potentially updated to match documentation.

PR #7, opened 2022-08-02
First draft of tutorial docs

bamader:
This PR introduces a first pass at a geospatial tutorial, as well as some paragraph-level explanations within each of the geospatial package directories (normal and FHIR-wrapper). The structure and content of the tutorial are based heavily on the intro to numpy linked in the ticket for this task, while the directory structure is the agreed-upon taxonomy from our slack discussion.


tutorials/geospatial-tutorial.md
```diff
@@ -0,0 +1,213 @@
...
+district: Optional[str] = None
+country: Optional[str] = None
+county_name: Optional[str] = None
+precision: Optional[str] = None
+```
+
+All attributes which are `Optional`-typed may or may not be included in the results of a specific geocoding search, depending on whether a vendor implements returning data of that type.
+
+## Importing
+Using the geospatial module's functionality begins with simple imports. The Geocode Client abstract classes and `GeocodeResult` likely do not need to be imported (see the common uses section below), but can be directly from the relevant package:
```
KennethSkylight, line 54: I find it confusing to flip back and forth between FHIR and and non-FHIR. What do you think about breaking them up into their own sections with their own importing, common uses, and whatnot.

If I am a user who is trying to get my FHIR resource geocoded, I feel that I just want to read stuff related to FHIR and since its entangled, it makes it harder to read.

bamader: I thought about splitting it up, but I felt that that would create too many super short sections. Because the FHIR functionality is just based around wrappers for the normal geospatial functionality, splitting them off separately would create a lot of one or two line sections. So even though that particular part might be easier to read, I felt it would weaken the document overall. Plus I kind of thought it might be good to force FHIR readers to see the root part of the geospatial functionality as well, since they still need to understand geocode clients and results before they can really use the FHIR versions. It's not a hill I'm willing to die on though, so if you (and maybe Spencer) feel strongly I can reorganize.

spencer-skylight (approved):
Looks good to me!  And I enjoyed the David Lynch reference.

PR #8, opened 2022-08-05
Raw Data Harmonization

bamader:
This PR adds the harmonization functionality for the raw-data versions of the module (i.e. FHIR wrappers aren't currently written). Standardization and HL7 functionality are covered here, as well as unit tests for every function in the module. I went with a somewhat different directory structure than we initially discussed for harmonization, but nothing breaking: rather than put all the code, period, into a single `harmonization.py` file, I made a `harmonization/` directory in which there are a handful of underscore files (`_hl7`, `_standardization`) that group code based on its content. I thought that not only is this more maintainable for us because the code is more split out and modular, it's easier for a user to lookup things for a specific task than to sift through an entire mega module. Plus, it lets us cordon off the private helper functions that we decided not to expose to the user via an `__init__.py` file that imports everything that *will* be exposed at the top level.

I'm making this a PR on its own right now for two reasons:
1. so that I make PRs not as monolithic as they could be, because without any of the FHIR code this is already almost 800 lines of changes
2. to unblock the development of the FHIR wrappers if someone else wants to work on them. My plan is to keep pressing ahead with the FHIR code for this next week, but I figure making these modular would be good either way.


phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+    try:
+        # Conversion from \n to \r EOL characters is needed for hl7
+        # module, and doesn't impact conversion ability
+        parsed_message = hl7.parse(message.replace("\n", "\r"))
+
+        # Normalize Dates
+        # MSH-7 - Message date/time
+        _normalize_hl7_datetime_segment(parsed_message, "MSH", [7])
+
+        # PID-7 - Date of Birth
```
KennethSkylight, line 33: Instead of comments, these magic numbers should be constants

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+        normalized_datetime += hl7_datetime_parts[1][:5]  # . plus first 4 digits
+
+    # Add timezone information if present
+    if hl7_datetime_parts[2] and len(hl7_datetime_parts[2]) >= 5:
+        normalized_datetime += hl7_datetime_parts[2][:5]  # +/- plus 4 digits
+
+    return normalized_datetime
+
+
+def _clean_hl7_batch(batch: str, delimiter: str = "\n") -> str:
```
KennethSkylight, line 264: @spencer-skylight what are your thoughts for docstrings docs for private functions? Personally I think they should be optional, but wanted to get your thoughts on this

phdi/harmonization/_standardization.py
```diff
@@ -0,0 +1,176 @@
...
+    phones_to_clean = raw_phone
+    if isinstance(raw_phone, str):
+        phones_to_clean = [raw_phone]
+    outputs = []
+
+    for phone in phones_to_clean:
+        standardized = ""
+        for country in countries:
+
+            # We were able to pull the phone # and corresponding country
```
KennethSkylight, line 97: 97, 102, 106, and 117 are unnecessary comments

phdi/harmonization/_standardization.py
```diff
@@ -0,0 +1,176 @@
...
+      each representing a name
+    :param trim: Whether to strip leading/trailing whitespace
+    :param case: What case to enforce on each name; the default is to
+      convert all inputs to upper case
+    :remove_numbers: Whether to remove numeric characters from the inputs
+    :return: Either a string or a list of strings, depending on the
+      input of raw_name, holding the cleaned name(s)
+
+    """
+    # Allow users to pass in either a string or list of strings
```
KennethSkylight, line 152: Redundant, you explained this in the params docstrings

phdi/harmonization/_standardization.py
```diff
@@ -0,0 +1,176 @@
...
+            cleaned_name = cleaned_name.strip()
+        if case == "upper":
+            cleaned_name = cleaned_name.upper()
+        if case == "lower":
+            cleaned_name = cleaned_name.lower()
+        if case == "title":
+            cleaned_name = cleaned_name.title()
+        outputs.append(cleaned_name)
+
+    # Return same type as user input
```
KennethSkylight, line 173: unnecessary

phdi/harmonization/_hl7.py
KennethSkylight: Actually, after looking at it again, I'm not sure if constants would really even help since they're specific and I don't think it would any more sense. So I don't think these need to be changed

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+    cleaned_batch = _clean_hl7_batch(content)
+    message_lines = cleaned_batch.split(delimiter)
+    next_message = ""
+    output = []
+
+    for line in message_lines:
+        if line.startswith(("FHS", "BHS", "BTS", "FTS")):
+            continue
+
+        # If we reach a line that starts with MSH and we have
```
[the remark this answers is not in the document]
spencer-skylight: My 2c... I would tend to allow this kind of comment because it clarifies how the looping and next message detection works here.  I suspect the author had to think about this, and this way of viewing the looping logic made sense and could help future developers who use this code in the future.

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+    try:
+        # Conversion from \n to \r EOL characters is needed for hl7
+        # module, and doesn't impact conversion ability
+        parsed_message = hl7.parse(message.replace("\n", "\r"))
+
+        segment: hl7.Segment = None
+        try:
+            segment = parsed_message.segment(segment_id=segment_id)
+        except KeyError:
+            # If the segment is not found, there is nothing to do
```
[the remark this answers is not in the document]
spencer-skylight: I'm not sure it's immediately clear to a developer coming in to this code that KeyError = segment not found so I find some value in leaving this comment.

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+    :param default_value: If the selected field is blank, set the
+      field's value to this
+    :return: The HL7 message with default value inserted at the
+      specified segment location, if possible. If not, then the
+      original message.
+    """
+
+    parsed_message: hl7.Message = None
+    try:
+        # Conversion from \n to \r EOL characters is needed for hl7
```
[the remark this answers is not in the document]
spencer-skylight: IMO, this comment gives some context on why the EOL conversion is being done.

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+            segment.extract_field(field_num=field_num) is None
+            or segment.extract_field(field_num=field_num) == ""
+        ):
+            segment.assign_field(value=default_value, field_num=field_num)
+
+    except Exception:
+        logging.exception(
+            "Exception occurred while cleaning message.  "
+            + "Passing through original message."
+        )
```
spencer-skylight, line 188: See comment above - we may want to remove use of the logging module.

phdi/harmonization/_standardization.py
```diff
@@ -0,0 +1,176 @@
...
+      - alpha_3 would be "USA"
+      - numeric would be "840"
+
+    :param raw_country: String representation of the country to be
+      put in ISO 3611 standardized form
+    :param code_type: One of 'alpha_2', 'alpha_3', or 'numeric'; the
+      desired identifier type to generate
+    """
+
+    # First, identify what country the input is referencing
```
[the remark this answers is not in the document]
spencer-skylight: I think the comments in this function are helpful for organizational purposes. Perhaps pulling these sections of the code into private helpers would be better although I'd leave it to Brandon to decide whether this in scope for the work we are doing here.

phdi/harmonization/_standardization.py
```diff
@@ -0,0 +1,176 @@
...
+    :param countries: An optional list containing 2 letter ISO codes
+      associated with the phone numbers, signifying to which countries
+      the phone numbers might belong
+    :return: Either a string representing the cleaned phone numbers,
+      or a list of strings of cleaned numbers; if one phone number in
+      the input list couldn't be standardized, the index in the returned
+      list for that phone number will be the empty string
+    """
+
+    # Base cases: we always want to try the phone # on its own first;
```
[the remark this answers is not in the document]
spencer-skylight: I think most of the comments in this function are helpful for organizational purposes. I don't think they are required, but they help make the code more readable so my vote would be to leave them in.

spencer-skylight (commented):
A few things that need to be updated:
* `phdi.fhir.harmonization.__init__.py` I think should have standardize_names and standardize_phones
* flake8 issues resolved

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+        # Normalize Dates
+        # MSH-7 - Message date/time
+        _normalize_hl7_datetime_segment(parsed_message, "MSH", [7])
+
+        # PID-7 - Date of Birth
+        # PID-29 - Date of Death
+        # PID-33 - Last update date/time
+        _normalize_hl7_datetime_segment(parsed_message, "PID", [7, 29, 33])
+
+        # PV1-44 - Admisstion Date
```
bryan-skylight, line 38: typo: "Admisstion" -> "Admission"

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
+import logging
+import re
+import hl7
+
+from typing import List
+
+
+def standardize_hl7_datetimes(message: str) -> str:
```
bryan-skylight, line 8: Recommendation for a later PR: generalize this function such that it's not hardcoded like this. Allow the user to pass in a dictionary like {"pid": [7, 29, 33], "msh": [7]} and loop through the key:value pairs. They should also be able to specify the format that they want the timestamp in.

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # RXA-3 Date/time start of administration
+        # RXA-4 Date/time end of administration
+        # RXA-16 Substance expiration date
+        # RXA-22 System entry date/time
+        _normalize_hl7_datetime_segment(parsed_message, "RXA", [3, 4, 16, 22])
+
+    except Exception:
+        logging.exception(
+            "Exception occurred while cleaning message.  "
```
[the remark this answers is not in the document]
bryan-skylight: There are two scenarios in which error handling needs to be conducted:

1. Manual mode, where an analyst is manually running code in a notebook and will be able to see any exceptions raised, and
2. Automated mode, where a pipeline has been built and the code is being run from a script where no one will see the error.

In the first case, I think we should just raise the exception and ensure that the error message is clear so the analyst can address the issue. No need for logging. In the second case, I think we need a much more robust approach to raising the error, beyond just `logging.exception`. As was noted during the LAC meeting, the raising of an error should be actionable, not just some message stored in a file somewhere.

For the purposes of this PR, I'd suggest just leaving this as is given that I believe the point of the PR is to port existing functionality over to the new taxonomy. I'd also add a TODO comment highlighting the need to update this to fit our new approach in the future.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
bryan-skylight: I agree with Spencer. The code is clear to developers that are used to seeing this kind of loop, but our end users may not be as familiar. It's important to keep in mind who we're developing for and how tech savvy they may or may not be.

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+        # at that time.
+        if next_message != "" and line.startswith("MSH"):
+            output.append(next_message)
+            next_message = ""
+
+        # Otherwise, continue to add the line of text to next_message
+        if line != "":
+            next_message += f"{line}\r"
+
+    # Since our loop only adds messages to output when it finds
```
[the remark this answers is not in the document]
bryan-skylight: I think your comment here actually demonstrates why we need the comments above. I get the impression that lines 131-137 were clear to you and easily interpretable, but the intent of lines 142-143 was not as readily understood, which is why you feel some comments are necessary and others aren't. Our end users, and those who are likely to be reading our code, may have less exposure to code than you do and so both sections may not be very clear to them. This is why I advocate for airing on the side of verbosity in our comments.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
bryan-skylight: I agree with Kenneth here. There is only a single line of code in the `try` statement, and it's pretty clear what it's trying to do. And there is only one exception type being handled, so it's also clear what it's handling. I think we can remove this comment. 

But let's also keep in mind that this is PR is mostly geared towards just porting code over to the new taxonomy, albeit with some small changes. If we feel that a cleanup of the code is necessary, let's create a ticket and separate PR for that.

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+    :return: None (overwrites the input segment)
+    """
+    try:
+        for segment in message.segments(segment_id):
+            for field_num in field_list:
+                if len(segment) > field_num and segment[field_num][0] != "":
+                    cleaned_datetime = _normalize_hl7_datetime(segment[field_num][0])
+                    segment[field_num][0] = cleaned_datetime
+    except KeyError:
+        logging.debug(f"Segment {segment_id} not found in message.")
```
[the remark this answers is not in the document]
bryan-skylight: If you replace it with a comment, then the exception handler doesn't do anything and the exception would be raised, which removes the need for the `try/except` logic altogether. But again, see my comments about not overloading this PR.

phdi/harmonization/__init__.py
```diff
@@ -0,0 +1,9 @@
+from phdi.harmonization._hl7 import (
```
bryan-skylight, line 1: See my comment in the `_hl7.py` file about the use of underscore prefixes in file names.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
KennethSkylight: My counterpoint to that is that our end users are just importing the packages, they aren't looking at our code itself, but just importing it and using it. 

People who contribute to our repo are going to be a different audience than those who import our packages.

bamader:
After looking through the opinions on comments here, I'm going to leave all of them in unless there was consensus that one should be removed or abridged. Without reiterating my (long) train of thought from slack, I find comments exceedingly important for developers and users. They're adding a lot of valuable context and information for both while presenting few if any downsides. I think that just because we _could_ get away with omitting a comment doesn't mean we _should_.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
KennethSkylight: I just noticed that this repeated. So wouldn't the better fix to be to create a private function that is called `_remove_EOL` or something then we don't have to maintain the comments and the code in two places

phdi/harmonization/_hl7.py
KennethSkylight: Cool I'm fine with these docstrings for this private function, was just asking in general

phdi/harmonization/_standardization.py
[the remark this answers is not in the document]
KennethSkylight: I agree private helpers would make it better.

phdi/harmonization/_standardization.py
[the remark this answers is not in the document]
KennethSkylight: To me, the comment is just repeating what the code is clearly saying. It's also laid out in the docstrings what is happening

phdi/harmonization/_hl7.py
```diff
@@ -0,0 +1,283 @@
...
+    * Convert segment terminators back from ``\\r`` to ``\\n``
+
+    :param message: The raw HL7 message to sanitize
+    :return: The HL7 message with potential problem formats resolved,
+      if it was possible. If not, the original message.
+    """
+    parsed_message: hl7.Message = None
+    try:
+        # Conversion from \n to \r EOL characters is needed for hl7
+        # module, and doesn't impact conversion ability
```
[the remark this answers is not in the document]
bamader: Unfortunately it seems like this is a yes. I just spent some time combing over the HL7 package docs, and everywhere that they make invocations to parsing or even sending/receiving HL7 messages, there are always calls like `.replace('\r', '\n')`. It's never spelled out explicitly but the overwhelming frequency of their usage pattern makes me think we really do need this replacement :/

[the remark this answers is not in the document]
bryan-skylight: Wait, doesn't that prove my point? They're replacing `\r` with `\n` because they don't want to work with carriage returns either. Why are we doing this for them if they're already handling it?

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
bamader: We're opting to keep comment verbosity as well as comments that provide organizational, contextual, or even just summary information. As such, marking this as resolved.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
bamader: Agreed, this is the same principle I share. Marking as resolved.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
bamader: I'm going to leave it in because if I hadn't worked with the HL7 module, I wouldn't know that `.segment()` is actually an instance-retrieving function that looks up with segment ID as an actual key. The comment isn't hurting anything and may clarify something for someone down the road.

phdi/harmonization/_hl7.py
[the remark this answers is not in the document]
bamader: I agree with Spencer. Because the actual conversion of EOL characters is in itself a one-line function call, making a full function for it feels bizarre. I prefer the inline comment.

phdi/harmonization/_standardization.py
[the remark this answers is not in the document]
bamader: I'm going to pass on a private helper here for now. We can revisit that later if we so choose, but I almost am of the opinion that it doesn't really need one. The function's purpose is to implement very basic country code standardization, which starts with figuring out what country is being referenced. If we pull that out (and pull out everything that's commented into a private helper), I would argue we lose the meaningful logic of how that process is done, _in the function where the user would expect to find it_. It would be like a series of appendix references in a book or a paper, like "see X, p. 259; X: do this thing, see Y, p. 135; Y: statement of Z, see Z, p.399". I think this is non-burdensome enough and centralied logically tha tit makes sense to keep it all here.

phdi/harmonization/_standardization.py
bamader: I'm going to leave them in because I spent a fair bit of time looking into the `phonenumbers.NumberParseException` and reading about what it did. I had to do that because it wasn't clear to me what the objective of calling `parse` in this case was and exactly how much `NumberParseException` covered. And leaving in the 117 because it's very likely the reader has scrolled past the docstring at the top and its nice for them to get reminded of why stuff is there without having to reference something else or scroll.

phdi/harmonization/_standardization.py
bamader: Actually on second look here and the unanimity of some other stuff, I agree with you. Will delete this comment and the prior instance.

phdi/fhir/harmonization/__init__.py
```diff
@@ -0,0 +1,82 @@
+from typing import Literal
+import copy
+
+from phdi.fhir.harmonization.standardization import (
+    _standardize_names_in_resource,
+    _standardize_phones_in_resource,
+)
+
+
+def standardize_names(
```
spencer-skylight, line 10: Can you move these functions to `standardization.py` and just import those functions from here, similar to how we did the base harmonization?

phdi/harmonization/standardization.py
```diff
@@ -0,0 +1,178 @@
...
+      input of raw_name, holding the cleaned name(s)
+
+    """
+    names_to_clean = raw_name
+    if isinstance(raw_name, str):
+        names_to_clean = [raw_name]
+    outputs = []
+
+    for name in names_to_clean:
+        # Start by removing all punctuation
```
KennethSkylight, line 162: lol this super nitpicky, but I think just `remove all punctuation` is fine

phdi/harmonization/hl7.py
```diff
@@ -0,0 +1,289 @@
...
+
+        segment: hl7.Segment = None
+        try:
+            segment = parsed_message.segment(segment_id=segment_id)
+        except KeyError:
+            # If the segment is not found, there is nothing to do
+            return message
+
+        if (
+            segment.extract_field(field_num=field_num) is None
```
KennethSkylight, line 179: Using `!segment.extract_field(field_num=field_num)` may just work here because `None` and empty string are falsey.

phdi/harmonization/hl7.py
```diff
@@ -0,0 +1,289 @@
...
+      normalize the fields of
+    :param field_list: The list of fields contained in the
+      indexed message component, which are themselves indices to
+      data strings
+    :return: None (overwrites the input segment)
+    """
+    try:
+        for segment in message.segments(segment_id):
+            for field_num in field_list:
+                if len(segment) > field_num and segment[field_num][0] != "":
```
KennethSkylight, line 218: It's a little confusing to me what `segment[field_num][0]` is. Could we add a variable with a descriptive name to make it clear?

KennethSkylight (commented):
A few questions and super nit picky stuff, but other than that looks good.

tests/harmonization/test_hl7_underscore_funcs.py
```diff
@@ -0,0 +1,127 @@
...
+
+from phdi.harmonization.hl7 import (
+    _default_hl7_value,
+    _normalize_hl7_datetime_segment,
+    _normalize_hl7_datetime,
+    _clean_hl7_batch,
+)
+
+
+def test_default_hl7_value():
```
spencer-skylight, line 12: Very nitty, but I think I like the naming convention you used with the FHIR tests.  Could we call this test_hl7_helpers.py?

spencer-skylight (changes requested):
I think the only real blocking comment for me is the slight restructuring of phdi.fhir.harmonization. A few other very minor things, and should be GTG!

phdi/harmonization/hl7.py
bamader: You're totally right about this. However, we made a decision (long before you joined the team so not like you'd have any way of knowing this) that we as a team aren't going to use implicit truthy and falsey comparisons. Instead we're going to write out the full explicit `is None` or `is empty` or whatever the condition check may be. We felt this was a bit more readable, so this is just continuing that practice.

phdi/harmonization/hl7.py
bamader: @spencer-skylight Can we change `field_num` to something like component, or is field already the right nomenclature? I want to make sure we're using the right terminology to address Kenneth's request here.

@KennethSkylight Basically this is an HL7 terminology minefield lol. An HL7 message is broken down into specific _segments_, which are things like PID, OBX, etc. A segment is an order-specific "list", in essence, of sub-fields that contain health-related information. I _believe_ the fields within a segment are actually just called _fields_ in a technical sense, but I could be wrong and they could be components. We just want to make sure we're using the right word here so that people familiar with HL7 will know what the variables mean (since we're making that assumption of familiarity for our user base).

phdi/harmonization/hl7.py
KennethSkylight: Interesting, sounds good to me

phdi/harmonization/hl7.py
spencer-skylight: `field` is the correct nomenclature for this...
@KennethSkylight 's interpretation is correct, and to address his question, the pipe (|) character separates fields within a segment (e.g. name, gender, etc).  The carat (^) character separates components within a field (e.g. first name, last name, etc). There are also other characters that can further divide components into subcomponents - the ampersand (&), and separate repeated field values - the tilde (~).  

So I think the nomenclature is good. To make it slightly more clear what that value means, you could do something like this...
```suggestion
                component_num = 0  # The datetime value is always found in the 1st component
                field_value = segment[field_num][component_num]
                if len(segment) > field_num and field_value != "":
```

phdi/fhir/utils.py
```diff
@@ -0,0 +1,47 @@
...
+      found in the input bundle
+    """
+    return [
+        resource
+        for resource in bundle.get("entry")
+        if resource.get("resource").get("resourceType") == resource_type
+    ]
+
+
+def get_field(resource: dict, field: str, use: str, default_field: int) -> str:
```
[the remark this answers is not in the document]
bamader: Yep done

tests/harmonization/test_hl7_underscore_funcs.py
bamader: Sure thing

phdi/harmonization/hl7.py
bamader: Well, have to do a modified version to make sure `len(segment) > field_num` but otherwise this is good

spencer-skylight (approved):
LGTM!

PR #12, opened 2022-08-11
Transport Base Module

spencer-skylight:
Create `phdi.transport module`

A few tweaks were made in addition to moving this code:
* Previously we attempted to return an unset `response` if an unsupported `request_type` was supplied. Now we return a `ValueError`.  Also documented that an exception is raised in this case
* `request_type` is typed as a literal with "GET" and "POST" options allowed
* Added TODO comments to circle back on removing logging calls
* There were no existing unit tests for this function, so those are new


phdi/transport/http.py
```diff
@@ -0,0 +1,71 @@
...
+    if request_type == "POST":
+        try:
+            response = http.post(
+                url=url,
+                headers=headers,
+                json=data,
+            )
+        except Exception:
+            logging.exception(
+                "POST request to " + url + " failed for data: " + str(data)
```
KennethSkylight, line 56: Use f string
```suggestion
               f'POST request to {ur}l failed for data: {str(data)}'
```

KennethSkylight: Unless this line will be removed because of your logging comment in the description?

KennethSkylight (commented):
I think you mentioned you were going to remove the logging as a todo, so ignore if that is the case

phdi/transport/http.py
spencer-skylight: Yeah, this will go away with the exception logging, but still a good idea.  I also added TODO comments as a flag to come back to this.

phdi/transport/http.py
```diff
@@ -0,0 +1,71 @@
...
+    request_type: Literal["GET", "POST"],
+    allowed_methods: List[str],
+    headers: dict,
+    data: dict = None,
+) -> requests.Response:
+    """
+    Carryout an HTTP Request using a specific retry strategy. Essentially
+    a wrapper function around the retry strategy implementation of a
+    mounted HTTP request.
+    :param url: The url at which to make the HTTP request
```
bamader, line 21: Nit: I've been leaving a blank newline between the end of the summary docstring and the start of the params list, which I like for visual appearance

phdi/transport/http.py
```diff
@@ -0,0 +1,71 @@
...
+    :param retry_count: The number of times to re-try the request, if the
+    first attempt fails
```
bamader, line 23: Nit: I've been doing two-space indents whenever a param goes beyond the first line, as I think it improves readability

bamader (approved):
Looks good aside from just a few nits around formatting for the comments! (I'm approving this so that if you don't get to the comment formatting tonight I'm not blocking this being merged while I'm out tomorrow)

PR #13, opened 2022-08-11
Adding pypi fields to pyproject toml. Also fixing the version for fhirpathpy

KennethSkylight:
To upload our package to pypi, we want to include the readme and relevant links for users to see when they arrive on our pypi page


pyproject.toml
```diff
@@ -1,8 +1,12 @@
 [tool.poetry]
 name = "phdi"
-version = "0.1.0"
-description = ""
-authors = []
+version = "0.1.1"
+description = "Public health data infrastructure Building Blocks is a library to help public health departments work with their data"
+authors = [Kenneth Chow <kenneth@skylight.digital>, Brandon Mader <brandon@skylight.digital>, Spencer Kathol <spencer@skylight.digital>]
```
spencer-skylight: I think you need to wrap the list elements in double-quotes.

pyproject.toml
```diff
@@ -1,8 +1,12 @@
 [tool.poetry]
 name = "phdi"
-version = "0.1.0"
-description = ""
-authors = []
+version = "0.1.1"
```
spencer-skylight, line 3: Is it possible to keep the version at 0.1.0 (and does it permit the -dev prefix?) in this initial time before the pypi registration is actually complete? It currently disagrees with the documentation location.

I will also look in to how the version string and related tracking is normally managed, in an attempt to automate as much as possible.

spencer-skylight (changes requested):
Looks good - just a question about version I think we need to figure out.

pyproject.toml
KennethSkylight: Looks like I can do 0.1.dev1

According to this: https://peps.python.org/pep-0440/#examples-of-compliant-version-schemes

Would that work?

bamader (commented):
Agree with Spencer -- this looks good aside from the version change I'd like to see implemented, if that's possible.

pyproject.toml
KennethSkylight: Ok changed

spencer-skylight (approved):
LGTM!

bamader (approved):

PR #14, opened 2022-08-11
Cloud module credmanager

spencer-skylight:
Move the BaseCredentialManager and AzureCredentialManager into the new SDK taxonomy.

### DIFF
--- a/phdi/cloud/core.py
+++ b/phdi/cloud/core.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from typing import Any, Union, IO
 
 
 class BaseCredentialManager(ABC):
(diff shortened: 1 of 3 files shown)


phdi/cloud/core.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        Create a new BaseCredentialManager object.
+        This object type is not intended to be called directly, but the constructor
+        may be called from base classes.
+
+        :param resource_location: URL or other location of the requested resource.
+        :param scope: A space-delimited list of scopes to limit access to resource.
+        """
+        self.resource_location = resource_location
+        self.scope = scope
```
bamader, line 20: Do we want to define the constructor in the ABC? We don't actually have to do it here because it's abstract; that means if we can't be sure that every type of credentials manager will hold a string `resource_location` and a string `scope` I'd argue we shouldn't define this here. We should only put this here if those properties will exist in exactly this form for every such cred manager.

bamader, line 20: In particular, I think `resource_location` and `scope` may be more appropriately typed as `property`s, which you define as though they were a function on the ABC and then inheriting classes create as a private instance variable accessible through standard `.property()` calls.

bamader (commented):
The AzureCredManager implementation looks good. Just some questions around how we're defining the abstract base class (i.e. whether to define a constructor with specified variables and whether to make them properties or not).

phdi/cloud/core.py
spencer-skylight: I think you're right - we don't necessarily know whether other cloud implementations will need these properties or implementation, so I've removed the constructor.

spencer-skylight: I added the properties to the azure class which is in line with how you structured the GeocodeClient work, and I think appropriate here as well.

bamader (approved):
Looks good!

PR #16, opened 2022-08-12
Transport module fhir

spencer-skylight:
Create `phdi.fhir.transport` module.


DanPaseltiner (approved):
LGTM.

bamader (changes requested):
A small thing, but I think we'd said that we're going to adopt a convention of using `__init__` files only to hold the imports that will be publicly exposed to the user. So I think everything that's currently in `transport/__init__.py` should be moved to another file, and then init can import them and use `__all__` to expose the namespaces to the users. Otherwise, looks good!

bamader (approved):

PR #17, opened 2022-08-12
Make get_access_token return a string, allow token reuse

spencer-skylight:
Previously, the BaseCredentialManager returned a cloud-specific object from get_access_token.  After talking with Dan about GCP, we are changing to string to make our wrapper functions more cloud agnostic.  The Azure AccessToken object has been made available as a property of the class. 

Also, the AzureCredentialManager is now reusing tokens if the access_token has not expired.

### DIFF
--- a/phdi/cloud/azure.py
+++ b/phdi/cloud/azure.py
@@ -1,6 +1,7 @@
 from .core import BaseCredentialManager
 from azure.core.credentials import AccessToken
 from azure.identity import DefaultAzureCredential
+from datetime import datetime, timezone
 
 
 class AzureCredentialManager(BaseCredentialManager):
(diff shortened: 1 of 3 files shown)


DanPaseltiner (approved):
This looks good to me. I would like to flag that implementing this means we will need to change to lines like `access_token = cred_manager.get_access_token().token` to `access_token = cred_manager.get_access_token()` in at least `upload_bundle_to_fhir_server()` and `_http_request_with_reauth()`. I believe these functions are still in the works for this repo so I don't think this should be blocking now, but something we will need to look out for in future PRs that seek to merge these functions.

spencer-skylight:
Thanks Dan! I have an outstanding PR with those functions, and will update those calls as you described in that PR.

PR #18, opened 2022-08-12
Linkage Module

KennethSkylight:
This PR moves the linkage module into the new file path. The tests are moved as well


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
```
spencer-skylight, line 2: I think we should delete from source control at least for now (it may be a good idea to explicitly add `/.vscode/settings.json` it to `.gitignore`).  

We could still add recommended vscode settings to the repository in the future, but if/when we do that, we should synchronize all settings documented in `CONTRIBUTING.md`. I'm also not sure what would happen if users wanted to modify this file with their own desired extensions that we don't necessarily want in the repo, so that's something else to figure out.

phdi/linkage/README.md
```diff
@@ -0,0 +1,3 @@
+# The Linkage Package
+
+This package contains functions useful in the geocoding of raw input data (i.e. strings and dictionaries having no particular formatting standard). The `__init__.py` file defines the abstract `GeocodeClient` that vendor-specific implementations inherit from, as well as the standard `GeocodeResult` class that encapsulated returned geocoding information in a standardized data structure. All other files in this package directory are vendor-specific implementations of geocoding, with one file per vendor. Each has access to the same basic geocoding utilities, `.geocode_from_str()` and `.geocode_from_dict()`. For an overview of the package as well as common use cases, see the _geospatial-tutorial_ in `/tutorials` at the project root. For a more in-depth explanation of specific formatting and parameters, see the docstrings and comments for each function.
```
spencer-skylight, line 3: This still has content for geospatial, and should be updated.

phdi/linkage/__init__.py
```diff
@@ -83,3 +80,17 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
     to_encode = (linking_identifier + salt_str).encode("utf-8")
     hash_obj.update(to_encode)
     return hash_obj.hexdigest()
+
+
+def _get_one_line_address(address: dict) -> str:
```
spencer-skylight, line 85: IMO this should be moved to `phdi.fhir.utils` and made public. This is the exact same implementation as in `phdi.fhir.geospatial`, and since we are calling it from 2 submodules at this point, and seems like something that could be useful to STLTs as a standalone function, we could make it public (remove the leading _).

@bamader I wanted to get your opinion as well. What do you think?

spencer-skylight (changes requested):
A few minor suggestions.

.vscode/settings.json
spencer-skylight: I went ahead and made this change.

phdi/linkage/__init__.py
spencer-skylight: After chatting with Brandon - this makes sense to him as well.

KennethSkylight: Moved

spencer-skylight (approved):
LGTM!

PR #20, opened 2022-08-16
Remove kwargs from signature

bamader:
This small PR removes `**kwargs` from the function signature of `fhir.transport.http_request_with_reauth`, replacing it with the direct enumerated parameters instead. No usage needs to change at all to make this work, as the underlying `http_request_with_retry` was already using direct parameters; this PR merely brings this function in line with our HCD approach and not using kwargs in the codebase.


spencer-skylight (approved):
LGTM

PR #21, opened 2022-08-17
Cloud module storage

spencer-skylight:
Create the CloudContainerConnection interface and related Azure implementation.

### DIFF
--- a/phdi/cloud/azure.py
+++ b/phdi/cloud/azure.py
@@ -18,8 +18,8 @@ class AzureCredentialManager(BaseCredentialManager):
     """
 
     @property
-    def storage_account_url(self) -> str:
-        return self.__storage_account_url
+    def resource_location(self) -> str:
+        return self.__resource_location
 
     @property
     def scope(self) -> str:
@@ -29,19 +29,19 @@ class AzureCredentialManager(BaseCredentialManager):
     def access_token(self) -> AccessToken:
         return self.__access_token
 
-    def __init__(self, storage_account_url: str, scope: str = None):
+    def __init__(self, resource_location: str, scope: str = None):
         """
         Create a new AzureCredentialManager object.
 
-        :param storage_account_url: URL or other location of the requested resource.
+        :param resource_location: URL or other location of the requested resource.
         :param scope: A space-delimited list of scopes to limit access to resource.
         """
-        self.__storage_account_url = storage_account_url
+        self.__resource_location = resource_location
         self.__scope = scope
         self.__access_token = None
 
         if self.scope is None:
-            self.__scope = f"{self.storage_account_url}/.default"
+            self.__scope = f"{self.resource_location}/.default"
 
     def get_credential_object(self) -> object:
         """
--- a/phdi/cloud/azure.py
+++ b/phdi/cloud/azure.py
@@ -1,6 +1,6 @@
+import io
 import json
 import logging
-import pathlib
 import requests
 
 from .core import BaseCredentialManager, CloudContainerConnection
@@ -9,6 +9,7 @@ from azure.core.exceptions import ResourceExistsError
 from azure.identity import DefaultAzureCredential
 from azure.storage.blob import ContainerClient
 from datetime import datetime, timezone
+from typing import IO
 
 
 class AzureCredentialManager(BaseCredentialManager):
@@ -115,7 +116,23 @@ class AzureCloudContainerConnection(CloudContainerConnection):
         creds = self.cred_manager.get_credential_object()
         return ContainerClient.from_container_url(container_url, credential=creds)
 
-    def store_data(
+    def download_object(
+        self, container_name: str, filename: str, io_stream: IO = None
+    ) -> IO:
+        container_location = f"{self.resource_location}/{container_name}"
+        container_client = self._get_blob_client(container_location)
+        blob_client = container_client.get_blob_client(filename)
+
+        downloader = blob_client.download_blob()
+
+        if io_stream is None:
+            io_stream = io.BytesIO()
+
+        downloader.download_to_stream(io_stream)
+
+        return io_stream
+
+    def upload_object(
         self,
         container_name: str,
         filename: str,
@@ -134,61 +151,68 @@ class AzureCloudContainerConnection(CloudContainerConnection):
         :param message_json: The content of a message encoded in json format.
         :param message: The content of a message encoded as a string.
         """
-        container_location = str(pathlib.Path(self.resource_location / container_name))
-        client = self._get_blob_client(container_location)
-        blob = client.get_blob_client(filename)
+        container_location = f"{self.resource_location}/{container_name}"
+        container_client = self._get_blob_client(container_location)
+        blob_client = container_client.get_blob_client(filename)
 
         if message_json is not None:
-            blob.upload_blob(json.dumps(message_json).encode("utf-8"), overwrite=True)
+            blob_client.upload_blob(
+                json.dumps(message_json).encode("utf-8"), overwrite=True
+            )
         elif message is not None:
-            blob.upload_blob(bytes(message, "utf-8"), overwrite=True)
+            blob_client.upload_blob(bytes(message, "utf-8"), overwrite=True)
 
-    def store_message_and_response(
-        self,
-        container_name: str,
-        message_filename: str,
-        response_filename: str,
-        message: str,
-        response: requests.Response,
-    ):
-        """
-        Store information about an incoming message as well as an http response for a
-        transaction related to that message.  This method can be used to
-        record a failed response to a transaction related to an inbound transaction for
-        troubleshooting purposes.
+    def list_containers(self):
+        return super().list_containers()
 
-            :param container_url: The url at which to access the container
-            :param prefix: The "filepath" prefix used to navigate the
-                virtual directories to the output container
-            :param bundle_type: The type of data being written
-            :param message_filename: The file name to use to store the message
-                in blob storage
-            :param response_filename: The file name to use to store the response content
-                in blob storage
-            :param message: The content of a message encoded as a string.
-            :param response: HTTP response information from a transaction related to the
-                `message`.
-        """
-        try:
-            # First attempt is storing the message directly in the
-            # invalid messages container
-            self.store_data(
-                container_name=container_name,
-                filename=message_filename,
-                message=message,
-            )
-        except ResourceExistsError:
-            logging.warning(
-                f"Attempted to store preexisting resource: {message_filename}"
-            )
-        try:
-            # Then, try to store the response information
-            self.store_data(
-                container_name=container_name,
-                filename=response_filename,
-                message=response.text,
-            )
-        except ResourceExistsError:
-            logging.warning(
-                f"Attempted to store preexisting resource: {response_filename}"
-            )
+    def list_objects(self):
+        return super().list_objects()
+
+
+def store_message_and_response(
+    client: AzureCloudContainerConnection,
+    container_name: str,
+    message_filename: str,
+    response_filename: str,
+    message: str,
+    response: requests.Response,
+):
+    """
+    Store information about an incoming message as well as an http response for a
+    transaction related to that message.  This method can be used to
+    record a failed response to a transaction related to an inbound transaction for
+    troubleshooting purposes.
+
+        :param client: An instance of `AzureCloudContainerConnection` used to
+          upload the request
+        :param container_url: The url at which to access the container
+        :param prefix: The "filepath" prefix used to navigate the
+          virtual directories to the output container
+        :param bundle_type: The type of data being written
+        :param message_filename: The file name to use to store the message
+          in blob storage
+        :param response_filename: The file name to use to store the response content
+          in blob storage
+        :param message: The content of a message encoded as a string.
+        :param response: HTTP response information from a transaction related to the
+          `message`.
+    """
+    try:
+        # First attempt is storing the message directly in the
+        # invalid messages container
+        client.store_data(
+            container_name=container_name,
+            filename=message_filename,
+            message=message,
+        )
+    except ResourceExistsError:
+        logging.warning(f"Attempted to store preexisting resource: {message_filename}")
+    try:
+        # Then, try to store the response information
+        client.store_data(
+            container_name=container_name,
+            filename=response_filename,
+            message=response.text,
+        )
+    except ResourceExistsError:
+        logging.warning(f"Attempted to store preexisting resource: {response_filename}")
--- a/phdi/cloud/azure.py
+++ b/phdi/cloud/azure.py
@@ -1,11 +1,9 @@
 import io
 import json
-import logging
 import requests
 
 from .core import BaseCredentialManager, BaseCloudContainerConnection
 from azure.core.credentials import AccessToken
-from azure.core.exceptions import ResourceExistsError
 from azure.identity import DefaultAzureCredential
 from azure.storage.blob import ContainerClient, BlobServiceClient
 from datetime import datetime, timezone
--- a/phdi/cloud/azure.py
+++ b/phdi/cloud/azure.py
@@ -94,8 +94,8 @@ class AzureCloudContainerConnection(BaseCloudContainerConnection):
         Create a new AzureCloudContainerConnection object.
 
 
-        :param storage_account_url: Storage account location of the requested resource.
-        :param cred_manager: The Azure credential manager.
+        :param storage_account_url: Storage account location of the requested resource
+        :param cred_manager: The Azure credential manager
         """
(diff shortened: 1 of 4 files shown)


phdi/cloud/azure.py
```diff
@@ -1,12 +1,20 @@
...
 from azure.identity import DefaultAzureCredential
+from azure.storage.blob import ContainerClient, BlobServiceClient
 from datetime import datetime, timezone
+from typing import IO, List
 
 
 class AzureCredentialManager(BaseCredentialManager):
     """
-    This class provides an Azure-specific credential manager.
+    This class implements the PHDI credential manager interface for connecting to Azure.
```
bryan-skylight, line 17: The use of "interface" in this context is confusing. Interfaces serve as the blueprint for classes, which classes then implement. In this scenario, BaseCredentialManager is the interface, and AzureCredentialManager is the implementation. I'd recommend we remove the word "interface" from the doctstring. I'm also not sure that we need to call it "the PHDI credential manager". I don't know why, but it gives me a weird reaction when I read it. Like hearing people refer to it as The Ohio State University.

phdi/cloud/core.py
```diff
@@ -20,3 +21,60 @@ def get_access_token(self) -> str:
         Returns an access token using the managed credentials
         """
         pass
+
+
+class CloudContainerConnection(ABC):
```
bryan-skylight, line 26: We need to come up with a common naming convention for implementing interfaces in our code base. Currently, we have "BaseCredentialManager" and this interface with no prefix. 

Python's PEP8 doesn't have recommendations on this, but in general interfaces are named `I<Name>`. I'd recommend we either go that route, or we stick with `Base<Name>`. Either way, we should do something to indicate that it's an interface.

phdi/cloud/core.py
```diff
@@ -20,3 +21,60 @@ def get_access_token(self) -> str:
...
+        self, container_name: str, filename: str, stream: IO = None
+    ) -> IO:
+        """
+        Downloads a blob from storage.
+
+        :param container_name: Storage container name.
+        :param filename: Location of file within storage.
+        :param stream: (optional) stream object that should be used to write output
+          contents of blob.
+        :return: The `stream` parameter, if supplied. Otherwise a new stream object
```
bryan-skylight, line 38: I thought the consensus on docstrings was that we no longer included the `return` field?

phdi/cloud/core.py
```diff
@@ -20,3 +21,60 @@ def get_access_token(self) -> str:
...
+        container_name: str,
+        filename: str,
+        message_json: dict = None,
+        message: str = None,
+    ) -> None:
+        """
+        Uploads content to storage.
+        Exactly one of message_json or message should be provided.
+
+        :param container_name: Storage container name.
```
bryan-skylight, line 55: This parameter description is not very helpful in my opinion. Perhaps we could change it to something like "The name of the container where the object should be stored"?

phdi/cloud/core.py
```diff
@@ -20,3 +21,60 @@ def get_access_token(self) -> str:
...
+        :param filename: Location of file within storage container.
```
bryan-skylight, line 56: Similarly, this description could possibly be something like "The filename to use for the object being uploaded". As it's currently written, it almost sounds as if the object should already exist in the storage container.

phdi/cloud/core.py
```diff
@@ -20,3 +21,60 @@ def get_access_token(self) -> str:
...
+        :param message_json: The content of a message a json-formatted dict.
```
bryan-skylight, line 57: This description is missing a word I think. "The content of a message _as_ a json-formatted dict". Although, I'd also point out that JSON-formatted dict is redundant and we might be better off saying "The content of a message in JSON format".

In addition, I believe our human-centered design principles state that the data should be the first parameter in a function, meaning `message_json` and `message` should come before `container_name` and `filename`.

phdi/cloud/azure.py
bryan-skylight: The use of "interface" in this context is confusing. Interfaces serve as the blueprint for classes, which classes then implement. In this scenario, BaseCredentialManager is the interface, and AzureCredentialManager is the implementation. I'd recommend we remove the word "interface" from the doctstring. I'm also not sure that we need to call it "the PHDI credential manager". I don't know why, but it gives me a weird reaction when I read it. Like hearing people refer to it as The Ohio State University.

bamader: What about something simple like "this class defines a credential manager for connecting to Azure"? In Spencer's defense, I think the sentence construction he has written is logical if both "the" and "interface" remain in the syntax. The implication is that PHDI will provide one credential manager for connecting to Azure, and this class is that credential manager. The other alternative is to replace "the" with "a", since you can't drop the article entirely without it being syntactically incorrect, and that one strikes me as equally valid (it just doesn't have the implication of single provision that using "the" does).

phdi/cloud/core.py
bamader: I think technically it's not even an interface in the formal sense, since it can define a constructor and its own instance variables / properties. I would actually really push back against prefixing these with an I as I think it impairs readability and provides no additional value since ABC is used in the same line as every name definition. I personally would also not opt for using a `Base<Name>` convention either, since for things like `FhirGeocodeClient` that gets clunky and again I think the ABC in the definition is doing that work for us. But, if I were forced to choose between these two, I would opt for Base over I.

phdi/cloud/core.py
bamader: The SDK team agreed unanimously before we started taxonomy porting that we wanted to add it back for the new ports. We don't have any of the types specified, as Sphinx can auto-parse those, but we felt it was important from a design perspective to add some specific details around returned values, which could be clarifying to a user. Since it doesn't break up the code flow at all we felt it was a non-harmful add that conferred some additional value for users.

phdi/cloud/azure.py
```diff
@@ -72,3 +80,180 @@ def _need_new_token(self) -> bool:
...
+        self.__storage_account_url = storage_account_url
+        self.__cred_manager = cred_manager
+
+    def _get_container_client(self, container_url: str) -> ContainerClient:
+        """
+        Obtains a client connected to an Azure storage container by
+        utilizing the first valid credentials Azure can find. For
+        more information on the order in which the credentials are
+        checked, see the Azure documentation:
+        https://docs.microsoft.com/en-us/azure/developer/python/sdk/authentication-overview#sequence-of-authentication-methods-when-using-defaultazurecredential
```
bamader, line 115: This might need a `#noqa` to avoid flake getting angry

bamader (commented):
Most of the code looks good to me aside from the parameter things Bryan pointed out. Once we get those cleaned up I'm good to approve, but let's also discuss any conventions we want around AbstractBaseClasses

phdi/cloud/azure.py
spencer-skylight: I went with Brandon's suggestion here on the simple comment.

phdi/cloud/core.py
spencer-skylight: Prefixed with `Base`.

phdi/cloud/core.py
spencer-skylight: Per Slack conversation, removed `return` comments from this PR.
Ticket added to circle back on previously merged code: https://app.clickup.com/t/3bukhp7

phdi/cloud/core.py
spencer-skylight: Updated each container_name comment.

phdi/cloud/core.py
spencer-skylight: Comment updated as suggested.

As currently defined, the `message_json` and `message` parameters are optional, and can't be listed before required parameters. This could be refactored in a few different ways to make this more in line with HCD principles. I've got a few opinions on this, but thinking this might be a task to circle back on.  Is it OK to create a separate ticket for this?

phdi/cloud/azure.py
spencer-skylight: Flake seems to be happy.  I think it is more lenient about URLs.

phdi/cloud/azure.py
```diff
@@ -64,11 +71,183 @@ def _need_new_token(self) -> bool:
...
+    try:
+        # Then, try to store the response information
+        client.upload_object(
+            container_name=container_name,
+            filename=response_filename,
+            message=response.text,
+        )
+    except ResourceExistsError:
+        # TODO evaluate whether we want to log errors or throw them
+        logging.warning(f"Attempted to store preexisting resource: {response_filename}")
```
bamader, line 253: Maybe it's a nit or beyond PR scope, but how would you feel about just changing these into raised Exceptions as part of this PR work? I know it's technically just a port but I've been doing that in mine and it honestly seems small enough to merit just doing here. If you'd rather ticket that later that's okay with me though.

bamader (approved):
LGTM. I left one question around converting those `logging` statements into raised exceptions--they occur in three back to back places in one file, and the lift seemed small enough to merit just doing now. However, if you'd rather keep this strictly as a port, I won't block, so giving approval in case you decide to merge later or make the change and I've gone offline.

phdi/cloud/azure.py
spencer-skylight: In some cases it's not as straightforward how we should be reacting in certain error scenarios, but this seems relatively cut and dry that we can allow the caller to handle. I went ahead and removed the exception handling here.

phdi/cloud/core.py
bryan-skylight: Yup, you're right, I completely spaced on the optional vs required parameter positioning. Agreed that this should be a separate ticket and followed-up on.

phdi/cloud/azure.py
```diff
@@ -64,11 +69,175 @@ def _need_new_token(self) -> bool:
...
+        :param container_url: The url at which to access the container
+        """
+        creds = self.cred_manager.get_credential_object()
+        return ContainerClient.from_container_url(container_url, credential=creds)
+
+    def download_object(
+        self, container_name: str, filename: str, stream: IO = None
+    ) -> IO:
+        """
+        Downloads a blob from storage. The `stream` parameter, if supplied.
```
bryan-skylight, line 122: The description here doesn't make sense after moving the return description to here. Can we tweak this to read a little better?

phdi/cloud/azure.py
```diff
@@ -64,11 +69,175 @@ def _need_new_token(self) -> bool:
...
+
+        return blob_name_list
+
+
+def store_message_and_response(
+    client: AzureCloudContainerConnection,
+    container_name: str,
+    message_filename: str,
+    response_filename: str,
+    message: str,
```
bryan-skylight, line 212: `message` should be the first parameter.

bryan-skylight (changes requested):
Overall, this looks great. I just have a couple of blocking things, one of which is a question and the other is a requested change.

phdi/cloud/core.py
spencer-skylight: Created ticket. https://app.clickup.com/t/3c06bt8

phdi/cloud/azure.py
spencer-skylight: Updated.

phdi/cloud/azure.py
```diff
@@ -64,11 +69,175 @@ def _need_new_token(self) -> bool:
...
+        blob_client = container_client.get_blob_client(filename)
+
+        downloader = blob_client.download_blob()
+
+        if stream is None:
+            stream = io.BytesIO()
+
+        downloader.download_to_stream(stream)
+
+        return stream
```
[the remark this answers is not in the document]
spencer-skylight: There are a few cases I wanted to support:
* Binary type documents (e.g. images, PDF)
* Writing to file stream (avoiding the requirement to keep downloaded content in memory)
* Avoids us needing to handle character encoding.  In other places, right now we assume UTF-8, but ISO 8859-1 and Windows-1252 are also common. 

If you don't think these functions fit in to our use cases (after all, we only support uploading as UTF-8 text currently), I can change to return content converted to a string.

phdi/cloud/azure.py
spencer-skylight: Updated.

bryan-skylight (changes requested):
For the most part, this looks really good. I have a question about the inclusion of `store_message_and_response` that I'd like to hash out before merging, but otherwise this looks like it's good to go.

phdi/cloud/azure.py
```diff
@@ -64,11 +69,177 @@ def _need_new_token(self) -> bool:
...
+        blob_properties_generator = container_client.list_blobs(name_starts_with=prefix)
+
+        blob_name_list = []
+        for blob_propreties in blob_properties_generator:
+            blob_name_list.append(blob_propreties.name)
+
+        return blob_name_list
+
+
+def store_message_and_response(
```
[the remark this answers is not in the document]
spencer-skylight: TLDR - I agree, removing.

I agree it's more implementation-specific and not a building block per se. I see it as a utility function and a suggestion of how a STLT might store a response error in conjunction with ingested message content. It is still the best option I can think of in the absence of another PII-safe logging or error management framework. On the other hand, it's clunky and has significant limitations in terms of error reporting and management. It might be better to encourage STLTs to brainstorm and implement something more robust. If they do end up implementing a similar approach, the function is so simple, they wouldn't need to spend much time recreating it.

phdi/cloud/azure.py
```diff
@@ -64,11 +69,177 @@ def _need_new_token(self) -> bool:
...
+        elif message is not None:
+            blob_client.upload_blob(bytes(message, "utf-8"), overwrite=True)
+
+    def list_containers(self) -> List[str]:
+        """
+        List names for this CloudContainerConnection's containers
+
+        """
+        creds = self.cred_manager.get_credential_object()
+        print(self.storage_account_url)
```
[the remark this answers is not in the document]
spencer-skylight: Just left behind from testing. Thanks for catching this!

phdi/cloud/azure.py
```diff
@@ -38,19 +44,18 @@ def __init__(self, resource_location: str, scope: str = None):
...
-        :return: Returns an instance of one of the *Credential objects from the
+        This returns an instance of one of the *Credential objects from the
         `azure.identity` package.
         """
         return DefaultAzureCredential()
 
     def get_access_token(self, force_refresh: bool = False) -> str:
         """
-        Obtain an access token from the Azure identity provider.
+        Obtain an access token from the Azure identity provider.  The access token,
```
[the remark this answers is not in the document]
spencer-skylight: Good catch - updated.

phdi/cloud/azure.py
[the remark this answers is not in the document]
spencer-skylight: > I'm not sure how this approach helps us avoid keeping downloaded content in memory.

I converted the StringIO stream test case to a file-based test case `test_download_object_pass_file` to illustrate this.  This should cause the downloader to send the content directly to the file stream as it's being downloaded.

> I would say that this is not accurate. It avoids us needing to handle character encoding immediately, but eventually we'll need to convert those bytes into an object in order to interact with it, at which point we'll need to worry about the encoding.

I should have clarified this approach makes it possible for the caller to simply and efficiently handle encoding. I added a test case `test_download_object_pass_cp1251encoded` to prove this functionality works.  The content is downloaded as an encoding-agnostic byte stream, and a TextIOWrapper wraps that byte stream and interprets the file characters with the supplied encoding property. It also tests that attempting to read the same content as utf-8 does not work.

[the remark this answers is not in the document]
spencer-skylight: As discussed, we now return a string from `download_object`. I also added support for using a specified encoding.

phdi/cloud/azure.py
```diff
@@ -64,11 +67,131 @@ def _need_new_token(self) -> bool:
...
+
+    def download_object(
+        self, container_name: str, filename: str, encoding: str = "UTF-8"
+    ) -> str:
+        """
+        Downloads a character blob from storage and returns it as a string.
+
+        :param container_name: The name of the container containing object to download
+        :param filename: Location of file within Azure blob storage.
+        :param encoding: Encoding applied downloaded content.
```
bryan-skylight, line 124: typo: "Encoding applied downloaded content." -> "Encoding applied to the downloaded content."

nit: Some parameter descriptions end in bullets, but some don't. I'm indifferent to which one we use, but we should be consistent.

bryan-skylight (approved):
flourish of approval

phdi/cloud/azure.py
spencer-skylight: Updated.  I think periods should be used to end a sentence, but not a sentence fragment. I created a ticket to decide on this and update our documentation.  https://app.clickup.com/t/3cq92a1

PR #22, opened 2022-08-17
Fix GcpCredentialManager, upload_bundle_to_fhir_server, and associated functions

DanPaseltiner:
This PR addresses the following:

1. I got made a mistake yesterday with the GcpCredentialManager `valid` and `expired` are attributes not methods of the `Credentials` class. I should have caught this before opening the PR, but I think I got mixed up and testing the actual functionality v. unit testing. Anyway, I have confirmed that the GcpCredentialManager in this form works properly by using it to upload a FHIR bundle to a GCP FHIR Store.

2. We have changed the `get_access_token()` method on all CredentialManagers to simply return a string, but we had not updated accordingly in places that call this method.


spencer-skylight (approved):
This looks good! I should have caught the non-GCP changes when we started returning the raw access token from BaseCredentialManager.get_access_token().  Since we had to mock the return value for this in some of the tests, they still passed. I did a project search for `.token` and think you caught everything this time around.

PR #24, opened 2022-08-18
New taxonomy tabulation and tests

bamader:
This PR migrates the functionality of the original `schemas.py` file to the base (non-FHIR) functionality of our new taxonomy. Minor changes to function names, parameters, and docstrings are spread throughout.


phdi/tabulation/tables.py
```diff
@@ -0,0 +1,103 @@
...
+import pyarrow.parquet as pq
+import yaml
+
+from pathlib import Path
+from typing import Literal, List
+
+
+# @TODO: This function should also allow a user to load a JSON file
+# containing a schema; implement that functionality once the port
+# is complete.
```
spencer-skylight, line 14: I thought we had picked yaml over json.  While users could technically represent the schema either way, and both should be parseable into the same structure to drive our functionality, having all users manage their configurations in a consistent way seems beneficial.  Is there a reason we need to support both?

spencer-skylight (commented):
This looks like a pretty direct and straightforward port.  I had a question about the TODO, but otherwise it looks good!

phdi/tabulation/tables.py
bamader: Ah that makes sense. No, the only reason I said support both was that there was a sentence in either the PRD or SRD that said we should be able to support either yaml or json. Personally I agree with you that driving our users into the same format is desirable, so I'm happy to remove this todo.

bamader:
lol I did the same thing as you tend to do Spencer, I had a response comment typed up from yesterday but forgot to actually post it

spencer-skylight (approved):
LGTM!

phdi/tabulation/__init__.py
```diff
@@ -0,0 +1,3 @@
+from phdi.tabulation.tables import load_schema, write_table, print_schema_summary
```
bryan-skylight, line 1: I've noticed that in some modules we use relative imports (e.g. `from .core import BaseCredentialManager`), and in other modules, such as here, we use absolute imports (e.g. `from phdi.tabulation.tables import ...` instead of `from .tables import ...`). This isn't something that needs to be addressed here, but is something we should probably include in our design guide. 

PEP8 states that relative imports for intra-package imports are highly discouraged, even though PEP328 provides functionality for how this is done.

bryan-skylight (commented):
LGTM! This is a direct port of the code we had in our old repo, correct?

phdi/tabulation/__init__.py
bamader: Yeah I hear you. I've been using absolute imports wherever possible, and I didn't even realize we had any relatives out there. I would definitely be in favor of sticking strictly to absolute imports.

bamader:
@bryan-skylight it's almost a strict port of the older code, but there are a few very small updates (docstring changes for more clarity, removing use of the `logging` module and raising exceptions instead)

PR #26, opened 2022-08-19
Linkage tutorial

KennethSkylight:
This is a tutorial on how the linkage module work. Since it is relatively straightforward, the tutorial is short


tutorials/linkage-tutorial.md
```diff
@@ -0,0 +1,158 @@
+# Tutorials: The Linkage Module
+
+This guide serves as a tutorial overview of the functionality available in both `linkage.py` and `fhir.linkage.py`. It will cover concepts such as data type basics, imports, and common uses invocations.
+
+## The Basics: How it links
+
+The linkage module adds a patient identifier to each patient record in a bundle. If two hash strings are the same, that means that the two patient records are referring to the same patient. The hash string generation is a straightforward name and address combination.
```
bamader, line 7: Two things:
1. Don't we generate a hash string with name, date of birth, and address?
2. We refer to the generated string in the first sentence as a patient identifier and in the second as a hash string. Let's add a sentence clarifying that the patient identifier that gets added is the result of combining the three data fields, encoding the combination as a hash string, and adding a salt value. Just to make things explicit.

tutorials/linkage-tutorial.md
```diff
@@ -0,0 +1,158 @@
+# Tutorials: The Linkage Module
+
+This guide serves as a tutorial overview of the functionality available in both `linkage.py` and `fhir.linkage.py`. It will cover concepts such as data type basics, imports, and common uses invocations.
+
+## The Basics: How it links
+
+The linkage module adds a patient identifier to each patient record in a bundle. If two hash strings are the same, that means that the two patient records are referring to the same patient. The hash string generation is a straightforward name and address combination. 
+
+## Common Uses
+Listed below are several example use cases for employing the geospatial module.
```
bamader, line 10: Typo: this should be the linkage module

tutorials/linkage-tutorial.md
```diff
@@ -0,0 +1,158 @@
...
+
+To add patient identifiers to the bundle, you can use the `add_patient_identifier` function in the FHIR linkage package
+
+```
+from phdi.fhir.linkage.link import add_patient_identifier
+
+bundle = {...your bundle here...}
+salt = 'some-salt-string'
+
+add_patient_identifier(bundle, salt, True)
```
bamader, line 61: 1. I believe the `add_patient_identifier` function should still return a bundle. When overwrite is set to true, we just add the new information into the existing bundle, but we still return it so that the syntax matches when `overwrite=False`. So this should be `bundle = add_patient_identifier(bundle, salt, overwrite=True)`.
2. I think it's typically worthwhile to add a block of code underneath the example to mock up what the returned value is and display it. In this case, you could do something like

```
bundle = add_patient_identifier(bundle, salt)
print(bundle.get("whatever-path-the-identifier-has"))
>>> "string-value-of-the-generated-hash"
```

bamader (changes requested):
Close, just a couple questions/notes. 

1. Will your ticket for raw linkage affect any of the content in here? If so, we should insert some placeholders so we remember to come back and add that content here.
2. We need a README.md file in the linkage package(s), like we have for geospatial and harmonization. It doesn't have to be long, just a paragraph highlighting things and directing users here for more info.

tutorials/linkage-tutorial.md
KennethSkylight: I'm gonna add the output of the block underneath because the output would be the same for these False and True examples. 

I thought the point of the overwrite is for it to be a modifier, so you can just use your original variable you passed in?

tutorials/linkage-tutorial.md
KennethSkylight: Changed

tutorials/linkage-tutorial.md
KennethSkylight: Changed

tutorials/linkage-tutorial.md
```diff
@@ -0,0 +1,147 @@
+# Tutorials: The Linkage Module
+
+This guide serves as a tutorial overview of the functionality available in both `linkage.py` and `fhir.linkage.py`. It will cover concepts such as data type basics, imports, and common uses invocations.
```
bryan-skylight, line 3: When I read "data type basics", I interpret it as "learning the foundations of strings, integers, booleans, etc". What are trying to convey here, and is there a better way that we could achieve that?

tutorials/linkage-tutorial.md
bryan-skylight: You're both right about the `overwrite` parameter. Setting `overwrite` to `True` is intended to make it so that you don't have to keep setting the variable back to itself, or creating new variables. But Brandon is also right that setting `overwrite` to `True` returns the bundle. However, Kenneth's implementation is still acceptable. Consider the following simple example.

```python
# test.py
def f(x):
    return x
    
f(5)
y = f(3)
print(y)
```

If we call this from the command line with `python test.py` we get the following output:

```python
>>> 3
```

So the first function call, even though it returns a value, isn't shown in stdout. This is true in Jupyter notebooks as well. If a cell contains multiple function calls that all return a value, only the last function's return value is displayed if not explicitly shown with `print`. However, if `add_patient_identifer(..., overwrite=True)` is the only command run, then Brandon is right that the return value will get printed to screen unless saved to a variable.

With all of that said, I do think the tutorial should be more explicit about what's happening when `overwrite=True`.

bryan-skylight (changes requested):
This looks great! I'd like to see the comments from Brandon and myself about how `overwrite` affects the variables implemented before merging, but overall this is pretty much ready to go.

tutorials/linkage-tutorial.md
bamader: Yeah the intent of my comment wasn't that this syntactically / programmatically is incorrect (it's not, it totally works), it was that we as a team had made a decision to make return cases like this explicit. It goes back to our original pipeline and first HCD refactor where we prioritized "pass by reference" rather than modifying an object behind the scenes / inside a function. The return makes things explicit to the user and allowed chaining functions all operating on return bundles, which is something we liked.

Sorry, Kenneth, I should've made my "should" clearer and elaborated on that discussion, because it was before you'd joined the team!

tutorials/linkage-tutorial.md
KennethSkylight: Ah, that was copied over from the old version. I changed it to be 

"It will cover concepts such as using FHIR bundles for linkage, imports, and common use invocations."

Lemme know what you think

tutorials/linkage-tutorial.md
KennethSkylight: I added extra context to the overwrite = True. Let me know what you think

tutorials/linkage-tutorial.md
```diff
@@ -0,0 +1,147 @@
...
+
+bundle = {...your bundle here...}
+salt = 'some-salt-string'
+
+add_patient_identifier(bundle, salt, True)
+print(bundle)
+```
+
+Using the `True` tag in the third parameter overwrites your bundle with the new resource tag. This means that in the previous example, the `bundle` variable will be directly modified by add_patient_identifier. If using `False`, save the output to a variable. 
+```
```
bamader, line 66: This is true, but again I'd like to see some mention of how the modified input bundle will still be returned even if the user sets `overwrite` to `True`. If a user were to check the docstring and the function code proper, they'd see the return type is always `dict` and there is always a call to `return bundle`. I think it really is worth pointing out that design decision here because it influences the consistency with which the user can call the function (as well as avoids some what may seem to be a discrepancy between the tutorial and the docstring). This could be easily achieved with just a slight addition to what you've written here by adding another sentence: "...If using `False`, save the output to a variable. It's worth noting that `add_patient_identifier` will _always_ return a bundle of some form, even if `overwrite = True` (in this case, the modified input bundle will be returned). This not only maintains syntactic consistency between the two settings of `overwrite`, but makes explicit the modification happening to the input `bundle` dictionary. Moreover, this allows chaining the function with other building block function calls, such as `standardize_phones(add_patient_identifier(bundle, salt))`, for example."

bamader (changes requested):
Almost there. Just a 1-2 sentence addition I mentioned about the return value when `overwrite = True`.

tutorials/linkage-tutorial.md
KennethSkylight: Lemme know if the explanation works for you.

tutorials/linkage-tutorial.md
bamader: This seems okay for our purposes. Thanks for changing!

bamader (approved):

bryan-skylight (approved):
LGTM!

PR #27, opened 2022-08-22
Always return requests.Response from http_request_with_retry

bryan-skylight:
# Description
This PR links to [#3bp6kdk](#3bp6kdk) in ClickUp and implements new functionality to allow `http_request_with_retry` to always return a `requests.Response` object. The function as currently written supported returning a `requests.Response` object or a `NoneType` object. The rest of the codebase was insufficiently handling the latter case, and was assuming that the function always returned a Response object.

Expecting a Response object is likely the right way to go, and the new functionality allows Exceptions and custom messages, as well as custom HTTP response codes, to be crafted as a Response object, allowing the calling function to determine how it should process it.

# Feedback Requested
- Is everyone okay with how this is being handled?  
- Are there alternative methods to how we might want this done?

### DIFF
--- a/phdi/transport/http.py
+++ b/phdi/transport/http.py
@@ -1,34 +1,8 @@
 import requests
 
 from requests.adapters import HTTPAdapter
-from requests.models import Response
-from typing import List, Literal, Union
+from typing import List, Literal
 from urllib3 import Retry
-
-
-def _generate_custom_response(
-    url: str, status_code: int, content: Union[str, bytes]
-) -> requests.Response:
-    """
-    Returns a custom requests.Response object with relevant information about the
-    request and why it failed. The primary purpose of this functionality is to mock a
-    Response object in instances where the requests.request method fails or was not
-    executed in order to provide consistency in functionality that expects a Response
-    object to be returned.
-
-    :param url: The URL of the request that failed or was never run.
-    :param status_code: The HTTP status code that should be returned to the user.
-    :param content: Relevant information about why the original request failed or
-    wasn't run. This could include the Exception, the traceback info, or a custom
-    message. It should be passed as a string in the form of a dictionary
-    (e.g. '{"exception": "<class 'requests.exceptions.ConnectTimeout'>})')
-    """
-    response = Response()
-    response.url = url
-    response.status_code = status_code
-    # Response._content requires the text to be bytes instead of unicode
-    response._content = content.encode() if isinstance(content, str) else content
-    return response
 
 
 def http_request_with_retry(
@@ -40,9 +14,8 @@ def http_request_with_retry(
     data: dict = None,
 ) -> requests.Response:
     """
-    Carryout an HTTP Request using a specific retry strategy. Essentially
-    a wrapper function around the retry strategy implementation of a
-    mounted HTTP request.
+    Executes an HTTP request, retrying the request if the returned HTTP status code
+    is one of a specified list of codes.
 
     :param url: The url at which to make the HTTP request
     :param retry_count: The number of times to re-try the request, if the
@@ -55,9 +28,14 @@ def http_request_with_retry(
       including Authorization and content-type
     :param data: JSON data in the case that the request requires data to be
       posted. Defaults to none.
-    :raises ValueError: An unsupported request_type is passed
-    :return: A requests.Response object with the outcome of the http request
     """
+
+    request_type = request_type.upper()
+    if request_type not in ["GET", "POST"]:
+        raise ValueError(
+            f"The HTTP '{request_type}' method is not currently supported."
+        )
+
     # Configure the settings of the 'requests' session we'll make
     # the API call with
     retry_strategy = Retry(
@@ -70,31 +48,18 @@ def http_request_with_retry(
     http.mount("https://", adapter)
 
     # Now, actually try to complete the API request
+    # TODO: Condense this down to make a single call using
+    # http.request(method=request_type, url=url, headers=headers, json=data)
     if request_type == "POST":
-        try:
-            response = http.post(
-                url=url,
-                headers=headers,
-                json=data,
-            )
-        except Exception as e:
-            content = f'{{"message": "POST request to {url} failed with the following exception: {type(e)}"}}'  # noqa
-            response = _generate_custom_response(
-                url=url, status_code=500, content=content
-            )
+        response = http.post(
+            url=url,
+            headers=headers,
+            json=data,
+        )
     elif request_type == "GET":
-        try:
-            response = http.get(
-                url=url,
-                headers=headers,
-            )
-        except Exception:
-            content = f'{{"message": "GET request to {url} failed with the following exception: {type(e)}"}}'  # noqa
-            response = _generate_custom_response(
-                url=url, status_code=500, content=content
-            )
-    else:
-        content = f'{{"message": "The {request_type} HTTP method is not currently supported"}}'  # noqa
-        response = _generate_custom_response(url=url, status_code=400, content=content)
+        response = http.get(
+            url=url,
+            headers=headers,
+        )
 
     return response
--- a/phdi/transport/http.py
+++ b/phdi/transport/http.py
@@ -18,7 +18,7 @@ def _generate_custom_response(
 
     :param url: The URL of the request that failed or was never run.
     :param status_code: The HTTP status code that should be returned to the user.
-    :param content: Relevant information about what the original request failed or
+    :param content: Relevant information about why the original request failed or
     wasn't run. This could include the Exception, the traceback info, or a custom
     message. It should be passed as a string in the form of a dictionary
     (e.g. '{"exception": "<class 'requests.exceptions.ConnectTimeout'>})')
--- a/phdi/transport/http.py
+++ b/phdi/transport/http.py
@@ -28,6 +28,8 @@ def http_request_with_retry(
       including Authorization and content-type
     :param data: JSON data in the case that the request requires data to be
       posted. Defaults to none.
+    :raises ValueError: An unsupported HTTP method (e.g. PATCH, DELETE, etc) was passed
+    to the request_type parameter.
     """
 
     request_type = request_type.upper()
--- a/phdi/transport/http.py
+++ b/phdi/transport/http.py
@@ -1,9 +1,34 @@
-import logging
 import requests
 
 from requests.adapters import HTTPAdapter
-from typing import List, Literal
+from requests.models import Response
+from typing import List, Literal, Union
 from urllib3 import Retry
+
+
+def _generate_custom_response(
+    url: str, status_code: int, content: Union[str, bytes]
+) -> requests.Response:
+    """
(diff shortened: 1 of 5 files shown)


phdi/transport/http.py
```diff
@@ -1,11 +1,36 @@
...
 import requests
 
 from requests.adapters import HTTPAdapter
-from typing import List, Literal
+from requests.models import Response
+from typing import List, Literal, Union
 from urllib3 import Retry
 
 
+def _generate_custom_response(
```
spencer-skylight, line 9: This should technically work, but I'm curious to know why you decided to implement this as a response object rather than raise an exception? My feeling is that this is a good place to raise exceptions (or allow one to be raised) to the caller. Communicating the error as an HTTP response implies to me that it was a response returned by an HTTP web service which isn't actually the case. For comparison, `requests` returns a `Timeout` error rather than constructing a `Response` object when a connection or response wasn't received in some period of time.

tests/transport/test_transport.py
```diff
@@ -66,3 +68,20 @@ def test_http_request_with_retry_get(mock_retry_strategy, mock_get):
     )
 
     assert response == return_value
+
+
+def test_generate_custom_response():
```
spencer-skylight, line 73: I think we should probably add test cases to ensure this function is called for certain error conditions (requests.get throws an exception). I'm not sure we need to go as far as a unit test when request_type is not one of the values listed in the type declaration `Literal["GET","POST"]` although that case does have explicit handling.

phdi/transport/http.py
```diff
@@ -1,11 +1,36 @@
...
+    """
+    response = Response()
+    response.url = url
+    response.status_code = status_code
+    # Response._content requires the text to be bytes instead of unicode
+    response._content = content.encode() if isinstance(content, str) else content
+    return response
+
+
 def http_request_with_retry(
```
spencer-skylight, line 34: The docstring should be updated to describe the new behavior. This method hasn't been revised yet to remove the :return: directive, so describing the return behavior could be moved to the method description as well.

spencer-skylight (changes requested):
Nice identification of something that should be changed and initiative to make the update! Just a question about the approach, and minor suggestions.

phdi/transport/http.py
bryan-skylight: That's a great question. I certainly considered using exceptions, but ultimately the objective was to handle scenarios in which an HTTP request failed to complete, either because it was never issued (i.e. when `request_type not in ["GET", "POST"]`) or when it failed. Raising an exception for the case when the HTTP method hasn't been implemented didn't really make much sense to me. It's not that it's an error, it's just that we haven't implemented it yet. And raising an exception when the request failed to execute definitely does make sense, but if you look at how we've implemented all of our conditionals when we use the `http_request_with_retry` method, you'll see that we almost exclusively rely on the status code ([example 1](https://github.com/CDCgov/phdi/blob/9cc4df2f827511391e39a37a1cddc2645776ff0a/phdi/azure.py#L82), [example 2](https://github.com/CDCgov/phdi/blob/9cc4df2f827511391e39a37a1cddc2645776ff0a/phdi/fhir/__init__.py#L61), [example 3](https://github.com/CDCgov/phdi/blob/9cc4df2f827511391e39a37a1cddc2645776ff0a/phdi/fhir/__init__.py#L348)).

So, with those things in mind, I decided on this solution as it does a couple of things. First, it meets the other functions where they're at and allows us to implement conditionals that are contextually appropriate for network requests (i.e. checking status codes instead of implementing `try/except` clauses). Second, it gives us the ability to use those exceptions later on down the line if we so choose. It would take a little parsing, but raising an exception based on the text in the `content` parameter would certainly be doable. Plus, if we decide to go the Exceptions route instead of the status code route, we can always modify that later.

Lastly, on the note about how `requests` implements error handling, I think it makes sense that they'd pass the exception, and not a Response object holding the exception's type. The `requests` module is providing functionality for literally every use case that involves making an HTTP request. That spans domains, functionality, and tech expertise, and thus requires supporting different use cases and how they handle their exceptions. We, on the other hand, are writing code for a specific domain and much smaller spectrum of tech expertise. It's more likely that our users care simply that the request failed, not necessarily why (this is speculation, but is a reasonable enough assumption to not feel this implementation deprives them of anything).

phdi/transport/http.py
bryan-skylight: Good call. I'll update.

tests/transport/test_transport.py
bryan-skylight: Agreed. I'll add these tests and update.

phdi/transport/http.py
spencer-skylight: I'm a little hesitant to dive in to this, but I know you're not one to shy away from a spirited debate :smile: and I'm still not convinced this is the best approach so here goes...

To summarize my argument for using exceptions:
There wasn't really an HTTP response that the resulting `Response` object represents. The fact we don't have a response is the result of a failure, and exceptions are Python's way of communicating failure in this case.

There is also a question of precedent we are setting. If we decide not to raise an exception, are we:
* Avoiding exceptions as a rule or
* Avoiding in this special case - if so what general criteria should we use to decide when to avoid exceptions?


To address your comments:
A primary argument seems to be that these error scenarios should generally be handled the same as other errors (e.g. 400 returned by downstream web service) and it would be easier not to introduce additional exception handling in our code or ask users to do so. I'd argue that using an unsupported action is most likely something that users will want to fix in development - i.e. they can't use our function to do a PUT (currently). An exception seems logical in this case as compared with a data issue coming back from a web service that only occurs on some messages. More generally, it seems to me that we should assume our users can handle exceptions since they are a very commonly used feature of Python.

When you mention "use those exceptions down the line", do you mean in SDK code or the calling code? 
* In SDK code wouldn't we simply replace the custom response code? It seems fine to fall back on this if we need to, but should probably be considered a breaking change since users may need to update their error handling logic. 
* Converting the responses to exceptions in the calling code seems like an anti-pattern to me.

We could expect users to build logic around exception text. I've done this in the past when there was no other option, but it does have a few pitfalls. If the code is internationalized, that parsing logic would need to be language-specific. As a developer, I'd prefer to view changes to error descriptions as a non-breaking. This wouldn't be the case if the text was being parsed and used in error handling logic. On a related note, current code suppresses stack trace information (and potentially other contextual information) by swallowing the `requests` exception and returning a custom `Response` with a generic error code and text.

On the topic that PHDI implements only a portion of the HTTP standard, I don't think this should dissuade us from raising an exception. In the end, it's still an error, and we can indicate that it's not supported by our implementation in the error detail text to convey this. A similar concept is represented by Python's `NotImplementedError`, though I don't think it's correct for our purposes, based on the [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError).

Now that I've laid out my thoughts, I don't want to belabor this any more. If, after review, you are still unmoved then I'll support your approach.

tests/transport/test_transport.py
bryan-skylight: This is no longer relevant since we're not using this code anymore. Marking this as resolved.

phdi/transport/http.py
bamader: This was a delight to read.

I'm gonna go stare into the void and contemplate `Exceptions` for the next seven years, brb

bamader: TLDR: For what it's worth after thinking about this and trying to map out some of the ways this might manifest down the line, I think I agree with Spencer's intuition here about not wrapping the error in a custom response and instead raising an exception. I won't belabor the point (because he outlined them so well), but the drivers to me are

1. his summary defense: "the fact that we don't have a response is the result of an error, and exceptions are [essentially] the most 'Pythonic' way of communicating this to the caller"
2. the perils of building logic around exception text (I've done this on a past project as well and it is an _absolute nightmare_ when that text changes, especially for systems that employ multiple languages)
3. assuming some competency on the part of the user. I think this was actually a fairly under-sold point, because it's something that's consistently guided our design patterns around human-centering. My first attempt at the HCD refactor many moons ago assumed not enough competency on the part of the user, so we ended up assuming a standard of "familiar with Python, not an expert in software development." I think that statement gives a fairly key guiding indicator for the way to handle this situation, because Spencer's right that the `Exception` as an entity is probably Python's most natural way of communicating an error in result generation. I would expect user's at a "familiar" level of experience to not only recognize this fact, but be comfortable working with it. On the other hand, wrapping exception logic into a custom response and then requiring the caller to sift through it does seem simultaneously like more work than should be expected at the user's average level (because they're not a true SWE) and also an anti-pattern to the familiarity assumption we're making.

bamader (approved):
LGTM!

phdi/transport/http.py
```diff
@@ -30,9 +28,14 @@ def http_request_with_retry(
       including Authorization and content-type
     :param data: JSON data in the case that the request requires data to be
       posted. Defaults to none.
```
spencer-skylight, line 30: Perhaps keep this? It also may be worth documenting that this may raise errors that requests.request might raise. IMO it would be overkill to try and list them - the [requests documentation](https://requests.readthedocs.io/en/latest/api/) doesn't even do that.

phdi/transport/http.py
```diff
@@ -15,9 +14,8 @@ def http_request_with_retry(
     data: dict = None,
 ) -> requests.Response:
     """
-    Carryout an HTTP Request using a specific retry strategy. Essentially
-    a wrapper function around the retry strategy implementation of a
-    mounted HTTP request.
+    Executes an HTTP request, retrying the request if the returned HTTP status code
+    is one of a specified list of codes.
```
spencer-skylight, line 18: Maybe list the codes that will cause a retry?

spencer-skylight (approved):
Just 2 non-blocking suggestions.  Looks great!

phdi/transport/http.py
spencer-skylight: Thanks for the back and forth on this Bryan and Brandon! I learned a lot from your feedback, and enjoyed the spirited debate about the pros and cons of the different options.

phdi/transport/http.py
bryan-skylight: I intentionally opted to not list them because I have thoughts on how this function is implemented. The gist is that we ask the user to provide a list of allowed methods, but hardcode the status codes that should be retried, which doesn't make sense to me. We should hard code the methods for which we'd do the retry since the user is passing in a single request type anyway, and then allow the user to specify which status codes they want to do a retry on.

phdi/transport/http.py
bryan-skylight: I think this comment was intended to be below line 33? If so, I'd like leave this as a follow up task that applies to all of our code in which we as a team decide on what our guidelines should be for including exceptions. Should we only include those raised explicitly? Include those that could be raised implicitly? Only include the implicit ones that are _likely_ to occur (how do you determine that?). Once that's decided, we can update all the docstrings.

For now, I'll simply add line 33 back in since I do think we should document the exceptions we're explicitly raising.

PR #30, opened 2022-08-24
Create core, refactor ABCs

bamader:
This PR standardizes the geospatial module to follow the same structure as the others. AbstractBaseClasses have been moved to a `core.py` file and appropriately prefixed with `Base`, and `__init__.py` now defines the member functions that can be invoked in the public namespace. This PR also removes `:return:` statements from docstrings in the geospatial module, as we wish to remove these explicit callouts from the docstring and instead fold their content into the summary block.


phdi/fhir/geospatial/core.py
```diff
@@ -0,0 +1,61 @@
...
+
+        :param address: A FHIR formatted dictionary holding address fields
+        :param lat: The latitude to add to the FHIR data as an extension
+        :param long: The longitude to add to the FHIR data as an extension
+        """
+        if "extension" not in address:
+            address["extension"] = []
+        address["extension"].append(
+            {
+                "url": "http://hl7.org/fhir/StructureDefinition/geolocation",
```
KennethSkylight, line 49: Could we add a comment explaining why this URL specifically needs to be here?

KennethSkylight (approved):
Looks fine to me, just had one question

phdi/fhir/geospatial/core.py
bamader: Good call, I'll put that in now

PR #31, opened 2022-08-24
Linkage split

KennethSkylight:
Instead of having one linkage module that loops through a bundle, we are splitting out the loop and the resource level modifications. 

That way, a user can perform resource level linkage if they prefer that.


bamader (changes requested):
The idea of giving the user more flexibility for doing resource- or bundle-level work is definitely an important one, and we should get that change implemented for sure. However, in other areas of the code base, the way we've implemented this is by having functions that are input-agnostic (i.e. don't reference `by_bundle` or `by_resource` in their title) and just type-check the data they're given. I'd like to see this code follow that same convention so that we keep things standardized and aligned (check out `fhir/harmonization` for some examples on handling bundle vs resource. I think that type-checking logic and handling could largely be straight copied here).

tests/fhir/linkage/test_fhir_link.py
```diff
@@ -17,13 +20,13 @@ def test_missing_address():
...
 
-    add_patient_identifier(bundle, "some-salt")
+    add_patient_identifier_by_bundle(bundle, "some-salt")
     expected = generate_hash_str("doe-19900101-", "some-salt")
     actual = bundle["entry"][0]["resource"]["identifier"][0]["value"]
     assert actual == expected
 
 
-def test_add_patient_identifier():
+def test_add_patient_identifier_by_bundle():
```
spencer-skylight, line 29: Could you add a few test cases where `overwrite=False` for the 2 functions?

tutorials/linkage-tutorial.md
```diff
@@ -50,25 +50,25 @@ Suppose you had a FHIR bundle that had the following data
 }
 ```
 
-To add patient identifiers to the bundle, you can use the `add_patient_identifier` function in the FHIR linkage package
+To add patient identifiers to the bundle, you can use the `add_patient_identifier_by_bundle` function in the FHIR linkage package
```
spencer-skylight, line 53: Nit: First off, nice job catching the linkage documentation updates. Could we work in to the docs that add_patient_identifier also exists to add an identifier at the patient resource level? I don't think it needs an example - just a sentence mention.

spencer-skylight (commented):
Just a few nit-picky suggestions.

phdi/fhir/linkage/link.py
```diff
@@ -7,13 +7,13 @@
 from phdi.linkage.link import generate_hash_str
 
 
-def add_patient_identifier(bundle: dict, salt_str: str, overwrite: bool = True) -> dict:
+def add_patient_identifier_by_bundle(
+    bundle: dict, salt_str: str, overwrite: bool = True
+) -> dict:
```
bamader, line 12: Can we rename this to either `add_patient_identifier_to_bundle` or `add_patient_identifier_in_bundle` (your choice)? `_by_` isn't the right descriptor of this to me--it feels off, and while we are technically separating functions by resource and by bundle, that naming convention doesn't work because we're not calling the other function `by_resource` so the rationale is masked from the user. IMO, `_in_` or `_to_` are more descriptive, more appropriate for what's actually happening in the function (all patients _in_ the function are having IDs added _to_ them), and better matches the naming convention that we'll use for other split functions (such as `standardize_names_in_bundle` or `geocode_addresses_in_bundle`).

phdi/fhir/linkage/link.py
```diff
@@ -28,39 +28,60 @@ def add_patient_identifier(bundle: dict, salt_str: str, overwrite: bool = True)
         bundle = copy.deepcopy(bundle)
 
     for resource in find_resource_by_type(bundle, "Patient"):
-        patient = resource.get("resource")
```
bamader, line 31: This line should stay in the function. Because `resource` at this point isn't an actual FHIR-formatted resource (just a dictionary with a `resource` key), we shouldn't be calling it one or passing it into the `add_patient_identifier` function as a resource. Rather, we should actually get the patient resource here and then pass the extracted resource into the function.

phdi/fhir/linkage/link.py
```diff
@@ -28,39 +28,60 @@ def add_patient_identifier(bundle: dict, salt_str: str, overwrite: bool = True)
...
+        linking identifier
+    :param salt_str: The suffix string added to prevent being
+        able to reverse the hash into PII
+    :param overwrite: Whether to write the new standardizations
+        directly into the given bundle, changing the original data (True
+        is yes)
+    """
+    if not overwrite:
+        resource = copy.deepcopy(resource)
+    patient = resource.get("resource")
```
bamader, line 53: This line shouldn't need to exist within this function. The function should already be given a patient resource, rather than a dictionary holding a patient resource inside one of its keys. See comment above about moving the logic to the for-loop. This is also important to make sure that our functions behave the way users would expect them to, both from how we're naming them and using docstrings (which indicates that the input `resource` actually _is_ a FHIR resource) and from an intuition perspective.

phdi/fhir/linkage/__init__.py
```diff
@@ -1,3 +1,6 @@
-from .link import add_patient_identifier
+from phdi.fhir.linkage.link import (
+    add_patient_identifier_by_bundle,
+    add_patient_identifier,
+)
 
-__all__ = ["add_patient_identifier"]
+__all__ = ["add_patient_identifier_by_bundle", "add_patient_identifier"]
```
bamader, line 6: Will need changed to reflect name change (see comment in `linkage/link.py`)

tests/fhir/linkage/test_fhir_link.py
```diff
@@ -45,9 +48,39 @@ def test_add_patient_identifier():
...
+
+    expected_new_identifier = {
+        "value": generate_hash_str(plaintext, salt_str),
+        "system": "urn:ietf:rfc:3986",
+        "use": "temp",
+    }
+
+    for resource in incoming_bundle["entry"]:
+        if resource["resource"]["resourceType"] == "Patient":
+            add_patient_identifier(resource, salt_str, True)
```
bamader, line 83: Will need to be modified to receive the `resource.get("resource")` extraction

tutorials/linkage-tutorial.md
bamader, line 53: Just marking in the one place, but will need to be modified to reflect the changed name

bamader (changes requested):
We're almost there, just two changes (and ripples of those changes in a couple places) requested:
1. more appropriate name for the bundle function
2. making the `add_identifier` function (for resources) actually accept a resource, rather than a dict with a `resource` key

tutorials/linkage-tutorial.md
KennethSkylight: Added

phdi/fhir/linkage/link.py
KennethSkylight: changed

phdi/fhir/linkage/link.py
KennethSkylight: nice catch, I also changed the name per Spencer's comment

phdi/fhir/linkage/link.py
KennethSkylight: Changed

phdi/fhir/linkage/__init__.py
KennethSkylight: changed

tests/fhir/linkage/test_fhir_link.py
KennethSkylight: changed

tutorials/linkage-tutorial.md
KennethSkylight: changed

tests/fhir/linkage/test_fhir_link.py
KennethSkylight: Spencer and I talked and we are moving this to a different PR

bamader (approved):
LGTM! If the decision to open those other test cases is going to be for another PR, let's get a ticket created marking that as a fast follows item so that we don't lose track of that work.

spencer-skylight (approved):
Looks good!

I added the test case documentation to an existing ticket we had to update unit tests:
https://app.clickup.com/t/3g9pbgz

PR #32, opened 2022-08-24
Transport module fhir export

spencer-skylight:
# Description
Migrate export-related functionality from the old FHIR module to the new taxonomy in accordance with the [PRD](https://docs.google.com/document/d/1Jzc8KG6idkap6njacoD_Cdqm5fN03LTYslIEhtK9UL8/edit).

In addition to the taxonomy migration, the following changes were made:
* In `phdi.fhir.cloud.azure` function `download_from_export_response` (and helper), expect an AzureCredentialManager to be supplied by the caller, and use it to log in. (and unit tests updated accordingly)
* Minor updates to docstrings.
* Removed logging calls
  * Debug and info
  * Exception logging where we are also raising an exception


phdi/fhir/cloud/azure.py
```diff
@@ -0,0 +1,54 @@
...
+
+from azure.storage.blob import download_blob_from_url
+from typing import Iterator, TextIO, Tuple
+
+from phdi.cloud.azure import AzureCredentialManager
+
+
+def download_from_fhir_export_response(
+    export_response: dict,
+    cred_manager: AzureCredentialManager,
```
bamader, line 11: Since we have a `BaseCredentialManager` and are moving away from defining just the Azure specific implementation, can we make this the base class instead? And for that matter, how much of a lift would it be to make this code generic for anything with a credential manager? Looking at some of the below functions, nothing seems particularly specific to Azure (since we just invoke a common member function `.get_credential_object()` on line 47), so could we just rename this whole file something more suitably generic?

phdi/fhir/transport/__init__.py
```diff
@@ -4,8 +4,11 @@
     upload_bundle_to_fhir_server,
 )
 
+from .export import export_from_fhir_server
```
bamader, line 7: Absolute path

phdi/fhir/transport/export.py
```diff
@@ -0,0 +1,201 @@
+import polling
+import requests
+
+from .http import http_request_with_reauth
```
bamader, line 4: absolute import

phdi/fhir/transport/export.py
```diff
@@ -0,0 +1,201 @@
...
+        separator = "&"
+
+    if container:
+        export_url += f"{separator}_container={container}"
+        separator = "&"
+
+    return export_url
+
+
+def __export_from_fhir_server_poll_call(
```
bamader, line 131: Can we make this one underscore in front of the function instead of two? I know it's a helper for other helpers but let's standardize it so that underscore functions lead with only one.

phdi/fhir/transport/export.py
```diff
@@ -0,0 +1,201 @@
...
+            raise requests.HTTPError(response=poll_response)
+
+
+def _compose_export_url(
+    fhir_url: str,
+    export_scope: str = "",
+    since: str = "",
+    resource_type: str = "",
+    container: str = "",
+) -> str:
```
bamader, line 91: Small organizational nit: let's put all the functions with underscores at the bottom of the file, so that all the publicly exposed functions a user can call and might want to read more about are at the top of the file in one group and all the private helpers are all together in another group below that.

bamader (changes requested):
Looks like a good port, just a couple of formatting / organizational nits to take care of before merging!

phdi/cloud/azure.py
```diff
@@ -79,6 +78,7 @@ def _need_new_token(self) -> bool:
 class AzureCloudContainerConnection(BaseCloudContainerConnection):
     """
     This class implements the PHDI cloud storage interface for connecting to Azure.
+
```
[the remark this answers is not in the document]
spencer-skylight: Nope - I'll remove.

phdi/fhir/cloud/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .azure import download_from_fhir_export_response
```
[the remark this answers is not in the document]
spencer-skylight: Updated.

phdi/fhir/cloud/azure.py
spencer-skylight: If this were itself defined in, or implementing a base class, I'd agree.  But since this is an Azure-specific function and it requires the return value of `get_credential_object()` to be an Azure credential manager not just any "credential object", I think putting the more specific type here makes sense.  

Does that explanation sway you?  Is it OK to leave this as is?

phdi/fhir/transport/__init__.py
spencer-skylight: I'll update this.

phdi/fhir/transport/export.py
spencer-skylight: Will update.

phdi/fhir/transport/export.py
spencer-skylight: Sure, I can do that. We don't have any guidance around when to use 1 vs 2. I'll keep it standard unless/until we do.

phdi/fhir/transport/export.py
spencer-skylight: Sure, I'll update.

phdi/fhir/cloud/azure.py
bamader: Ah that makes sense. Yes I think this is totally fine to leave as-is then. Thanks for clarifying.

bamader (approved):
LGTM!

PR #33, opened 2022-08-25
Document coverage evaluation

bamader:
This PR documents how the `coverage.py` package can be used to measure and evaluate our code coverage, as well as explore specific areas that may be deficient in terms of coverage. The information is currently stored in `CONTRIBUTING.md` because that's where we discuss running the tests, but if that isn't the right place for this I'm happy to move it.

The results of our initial coverage assessment are really good. We have overall 95% statement coverage, with the only noteworthy under-covered group being our cloud functionality (`azure.py`, `cloud.core.py`, etc.). That code is relatively new in port, though, so those tests are likely coming (or we still need to write them). Among modules that were already ported with their tests, we score horrendously on `fhir.transport.http.py`, which has only 36% coverage (the single lowest coverage score we have--everything else is around at least 80%, with most being in the 90s).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,11 +112,21 @@ poetry run make test
 
 If that fails, stating a file cannot be found, you can also try running `poetry run pytest` directly to run the tests.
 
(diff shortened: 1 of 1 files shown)


KennethSkylight (approved):
LGTM

PR #35, opened 2022-08-30
Cleanup old taxonomy files

bamader:
This PR removes all straggling modules and functions from the old taxonomy, since everything has been ported over to the new code structure. The chief areas of rewiring code were in configuring some of the more subtle imports to point to new ports instead of the old structure (e.g. `fhir_server_get` is not in `fhir.transport.http` rather than the old standalone `fhir.py`), and changing the objects that `unittest.mock` was mocking to point to new taxonomy instances.

I know that deletions are typically easy PRs to approve, but since this clears out old files from the repo, **everyone please try to give it an attentive read so we don't miss anything!**


spencer-skylight (approved):
LGTM! Only suggestion I have is to let Dan, Nick and Brady know and make sure deleting the old taxonomy won't break anything they're working on. :)

PR #36, opened 2022-08-31
Add coverage guidelines

bamader:
This PR adds coverage guidelines to the `contributing.md` file based on the robustness conversation the SDK workstream had earlier this week. Principles are suggested for when tests are appropriate, what cases they should be validating, and howe we use and regard the resulting coverage percentage.


bamader:
Ann doesn't have a Github account I can request review from, but it might be worth getting her eyes on this as well since robustness assurances are something that could impact usability or comms to STLTs.

CONTRIBUTING.md
```diff
@@ -128,6 +128,23 @@ You can also create an HTML-formatted and -tagged report by running `coverage ht
 
 More information can be found in the documentation for [coverage](https://coverage.readthedocs.io/en/6.4.4/).
 
+##### Code coverage guidelines
+
+We recognize there's a lot of human judgment when it comes to writing test cases. It's impossible to plan for every scenario in which a function might be used, or think of every error that an input data type might cause. To ensure the repository is as robust as reasonably possible, we've developed the following principles around writing tests and code coverage.
+
+1. **We strive for 80% repository coverage at all times.** There's no magic number when it comes to assessing coverage quality. A high coverage percentage doesn't actually guarantee high-quality testing, and focusing too much on getting a coverage number as close to 100% as possible could lead to a false sense of security as well as wasted time and effort. At the same time, it's important we ensure that our tools are tested in a robust way. A coverage statistic of 80% strikes a balance between fixating on the coverage percent and keeping tests robust and accountable.
```
spencer-skylight, line 135: What do you think about striving for 85%?  We are at 95% now, and may go higher when we revise our test cases.  If we start dipping near 85% we can always lower, but 85 seems like a good time to start discussing and revising.

CONTRIBUTING.md
```diff
@@ -128,6 +128,23 @@ You can also create an HTML-formatted and -tagged report by running `coverage ht
...
+
+2. **Any public-facing functions should include at least one test case.** If a user can call function, we want that function to have documented test coverage. Ideally, functions should have multiple tests (see below), but at a minimum, any public function should have at least one. For private or helper functions, discretion is given to the developer, but the principles below might suggest test cases for those, too.
```
spencer-skylight, line 137: I researched this yesterday and couldn't find an easy way to check this in an automated way.  I'm thinking we should group this with the other guidelines in point 4 below.  Does that make sense to you as well?

spencer-skylight (changes requested):
A few suggestions for consideration.  I think I'd be OK if this was merged as-is so I'm approving, but please consider the comments I left as well.

CONTRIBUTING.md
bamader: Sure, I've got no problem with 85% (I think that's what I threw out in our meeting anyway). I feel like I just went with 80% because I read the extra credit article and mentally was like "well let's just take the average between a perfect 100 and the 60 that google uses" lol

CONTRIBUTING.md
bamader: Looking at the way I've organized this, I actually think this content belongs where it is. I was trying to develop an organizational scheme where the first three points were our sort of "governing principles," if you will, which were more like abstract ideas and concepts we were striving to uphold when writing tests and considering coverage. Point 4, on the other hand, is intended to be a more specific "actionable" point that helps developers understand what practical / real tests to be writing. In reading it again, that organization comes across to me but I think only because I was the one who wrote it (i.e. I didn't actually explicitly call that out). If you agree with the organization, I could separate point 4 out into a new paragraph rather than the bulleted list and add a sentence about how these are our practical and actionable test guidelines?

spencer-skylight: Sure sounds good to me to keep this separated out.  I like point 4 as a bullet list - I think it's easier to scan/read that way.  Maybe you could just add a short sentence above the bullet list?

spencer-skylight (approved):
I think we've worked out everything... Approving!

CONTRIBUTING.md
bamader: Done!

PR #38, opened 2022-09-01
Handle missing address country field exception

bamader:
This PR fixes the bug that was identified in https://app.clickup.com/t/3d2h5bk by only operating on and appending country codes that are not `None` (previously, FHIR patients whose address dicts didn't include a `country` field were throwing errors because `None` couldn't be parsed into an ISO country code). It also introduces a test case into our unit tests that checks for this input edge case to prevent problems in the future.


spencer-skylight (commented):
LGTM!  Just trying to think of other edge cases... What happens if country is an empty string or a nonsense/unrecognized country?

bamader:
Both of those are actually edge cases of `phdi.harmonization.standardization.standardize_country_code` rather than edge cases of the `standardize_phones` function (the reason for that is because if the `country` field exists, we'll still try to append a standardized code to a list of countries, see line 179 in `fhir.harmonization.standardization`). The good news is that our code as written already accounts / works for those! If no valid code can be produced for a given input country in some form, we always just return `None`. This is actually totally fine because when it comes to actually trying to parse the phone number with a country code in `harmonization.standardization.standardize_phone` (lines 88-100 are the relevant ones), not only is `None` a valid parsing option, but we literally manually insert `None` if it's not already in the list of possible countries. So we're actually all set on that front. To prove this out I did add two test assertions for code standardization though, which should now be live in the PR as well.

spencer-skylight (approved):
LGTM!

PR #42, opened 2022-09-07
rename find_resource_by_type to find_entry_by_resource_type

m-goggins:
Renamed find_resource_by_type to find_entry_by_resource_type to better articulate the function returning a list of resources within entries by resource type.

Newbie question: In the FHIR documentation it says that an entry can have a resource _or information_. I haven't been able to find any examples of bundles without resources but could this function error on an entry in a bundle that only has information (and therefore no resourceType)?

From https://build.fhir.org/bundle.html:
`"[entry](https://build.fhir.org/bundle-definitions.html#Bundle.entry)" : [{ // [Entry in the bundle - will have a resource or information](https://build.fhir.org/terminologies.html#unbound)`

### DIFF
--- a/tests/assets/cp1252-sample.txt
+++ b/tests/assets/cp1252-sample.txt
@@ -1,4 +1,1 @@
-Testing windows-1252 encoding
-€
-œ
-Ÿ
+Testing windows-1252 encoding € œ Ÿ
--- a/tests/cloud/test_cloud.py
+++ b/tests/cloud/test_cloud.py
@@ -363,8 +363,6 @@ def test_download_object_cp1252(mock_get_client):
 
     assert download_content_buffer.readline() == "Testing windows-1252 encoding € œ Ÿ"
 
-    # assert download_content == "Testing windows-1252 encoding\n€\nœ\nŸ\n"
-
     with pytest.raises(UnicodeDecodeError):
         object_content.decode("utf-8")
 
--- a/tests/cloud/test_cloud.py
+++ b/tests/cloud/test_cloud.py
@@ -1,6 +1,7 @@
 import json
 import pathlib
 import pytest
+import io
 
 from datetime import datetime, timezone
 from unittest import mock
@@ -349,7 +350,11 @@ def test_download_object_cp1252(mock_get_client):
         object_container, object_path, "cp1252"
     )
 
-    assert download_content == "Testing windows-1252 encoding\n€\nœ\nŸ\n"
+    download_content_buffer = io.StringIO(download_content)
+
+    assert download_content_buffer.readline() == "Testing windows-1252 encoding € œ Ÿ"
+
+    # assert download_content == "Testing windows-1252 encoding\n€\nœ\nŸ\n"
 
     with pytest.raises(UnicodeDecodeError):
         object_content.decode("utf-8")
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,7 +24,7 @@ jobs:
           cache: pip
       - name: Install poetry
         run: |
-          curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
+          curl -sSL https://install.python-poetry.org | python3 -
           echo "${HOME}/.poetry/bin" >> $GITHUB_PATH
       - name: Install dependencies
         run: poetry install
--- a/phdi/fhir/linkage/link.py
+++ b/phdi/fhir/linkage/link.py
@@ -1,6 +1,6 @@
 import copy
 from phdi.fhir.utils import (
-    find_entry_by_resource_type,
+    find_entries_by_resource_type,
     get_field,
     get_one_line_address,
 )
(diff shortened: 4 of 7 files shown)


bamader (changes requested):
I like this change, good call to add it in. Could we actually make the function named `find_entries_by_resource_type` (pluralizing entry)? Reading the change, I'm realizing we should've made it plural a long time ago since it returns a list, so if we're going to do a name refactor we might as well correct that now.

bamader:
Regarding your newbie question in FHIR structure: this is a good question, but thankfully I think one that doesn't apply to us. FHIR bundles come in a couple different flavors, but the ones we're concerned with are "Batch" and "Transaction" types. These types of bundles will always have a resource list as their entries, because by construction and their usage, they're intended to ferry data to and from FHIR storage / FHIR interaction. So I think by virtue of what we're using the bundles for, we're safe from the error hypothetical you mention.

spencer-skylight (approved):
This looks good to me pending Brandon's function renaming suggestion.

I agree with Brandon that we shouldn't expect to process bundles that don't have a resource.  With that said, we could make this a little more tolerant of that case.  I think you could change:
```python
if entry.get("resource").get("resourceType") == resource_type
```

to
```python
if entry.get("resource",{}).get("resourceType") == resource_type
```

This way, if there's no resource, it will default to return an empty dict and fail the resourceType check without throwing an `AttributeError`.

I created a ticket for this: https://app.clickup.com/t/3h0hvhc which can be taken on separately without holding up this PR.

m-goggins:
Thanks all! Good call on the entry vs. entries. I figured you all had some naming convention but glad it is more aligned with the returned values.

I also added the empty dict return option for bundles without resources since it was an easy fix. 

Also, should I be linking the click up tickets to the PRs? I tried using click up's github integration but there was no option for this repo?

spencer-skylight (changes requested):
One more minor request - could you add a test case for the new change in functionality for the default dict?  I think it should be a matter of updating the `patient_bundle.json` asset to contain an entry without a resource.  

Maybe just add a "request" property instead something like:
{
  "request": {
    "method": "GET",
    "url": "something"
  }
}

TBH, I'm not sure what a valid FHIR bundle would look like without a resource, but this should test for a missing resource anyhow.

tests/assets/cp1252-sample.txt
```diff
@@ -1,4 +1 @@
-Testing windows-1252 encoding
-�
-�
-�
+Testing windows-1252 encoding � � �
```
bamader, line 1: Why this change?

bamader (commented):
Pending Spencer's suggestion to add a modified test case, this looks okay. It looks like there might be some weirdness going on with `cp1232-sample.txt` which I'm not sure what to make of.

tests/assets/cp1252-sample.txt
spencer-skylight: I can answer that since I was the original author of this file and worked with Marcelle on this change yesterday... 

The newlines on Marcelle's Windows machine were causing problems because Git was automatically converting the file to contain Windows newlines (\r\n) when downloading the sample file to her repo.  This shouldn't actually be a problem when actually downloading a file - it's more an artifact of how Git works with newlines in a windows environment.  So to sidestep the issue, we just removed the newlines.  The character encoding testing is still valid which is what this file was created to test.

spencer-skylight (approved):
LGTM!

tests/assets/cp1252-sample.txt
bamader: That's wild 😲 . Little differences like that amaze me every time they pop up. This one sounds real ugly though so I'm glad you two found a workaround.

tests/cloud/test_cloud.py
```diff
@@ -349,7 +350,11 @@ def test_download_object_cp1252(mock_get_client):
         object_container, object_path, "cp1252"
     )
 
-    assert download_content == "Testing windows-1252 encoding\n€\nœ\nŸ\n"
+    download_content_buffer = io.StringIO(download_content)
+
+    assert download_content_buffer.readline() == "Testing windows-1252 encoding € œ Ÿ"
+
+    # assert download_content == "Testing windows-1252 encoding\n€\nœ\nŸ\n"
```
bamader, line 357: Since we no longer need this, let's delete the commented out code

bamader (changes requested):
Sorry, just one little thing left (deleting the commented out line of code)!

tests/cloud/test_cloud.py
m-goggins: Good catch

bamader (approved):
Good to go!

m-goggins:
I started adding in :param returns: to all the functions. It's definitely a good way to familiarize myself with the scripts. I wanted to check in about the formatting that I had completed so far (mainly just removing periods at the end of params). I'm not totally sure what is happening in the cloud/ script files and could use an overview, as well as the apply_selection_criteria() function in fhir/tables.py

PR #45, opened 2022-09-08
Lots of fixes for coverage quality

bamader:
This PR started out with these two tickets: https://app.clickup.com/t/3g9parn and https://app.clickup.com/t/3g9pakh (i.e. bringing the geospatial module and its FHIR wrapper up to snuff in terms of our test coverage standards). In doing so, I ended up finding several other things that needed some fixing, including some bugs we might not have noticed and that our updates to testing coverage actually helped catch:

- adds inline exclusion tags for abstract methods and properties of ABCs, since unimplemented functions shouldn't count against coverage trace stats (read more about this in the changes to CONTRIBUTING.md)
    - this is in both geospatial and cloud, since those are currently our two ABCs
- changes some incorrect data typing of bundles in a function signature
- expands the data structure dealt with in geospatial testing to include a second delivery line for proper response parsing
- expands unit test coverage for fhir/geospatial because we didn't have test cases for when a user didn't want to overwrite incoming data
- determined because of the tests that we were importing the wrong `copy` function in several places, so that's been fixed


phdi/fhir/geospatial/smarty.py
```diff
@@ -74,7 +73,7 @@ def _geocode_patient_resource(self, patient: dict) -> None:
                     address, standardized_address.lat, standardized_address.lng
                 )
 
-    def geocode_bundle(self, bundle: List[dict], overwrite=True) -> List[dict]:
+    def geocode_bundle(self, bundle: dict, overwrite=True) -> dict:
```
spencer-skylight, line 76: Nice catch here.

spencer-skylight (commented):
This all looks good to me. It looks like this is a draft so I didn't really look for things that hadn't been done yet that you might still be working on - I'll finish that part of the review when this comes out of draft mode.

phdi/fhir/geospatial/smarty.py
```diff
@@ -1,4 +1,4 @@
-from copy import copy
+import copy
```
m-goggins, line 1: Good call on this. I ran into errors in some of the harmonization tests (particularly when overwrite=False) when we used the `from copy import copy` approach.

m-goggins (approved):
The changes look good to me!

PR #46, opened 2022-09-09
Add :return param: to docstrings

m-goggins:
Started adding :param return: items to all functions with docstrings but wanted someone to take a look at general formatting (e.g.,) removed periods from docstring params) before proceeding. Some of the docstring descriptions and return params are repetitive but for the sake of consistency, I think it's okay.

### DIFF
--- a/phdi/cloud/azure.py
+++ b/phdi/cloud/azure.py
@@ -31,6 +31,7 @@ class AzureCredentialManager(BaseCredentialManager):
 
         :param resource_location: URL or other location of the requested resource.
         :param scope: A space-delimited list of scopes to limit access to resource.
+        :param return:
         """
         self.__resource_location = resource_location
         self.__scope = scope
@@ -54,6 +55,7 @@ class AzureCredentialManager(BaseCredentialManager):
 
         :param force_refresh: force token refresh even if the current token is
           still valid
+        :param return: An access token from the Azure identity provider
         """
         if force_refresh or (self.access_token is None) or self._need_new_token():
             creds = self.get_credential_object()
@@ -65,7 +67,7 @@ class AzureCredentialManager(BaseCredentialManager):
         """
         Determine whether the token already stored for this object can be reused,
         or if it needs to be re-requested.
-
+        :param return: An access token from the Azure identity provider if still valid
         """
         try:
             current_time_utc = datetime.now(timezone.utc).timestamp()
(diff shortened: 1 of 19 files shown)


spencer-skylight (commented):
It's coming along! Matt Goldberg created an API style guide that I think we can leverage here:

https://docs.google.com/document/d/17PSrnDf1bf-mbbF2utV4aboodgBr8gjDaDLVotcQGYQ/edit?pli=1#heading=h.jenb7unpjy0h

I noticed a few things that we need to fix across the board:
* `:param return:` should actually be `:return:` per [Sphinx documentation](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html) - a quick find-and-replace should take care of this.
* Descriptions for getters should start with "Gets the...". Then choose as concise as you can think of for the description for the `:return:`, per the API style guide.
* Functions that do not return anything do not need an empty `:return:` value.

I'm still getting through this PR, but need to step away briefly, and there's enough here to make some progress on corrections.

phdi/tabulation/tables.py
```diff
@@ -16,6 +16,7 @@ def load_schema(path: str) -> dict:
     If the file can't be found, raises an error.
 
     :param path: File path to a YAML file holding a schema
+    :param return:
```
spencer-skylight, line 19: "A dictionary representing read from the supplied `path`."

spencer-skylight (changes requested):

phdi/tabulation/tables.py
m-goggins: Would "A dictionary representing read **schema** from the supplied path" be okay?

spencer-skylight: Yep, you're right I was missing that key term. I think your wording is correct, but the following reads more naturally to me. I'm OK either way.
"A dictionary representing a schema read from the supplied path"

m-goggins:
Not sure if the `return writer` statement on line 58 in phdi/tabulation/tables.py is supposed to be there

phdi/cloud/azure.py
```diff
@@ -49,11 +49,12 @@ def get_credential_object(self) -> object:
 
     def get_access_token(self, force_refresh: bool = False) -> str:
         """
-        Obtain an access token from the Azure identity provider.  Return the
+        Obtain an access token from the Azure identity provider. Return the
         access token string, refreshed if expired or force_refresh is specified.
 
         :param force_refresh: force token refresh even if the current token is
           still valid
+        :return: True if Azure access token is valid; false otherwise
```
spencer-skylight, line 57: This function does actually return the access token.

m-goggins: Changed to "Azure access token"

spencer-skylight (approved):
This is looking great - I made a few changes to clean things up. I think it's GTG!

PR #50, opened 2022-09-13
Code coverage harmonization

m-goggins:
- Up to 100% coverage for fhir and core harmonization scripts
- Added an additional test hl7 file with malformed/invalid segment_id fields
- Changed the import of copy to remove an error related to using copy.deepcopy() in the case of an invalid segment_id field


tests/harmonization/test_harmonization.py
```diff
@@ -18,6 +18,11 @@ def test_standardize_hl7_datetimes():
     message_2 = open(
         pathlib.Path(__file__).parent.parent / "assets" / "FileSingleMessageLongTZ.hl7"
     ).read()
+    message_3 = open(
```
KennethSkylight, line 21: could we make the message variable names more descriptive of what they're testing for like `message_invalid_segments`?

KennethSkylight (changes requested):
One small thing

tests/harmonization/test_harmonization.py
m-goggins: Yeah, good point! I saw more descriptive names in another test file later but forgot to change these ones.

KennethSkylight (approved):
LGTM!

PR #52, opened 2022-09-14
Transport code coverage

spencer-skylight:
# Description
Ensure full code coverage for the transport and transport FHIR module.

## Additional Changes
Two lines of code that raise exceptions in `export_from_fhir_server_poll` and `export_from_fhir_server` were removed because any non-success status would raise an exception in `_export_from_fhir_server_poll_call`, meaning there are no code paths available to raise the errors in the former 2 functions.


tests/fhir/transport/test_fhir_transport.py
```diff
@@ -90,6 +97,148 @@ def test_auth_retry_double_fail(patched_requests_session):
...
+@mock.patch("phdi.fhir.transport.http._log_fhir_server_error")
+def test_upload_bundle_to_fhir_server_embedded_failure(
+    patch_log_error, patch_http_request
+):
+    bundle = {
+        "resourceType": "Bundle",
+        "entry": [{"resource": {"resourceType": "Patient"}}],
+    }
+
+    mock_response = {
```
m-goggins, line 163: Seeing these mock_responses in the test cases makes it more readable to me than some of the mock cases in other test files.

m-goggins: For my own edification: do we have a rule of thumb for when we should be using the mock responses vs. local vars vs. stored assets?

spencer-skylight: We don't currently have guidelines standards around this specific aspect of unit testing.  IMO setting content blocks separately (in local variables or parameters defined outside the function) to check is clearer/better than specifying literal values as parameters or directly checking literal values. I think we could make better use of module parameters to define mock content to separate content blocks from code and keep things cleaner.

Stored assets are generally used when the content is too big and would make the unit test messy or difficult to read if it were placed directly within the code.

m-goggins (approved):

PR #53, opened 2022-09-14
Fix test_generate_table_fail

bryan-skylight:
# Description
This test relied on an assertion that a function, `write_table()` was never called, with the assumption that it wasn't called because the HTTP  status code from the request was 400 instead of 200. The reason the test passed is because `schema` was set to an empty dictionary, which bypassed the entire `for` loop in `generate_table`. As a result, the function was indeed never called, but not for the reason expected.

# Solution
1. Read in an actual schema file from the `assets` library.
2. Add an additional assertion that checks that `fhir_server_get` was actually called, ensuring that the reason `write_table` isn't called is because of the returned `status_code`.

# Test Plan
The following image should demonstrate that this fix is achieving the results we need.

image


m-goggins:
Aha! That makes sense to me.

m-goggins (approved):

bamader (approved):

PR #59, opened 2022-09-15
Conversion code coverage

spencer-skylight:
# Description
Flesh out code coverage for conversion module

## Additional Changes
Removed code that rethrows a ParseException, for two reasons:
* Looking at the [python-hl7 code](https://github.com/johnpaulett/python-hl7), `ParseException`s are only thrown processing batch messages, and since we are checking that the input begins with `MSH`, there is no path to this through our code.
* We can just let exceptions be raised by the hl7 module - there is no need to re-raise them

### DIFF
--- a/phdi/fhir/conversion/convert.py
+++ b/phdi/fhir/conversion/convert.py
@@ -116,34 +116,29 @@ def _get_fhir_conversion_settings(message: str, use_default_ccda=False) -> dict:
     """
     # Some streams (e.g. ELR, VXU) are HL7v2 encoded
     if message[:3] == "MSH":
-        try:
-            parsed_msg = hl7.parse(message)
-            extracted_code = str(parsed_msg.segment("MSH")[9])
+        parsed_msg = hl7.parse(message)
+        extracted_code = str(parsed_msg.segment("MSH")[9])
 
-            # HL7 MSH segment 9 has three components: message code, trigger
-            # event, and message structure. We can extract based on number of
-            # present separators and recombine to create a robust formatted code
-            extracted_code_tokenized = extracted_code.split(parsed_msg.separators[3])
-            formatted_code = ""
-            if (len(extracted_code_tokenized) >= 3) and (
-                extracted_code_tokenized[2] != ""
-            ):
-                formatted_code = extracted_code_tokenized[2]
-            elif len(extracted_code_tokenized) == 2:
-                formatted_code = (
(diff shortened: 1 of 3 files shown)


phdi/fhir/conversion/convert.py
```diff
@@ -116,34 +116,29 @@ def _get_fhir_conversion_settings(message: str, use_default_ccda=False) -> dict:
...
-                raise Exception("Could not determine HL7 message structure")
-
-            return {
-                "root_template": formatted_code,
-                "input_data_type": "HL7v2",
-                "template_collection": "microsofthealth/fhirconverter:default",
-            }
-        except hl7.exceptions.ParseException:
-            raise hl7.exceptions.ParseException("Input HL7 message could not be parsed")
+        parsed_msg = hl7.parse(message)
```
m-goggins, line 119: It makes sense to me that we would remove a try/except from such a large chunk of code, but I am wondering what happens if we are unable to parse the message.

m-goggins (approved):

phdi/fhir/conversion/convert.py
spencer-skylight: Previously, if `hl7.parse` raised a `ParseException`, we'd catch it and create a new exception of the same type.  The stack trace for the new exception would show the exception as originating from our function, and replace whatever the error description was from the `hl7` package with "Input HL7 message could not be parsed". Any other types of exceptions thrown by the `hl7` package would bubble up, and retain their stack trace and error detail from where they actually occurred.

With this change, we would bubble up ALL errors that were raised during the function calls we made in this code block. IMO this will give more accurate/closer to the source information about the cause of the error and be better for troubleshooting.

We could consider catching and re-raising our own error to make it easier for callers to know what exceptions they might need to handle from our code, but attaching them to the original error so we don't lose the detail and stack trace. Something like:
```python
except hl7.exceptions.ParseException ex:
  raise OurOwnException("Detail") from ex
```

But we are not doing this anywhere else in our code, so I think it's best to allow the default behavior of bubbling up errors for now.

m-goggins: Thank you for the explanation. So if `hl7.parse` raises an exception, the `ParseException` error will still be raised and will indicate that it is coming from `hl7.parse`?

spencer-skylight: Yes, that's basically it!

PR #62, opened 2022-09-22
Census geocoding

m-goggins:
Adds a call to the Census API for either a single string address (`searchtype=onelineaddress`) or a dictionary (`searchtype=address`). The resulting GeoCodeResult includes similar components to Smarty with the addition of the following:
- census tract
- census block
- geoid (A combination of state, county, census tract, and census block info)

Precision is not returned for the Census API, and it appears to always be zip5. 

For the MVP, we assume the following input for the call to the API. In future iterations, these could be additional inputs from the user:

```
:param benchmark: A numerical ID or name that references what version of the
            locator should be searched. # Public_AR_Census2020
:param vintage: A numerical ID or name that references what vintage of geography
            is desired for the geoLookup (only needed when returntype = geographies) # Census2020_Census2020
:param format: The format to be used for returning the standardized output
            (json, html) # json
:param layers: By default, State, County, Tract, and Block layers are displayed
            when “geographies” is the chosen returntype. If additional or different
            layers are desired, they can be specified in a comma delimited list by ID
            or name # layer = 10, which is census tract/block info. You can change this to get ACS data (I think)
```


phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+
+class CensusGeocodeClient(BaseGeocodeClient):
+    """
+    Implementation of a geocoding client using the Census API.
+    """
+
+    def __init__(self):
+        self.__client = ()
+
+    # # @property
```
bryan-skylight, line 15: Is this supposed to be commented out? If so, can we just remove it from the code and add it back later when we're ready to implement it? I'd prefer to avoid commented code making it into production.

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+        return, otherwise the return None.
+
+        :param address: The address to geocode, given as a string
+        :param searchtype: onelineaddress OR address # doesn't yet support coordinates
+        :return: A GeocodeResult object (if valid result) or None (if no valid
+          result)
+        """
+        # Check for street num and name at minimum
+        if address == "":
+            raise Exception("Must include street number and name at a minimum")
```
bryan-skylight, line 41: Can we make this error message consistent with the error message used in `smarty.py`? I don't have a preference for which language we use, but I would prefer that they be consistent.

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+
+        formatted_address = self._format_address(address, searchtype="onelineaddress")
+        url = self._get_url(formatted_address)
+        response = self._call_census_api(url)
+
+        return self._parse_census_result(response)
+
+    def geocode_from_dict(self, address: dict) -> Union[GeocodeResult, None]:
```
bryan-skylight, line 49: I know that the smarty module does this as well, so I assume you're just implementing similar functionality, but do we have a use case in mind for when a user would pass a dictionary instead of a string? I'm wondering if we want to change this interface.

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def geocode_from_str(self, address: str) -> Union[GeocodeResult, None]:
+        """
+        Geocode a string-formatted address using Census API with searchtype =
+        "onelineaddress". If a result is found, encode as a GeocodeResult object and
+        return, otherwise the return None.
+
+        :param address: The address to geocode, given as a string
+        :param searchtype: onelineaddress OR address # doesn't yet support coordinates
+        :return: A GeocodeResult object (if valid result) or None (if no valid
```
bryan-skylight, line 36: Can we be more descriptive with the return values in the docstrings? Instead of focusing on the type that is returned, which is already included in the Sphinx docs by default, can we specify the context of what's being returned?

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+            f"https://geocoding.geo.census.gov/geocoder/geographies/{address}"
+            + "&benchmark=Public_AR_Census2020"
+            + "&vintage=Census2020_Census2020"
+            + "&layers=[10]"
+            + "&format=json"
+        )
+        return url
+
+    @staticmethod
+    def _call_census_api(url):
```
bryan-skylight, line 145: Given that we're handling the request ourselves in this module, as compared to the Smarty module where we utilize `send_lookup`, I'm wondering if we should incorporate our `http_request_with_retry` method to handle things like 429 HTTP status codes.

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+        :param response: The Census API client instantiated for geocoding
+        :return: A parsed GeocodeResult object (if valid result) or None (if
+          no valid result)
+        """
+        if lookup is not None and lookup.get("addressMatches"):
+            addressComponents = lookup.get("addressMatches")[0].get("addressComponents")
+            blockComponents = (
+                lookup.get("addressMatches")[0]
+                .get("geographies")
+                .get("Census Blocks")[0]
```
bryan-skylight, line 175: How confident are we that if "addressMatches" is present that:
1. it will always have at least one element present, instead of an empty list, and
2. that any given element in "addressMatches" will have the "geographies" and "Census Blocks" keys?

I'm worried that we could get unexpected `KeyError` exceptions given that `.get("geographies")` will return `None` by default. Would it make sense to modify the default values such that we have something like this?

```python
lookup.get("addressMatches", [{}])[0]
.get("geographies", {})
.get("Census Blocks", [None])[0]
```

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+            coordinateComponents = lookup.get("addressMatches")[0].get("coordinates")
+
+            # Format the Census result into our standard dataclass object
+            return GeocodeResult(
+                line=[
+                    item.strip()
+                    for item in lookup["addressMatches"][0]["matchedAddress"].split(",")
+                ],
+                city=addressComponents.get("city", ""),
```
bryan-skylight, line 193: I could be wrong about this, but I think the smarty module would return `None` for values that aren't present. Can we confirm that the return values for missing elements is consistent across both modules?

tests/geospatial/test_census_geocode_client.py
```diff
@@ -0,0 +1,139 @@
...
+    censusResponseFullAddress = json.load(
+        open(
+            pathlib.Path(__file__).parent.parent
+            / "assets"
+            / "censusResponseFullAddress.json"
+        )
+    )
+
+    # Test when addressMatches is an empty list
+    censusResponseFullAddress_noAddressMatch = censusResponseFullAddress
```
bryan-skylight, line 39: nit: Just for my own understanding, why are you loading the file under one variable name and then immediately setting it to another variables name?

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+        "address". If a result is found, encode as a GeocodeResult object and return,
+        otherwise the return None.
+
+        :param address: a dictionary with fields outlined above
+        :return: A GeocodeResult object (if valid result) or None (if no valid
+          result)
+        """
+
+        # Check for street num and name at minimum
+        if address.get("street", "") == "":
```
bryan-skylight, line 61: I don't think we necessarily need to change this in this PR, but I do have hesitations about not allowing geocoding if there's no street address. In some scenarios, deidentified/anonymized data may only include city, state, and zip as it relates to, say, HIV. Geocoding across several states to show a choropleth map of the most impacted states would require the ability to geocode at this level, no? Is this a restriction we're placing on the data, or one that the API is placing?

tests/geospatial/test_census_geocode_client.py
```diff
@@ -0,0 +1,139 @@
...
+    # Test empty string address
+    address = ""
+    geocode_result = None
+    try:
+        geocode_result = census_client.geocode_from_str(address)
+    except Exception as e:
+        assert (
+            repr(e) == "Exception('Must include street number and name at a minimum')"
+        )
+        assert geocode_result is None
```
bryan-skylight, line 84: nit: can we make the variable name consistent with the preceding and proceeding tests (`geocoded_response`)?

tests/geospatial/test_census_geocode_client.py
```diff
@@ -0,0 +1,139 @@
...
+    missing_street_dict = {
+        "city": "New York",
+        "state": "NY",
+        "zip": "10003",
+    }
+    try:
+        geocoded_response == census_client.geocode_from_dict(missing_street_dict)
+    except Exception as e:
+        assert (
+            repr(e) == "Exception('Must include street number and name at a minimum')"
```
bryan-skylight, line 125: Can we add the same assertion here that `geocoded_response` is still `None` after the attempt to geocode?

phdi/geospatial/census.py
```diff
@@ -0,0 +1,204 @@
...
+                            formatted_address += f"&state={state}"
+                        elif element == zip:
+                            formatted_address += f"&zip={zip}"
+                return formatted_address
+
+            else:
+                return f"onelineaddress?address={street}"
+
+        else:
+            raise Exception("Cannot geocode given address")
```
bryan-skylight, line 112: Can we make these exceptions more descriptive? Perhaps `ValueError`?

tests/geospatial/test_census_geocode_client.py
bryan-skylight: Also, I think we should rework this test (and the one below) a little bit. As it stands, I think this mostly tests that the exception message is correct, and not so much that it throws an exception. I think the following might better test for what we want to catch here.

```python
with pytest.raises(ValueError) as e:
    geocoded_response = census_client.geocode_from_str(address)

assert "Must include street number and name at a minimum" in str(e.value)
assert geocoded_response = None
```

See https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions for more info.

bryan-skylight (changes requested):
This looks really great so far! I have some changes that I'd like to see implemented, especially around the tests, but overall I think this is close to being done. Thanks for putting this together so quickly!

phdi/geospatial/census.py
m-goggins: I like "Address must include street number and name at a minimum." I'll change it here and in smarty.py

phdi/geospatial/census.py
m-goggins: Ha! this was my question as well. I think this is another question for user research. Brandon said the dict version was created as a prototype data storage object where people would likely store address information they had that wasn't in string form.

phdi/geospatial/census.py
m-goggins: Sounds like another one for UX :) The census API requires a street name and number at a minimum

tests/geospatial/test_census_geocode_client.py
m-goggins: Ah good catch! I think I had intended to add other tests but moved them elsewhere. I'll just changed the loaded file name.

phdi/geospatial/census.py
m-goggins: I added `http_request_with_retry` into the `_call_census_api` method but kept the same return logic in order to return the result if valid and pass that to the` _parse_census_result` method

phdi/geospatial/census.py
m-goggins: I am working on this - Spencer and Matt have been going back and forth in a separate space about formatting docstrings and other documentation items.

phdi/geospatial/smarty.py
```diff
@@ -52,7 +52,7 @@ def geocode_from_str(self, address: str) -> Union[GeocodeResult, None]:
         # The smarty Lookup class will parse a BadRequestError but retry
         # 5 times if the lookup address is blank, so catch that here
         if address == "":
-            raise Exception("Cannot geocode an empty string")
+            raise ValueError("Address must include street number and name at a minimum")
```
spencer-skylight, line 55: Thank you for updating this to throw a more specific exception. We should probably add a `raises` declaration to the docstring as well.

tests/geospatial/test_smarty_geocode_client.py
```diff
@@ -167,11 +167,17 @@ def test_blank_geocode_inputs():
     try:
```
spencer-skylight, line 167: I think a better way to do this is:
```python
pytest.raises(ValueError, match=r"Address must include street number and name at a minimum")
  smarty_client.geocode_from_str("")
```

Or it looks like you've done `pytest.rasies(ValueError) as e` and checked properties of e in your new tests which I think is also fine.

spencer-skylight (changes requested):
This is great! I had a bunch of little comments, but nothing major.

One other nit - percentage-wise coverage is good, but the lines we aren't hitting are exceptions and edge cases we should probably be testing.  Could you take care of those as well?

phdi/cloud/core.py
```diff
@@ -28,9 +28,7 @@ def get_access_token(self) -> str:
 
 class BaseCloudStorageConnection(ABC):
     @abstractmethod
-    def download_object(
-        self, container_name: str, filename: str, encoding: str = "utf-8"
-    ) -> str:
+    def download_object(self, filename: str, encoding: str = "utf-8") -> str:
```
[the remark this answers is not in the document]
m-goggins: huh I'm not totally sure where that change came from since I didn't think I touched the cloud files. In any case, I will add it back in.

phdi/geospatial/census.py
```diff
@@ -0,0 +1,230 @@
...
+        formatted_address = self._format_address(address, searchtype="address")
+        url = self._get_url(formatted_address)
+        response = self._call_census_api(url)
+
+        return self._parse_census_result(response)
+
+    @staticmethod
+    def _format_address(address: Union[str, dict], searchtype: str):
+        """
+        Format address for Census API call according to address type.
```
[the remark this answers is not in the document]
m-goggins: I added some explanation in the description but I think this should be part of a larger conversation around how we want to customize the census api moving forward. The Census API has several params that I did not include as user options (searchtype, layer, etc). for this iteration of the Census coder. I think in future iterations we should allow the user to have more flexibility in the vintage and layers they want to use. For the sake of the quick release and matching Smarty, I think we should keep these standardized (pre filled-in).

phdi/geospatial/census.py
```diff
@@ -0,0 +1,230 @@
...
+            locator should be searched. # Future consideration
+        :param vintage: A numerical ID or name that references what vintage of geography
+            is desired for the geoLookup (only needed when returntype = geographies)
+            # Future consideration
+        :param format: The format to be used for returning the standardized output
+            (json, html) # Future consideration
+        :param layers: By default, State, County, Tract, and Block layers are displayed
+            when “geographies” is the chosen returntype. If additional or different
+            layers are desired, they can be specified in a comma delimited list by ID
+            or name
```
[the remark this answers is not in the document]
m-goggins: Ditto the comment above around getting the initial version of the Census API wrapper out and then adding in customization based on user research.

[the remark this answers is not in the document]
bryan-skylight: mando-way-this-is-the-way

Although I would advocate for removing the parameters from the docstring until they're implemented, just to avoid any confusion.

phdi/geospatial/smarty.py
```diff
@@ -38,41 +38,43 @@ def client(self) -> us_street.Client:
...
     def geocode_from_dict(self, address: dict) -> Union[GeocodeResult, None]:
         """
-        Geocode a dictionary-formatted address using SmartyStreets.
-        If a result is found, encode as a GeocodeResult object and
+        Geocodes a dictionary-formatted address using SmartyStreets.
+        If a result is found, encodes as a GeocodeResult object and
         return, otherwise the return None.
 
-        :param address: a dictionary with fields outlined above
+        :param address: a dictionary with fields outlined above.
```
spencer-skylight, line 68: Nit: capitalize "a"

docs/geocoding_beta/overview.md
```diff
@@ -36,6 +36,16 @@ Read through the [Geospatial Tutorial](https://github.com/CDCgov/phdi/blob/main/
...
 
+**Select a geocoding service**
+- Smarty
+    - Partner will need to obtain a Smarty license (API key); see "Receiving a CDC subsidized Smarty license (API key)" section below for details
+    - Returned information includes standardized address, latitude, and longitude 
+- Census
+    - No API key needed
+    - Returned information includes standardized address, latitude, longitude, Census Tract, Census Block, and Census GEOID
+- Both Smarty and Census
+    - Partner will need to obtain a Smarty license (API key); see "Receiving a CDC subsidized Smarty license (API key)" section below for details
```
spencer-skylight, line 47: This last section should be removed.

phdi/geospatial/census.py
```diff
@@ -0,0 +1,230 @@
...
+        url = self._get_url(formatted_address)
+        response = self._call_census_api(url)
+
+        return self._parse_census_result(response)
+
+    @staticmethod
+    def _format_address(address: Union[str, dict], searchtype: str):
+        """
+        Format address for Census API call according to address type.
+        :param address: The address to geocode, given as a string
```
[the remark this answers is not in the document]
spencer-skylight: Nit: Suggesion: "The address to geocode, given as a string or dictionary."

tests/geospatial/test_smarty_geocode_client.py
spencer-skylight: This one wasn't implemented - was that intentional? It's not a sticking point for me - just a minor suggestion.

spencer-skylight (approved):
Super close from my perspective - just a few super nit-picky follow ups, but no blockers.

phdi/geospatial/census.py
bryan-skylight: That makes sense. In an effort to both keep the return statement short and also provide more useful information to the user, can I propose that we change this to something like "A standardized address enriched with lat, lon, census tract, and more. Returns None if there is no valid result."?

phdi/geospatial/census.py
bryan-skylight: If the Census API requires a street name and number then I'm good with leaving this as is. Again, not for this PR, but we should note that this is a possible limitation of this API for that use case.

phdi/geospatial/smarty.py
```diff
@@ -38,41 +38,43 @@ def client(self) -> us_street.Client:
...
+        Geocodes a string-formatted address using SmartyStreets. If the result
         comes back valid, output is stored in a GeocodeResult object. If the
         result could not be latitude- or longitude-located, then Smarty failed
         to precisely geocode the address, so no result is returned. Raises
         an error if the provided address is empty.
 
-        :param address: The address to geocode, given as a string
+        :param address: The address to geocode, given as a string.
+        :raises ValueError: If address does not include street number and name.
         :return: A GeocodeResult object (if valid result) or None (if no valid
```
bryan-skylight, line 49: Can I ask that we update this return description to match the suggestion I made above?

phdi/geospatial/smarty.py
```diff
@@ -38,41 +38,43 @@ def client(self) -> us_street.Client:
...
-        Geocode a dictionary-formatted address using SmartyStreets.
-        If a result is found, encode as a GeocodeResult object and
+        Geocodes a dictionary-formatted address using SmartyStreets.
+        If a result is found, encodes as a GeocodeResult object and
         return, otherwise the return None.
 
-        :param address: a dictionary with fields outlined above
+        :param address: a dictionary with fields outlined above.
+        :raises ValueError: If address does not include street number and name.
         :return: A GeocodeResult object (if valid result) or None (if no valid
```
bryan-skylight, line 70: Same request here about modifying the return description

tutorials/geospatial-tutorial.md
```diff
@@ -72,30 +75,51 @@ Listed below are several example use cases for employing the geospatial module.
 Suppose a data element has an address field in which the entire address occurs in a string, e.g.
 
 ```python
-location = "1234 Mulholland Dr. NW Cincinnati OH 43897"
+location = "5905 Wilshire Blvd Los Angeles CA 90036"
```
bryan-skylight, line 78: Just for my own awareness, was there a driving factor for changing this address? My hypothesis is that you're replacing a fake address that doesn't exist with a real address of a public institution so that readers can follow along with the tutorial. Is that right?

bryan-skylight (changes requested):
Line 163 of `census.py` is the big blocker for me at this moment. Otherwise, this looks really great!

tests/geospatial/test_smarty_geocode_client.py
m-goggins: Not intentional! I fixed it to match the others

phdi/geospatial/census.py
```diff
@@ -0,0 +1,220 @@
...
+
+        response = http_request_with_retry(
+            http_url,
+            http_retry_count,
+            http_action,
+            [http_action],
+            http_header,
+        )
+
+        if response.status_code < 400:
```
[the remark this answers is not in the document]
m-goggins: I made the changes that you suggested and added in a test to make sure this line got coverage per @spencer-skylight 's suggestion, but I don't think it's a great test. Could one of you take a look at it?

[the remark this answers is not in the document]
spencer-skylight: Thanks for tagging me on this! I think it's testing a valid scenario, but we should be mocking the response from the Census API rather than executing actual requests against it. 

Also, the `else:` clause at the end of `_call_census_api` is probably unnecessary.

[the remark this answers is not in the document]
m-goggins: Hm I think we do need the else clause to return the resulting geocoded address.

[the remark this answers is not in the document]
m-goggins: Per our discussion earlier, I implemented the mocked up the response from the census API

tests/geospatial/test_census_geocode_client.py
```diff
@@ -0,0 +1,166 @@
...
+        + "onelineaddress?address=239+Greene+St+New+York%2C+NY"
+        + "&benchmark=Public_AR_Census2020"
+        + "&vintage=Census2020_Census2020"
+        + "&layers=[10]"
+        + "&format=json"
+    )
+
+    with pytest.raises(requests.exceptions.HTTPError) as e:
+        CensusGeocodeClient._call_census_api(url)
+        assert str(e) == "<ExceptionInfo HTTPError() tblen=2>"
```
bryan-skylight, line 29: nit: This is nit-pickiest thing in the world, but I'm a big fan of consistency. Can you make this test match the others in that a) the `assert` statement is outside of the with, and b) either modify this comparison to match the ones below (i.e. `"Exception Message" in str(e)`) or modify the ones below to match this approach (I think I like this approach more, but dealer's choice)

bryan-skylight (approved):
pitch-perfect-fat

PR #68, opened 2022-09-27
Create config.yml

bryan-skylight:
Prevents the ability to create a blank issue


bryan-skylight:
Was attempting to implement functionality to order issue templates, but it wasn't what I expected. Abandoning this effort.

PR #70, opened 2022-09-28
Convert issues to web forms

bryan-skylight:
# Description
For some reason, Nagela isn't being added as an assignee on the Geo Beta Program issues, and I think it's because her username needs to be lowercase. Trying that here.


bryan-skylight:
This actually isn't necessary. the problem was that Nagela was not a member of the repo at the time I tested. This has been resolved.

PR #71, opened 2022-09-30
FHIR Census geocoding

m-goggins:
Creates a version of the census geocoder for handling geocoding of FHIR resources and bundles. Importantly, a new method `_store_census_tract_extension ` has been added. It adds a census tract Address.line extension to a bundle for every element of an Address line. Note: while the base census geocoder is able to return census block and census geoid, the FHIR version does not yet support adding these items to a resource.

I also made a small change to the main census geocoder client, but I think we should have a discussion around it (perhaps outside github). Originally, the census geocoder returned the full matched address as the address line, e.g., ["239 GREENE ST", "NEW YORK", "NY", "10003"]. This does not match the smarty return (["239 GREENE ST"]) and became a problem when attempting to add a census tract extension for every element of the line; in the above case, the census tract extension would have been added 4 times. I modified so that only the 1st element of the address line is returned for the census geocoder, e.g., ["239 GREENE ST"]. 

A somewhat related problem is that the Census coder removes any 2nd address line information in the matched address, e.g. "239 Greene St, Apt 4L" becomes "239 Greene St." This is not a problem for census coding, but could be a problem in the case of overwriting an address in a patient resource (I think), because address information could potentially be lost.


m-goggins:
Does the code coverage drop dramatically for anyone else? It doesn't look like the coverage report includes fhir/geospatial/census.py for some reason. Wondering if this is just on my end.

tests/fhir/geospatial/test_fhir_census_client.py
```diff
@@ -0,0 +1,130 @@
+import json
+import pathlib
+import copy
+
+from phdi.fhir.geospatial.census import CensusFhirGeocodeClient
+from phdi.geospatial.core import GeocodeResult
+
+
+def test_geocode_resource_census():
```
spencer-skylight, line 9: Test test cases should patch the census geocoder.

spencer-skylight, line 9: Could you test a few edge cases:
* Patient already has an extension on `line`, and it's preserved.
* Existing extension on `line` has one row explicitly set to `None`.

I think you could cover both test cases by doing the following...
```python
patient['_line'] = []
patient['_line'].append("None")
patient['_line'].append({
"extension": [ ...an existing extension... ]
})

File: tests/fhir/geospatial/test_fhir_census_client.py [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+    # Case 1: Overwrite = False
+    returned_patient = census_client.geocode_resource(patient, overwrite=False)
+    assert standardized_patient == returned_patient
+    assert returned_patient != patient
+
+    # Case 2: Overwrite = True
+    assert standardized_patient == census_client.geocode_resource(
+        patient, overwrite=True
+    )
+    assert patient == census_client.geocode_resource(patient, overwrite=True)
```
spencer-skylight, line 72: I think you should be checking against `standardized_patient` rather than geocoding again.

phdi/geospatial/census.py
```diff
@@ -203,8 +203,7 @@ def _parse_census_result(lookup) -> Union[GeocodeResult, None]:
             # Format the Census result into our standard dataclass object
             return GeocodeResult(
                 line=[
-                    item.strip()
-                    for item in lookup["addressMatches"][0]["matchedAddress"].split(",")
+                    lookup["addressMatches"][0]["matchedAddress"].split(",")[0].strip()
```
spencer-skylight, line 206: Why did you have to remove address lines after the 1st one from the GeocodeResult?

spencer-skylight, line 206: What happens if no responses come back from the Census API? Will you always get a value for `addressMatches` and `addressMatches.matchedAddress`?

phdi/fhir/geospatial/census.py
```diff
@@ -0,0 +1,82 @@
...
+        :param overwrite: Whether to overwrite the address data in the given
+          bundle's resources (True), or whether to create a copy of the bundle
+          and overwrite that instead (False). Defaults to True.
+        :return: A FHIR bundle with geocoded address(es).
+        """
+        if not overwrite:
+            bundle = copy.deepcopy(bundle)
+
+        for entry in bundle.get("entry", []):
+            _ = self.geocode_resource(entry.get("resource", {}), overwrite=True)
```
spencer-skylight, line 80: I don't think you need to assign the `_` variable here - just call the function.

phdi/fhir/geospatial/census.py
```diff
@@ -0,0 +1,82 @@
...
+
+        return resource
+
+    def _geocode_patient_resource(self, patient: dict) -> None:
+        """
+        Handles geocoding of all addresses in a given patient resource.
+        :param patient: The patient resource whose addresses should be geocoded.
+        """
+        for address in patient.get("address", []):
+            onelineaddress = get_one_line_address(address)
```
spencer-skylight, line 46: Since FHIR has the address split into discrete components, and the Census geocoder supports geocoding discrete components, I'm thinking we should call geocode_from_dict instead of geocode_from_str...

spencer-skylight (changes requested):
Nice work! :guitar: I had a few suggestions to consider.

phdi/geospatial/census.py
m-goggins: This is what I was trying to outline in the comment I left on the PR (last paragraph):

I also made a small change to the main census geocoder client, but I think we should have a discussion around it (perhaps outside github). Originally, the census geocoder returned the full matched address as the address line, e.g., ["239 GREENE ST", "NEW YORK", "NY", "10003"]. This does not match the smarty return (["239 GREENE ST"]) and became a problem when attempting to add a census tract extension for every element of the line; in the above case, the census tract extension would have been added 4 times. I modified so that only the 1st element of the address line is returned for the census geocoder, e.g., ["239 GREENE ST"].

A somewhat related problem is that the Census coder removes any 2nd address line information in the matched address, e.g. "239 Greene St, Apt 4L" becomes "239 Greene St." This is not a problem for census coding, but could be a problem in the case of overwriting an address in a patient resource (I think), because address information could potentially be lost.

m-goggins: As long as the URL for the API call is valid, you will get a response that contains `addressMatches `and `matchedAddress`. If there is no corresponding match, the returned values will be empty and like the Smarty geocoder, return None. I believe `test_parse_census_result_failure()` tests for this scenario in test_census_geocode_client.py

phdi/fhir/geospatial/census.py
m-goggins: I was thinking that too originally. However, the geocode_from_dict() is expecting "street" (string) as a key instead of "line" (list of street values). I don't think I'm opposed to switching to geocode_from_dict() but it will impact the expected key-value pairs in geocode_from_dict() and make the function different from the smarty version. Thoughts?

spencer-skylight: Just documenting what we discussed... I think we can concatenate the multiline streets with a space. Doing that small amount of concatenation should be more easily interpretable by the Census API vs concatenating the whole address into a single line.

Our success with this does depend somewhat on the Census API's implementation, but it seems like a good place to start.

phdi/geospatial/census.py
spencer-skylight: Documenting what we discussed - we will remove "line" from the replaced fields because the Census geocoder doesn't always preserve things like unit number in the response info. 

This doesn't necessarily change the behavior of the (non-FHIR) geocoder.  The non-FHIR geocoder can still return line number in the `GeocodeResult` for the user to do with as they please.

phdi/fhir/geospatial/core.py
```diff
@@ -72,3 +72,43 @@ def _store_lat_long_extension(address: dict, lat: float, long: float) -> None:
...
+        census_extension = {
+            "url": "http://hl7.org/fhir/StructureDefinition/iso21090-ADXP-censusTract",
+            "valueString": census_tract,
+        }
+
+        if address.get("_line") is None:
+            address["_line"] = []
+        for element_counter in range(len(address["line"])):
+            try:
+                if address["_line"][element_counter].get("extension", "") is not None:
```
spencer-skylight, line 101: I think this if statement is unnecessary.  The line below should raise the exceptions you are using to handle edge cases.

phdi/fhir/geospatial/core.py
```diff
@@ -72,3 +72,43 @@ def _store_lat_long_extension(address: dict, lat: float, long: float) -> None:
...
+            except AttributeError:
+                address["_line"][element_counter] = {"extension": []}
+                address["_line"][element_counter].get("extension").append(
+                    census_extension
+                )
+            except IndexError:
+                address["_line"].append({"extension": []})
+                address["_line"][element_counter].get("extension").append(
+                    census_extension
+                )
```
spencer-skylight, line 114: Since you are initializing the extension array here, you know it's empty. Changing it like this would be clearer/more efficient.
```suggestion
                address["_line"].append({"extension": [ census_extension ]})
```

tests/fhir/geospatial/test_fhir_census_client.py
```diff
@@ -0,0 +1,221 @@
...
+                "url": "http://hl7.org/fhir/StructureDefinition/"
+                + "iso21090-ADXP-censusTract",
+                "valueString": "59",
+            }
+        ]
+    }
+
+    # Case 3: Patient already has 1 extension on _line, and it's preserved.
+    returned_patient = census_client.geocode_resource(patient, overwrite=True)
+    # print()
```
spencer-skylight, line 135: Let's get rid of these debug statements :slightly_smiling_face:

tests/fhir/geospatial/test_fhir_census_client.py
spencer-skylight: I think this still isn't covering the case where the patient address coming in to the geocoder has an existing extension on `line`.

To correct (and simplify) the code I suggested above, you could try this:
`patient['address'][0]['_line'] = [ None, { "extension": [ ... an existing extension ... ] } ]`

This should add an extension to the _input_ of the function you're testing, and you can make sure it doesn't cause errors or get overwritten.

spencer-skylight (changes requested):
A few minor suggestions.

tests/fhir/geospatial/test_fhir_census_client.py
m-goggins: I probably could have been clearer but I added an extension in a new asset that looks like the following. I think it accomplishes the same thing as your code, but correct me if I'm wrong.

`"address": [
                    {
                        "line": [
                            "239 Greene St",
                            "Apt 4L"
                        ],
                        "city": "New York",
                        "state": "NY",
                        "postalCode": "10003",
                        "country": "USA",
                        "use": "home",
                        "_line": [
                            {
                                "extension": [
                                    {
                                        "existing_url": "existing_extension"
                                    }
                                ]
                            },
                            null
                        ]
                    }
                ],`

phdi/fhir/geospatial/core.py
m-goggins: Ah good catch! I added the AttrubuteError but forgot to remove the if statement.

tests/fhir/geospatial/test_fhir_census_client.py
spencer-skylight: Yep, you're right - looks good!

spencer-skylight (approved):
LGTM!

PR #73, opened 2022-10-04
Migrate PHDI to use containerized conversion solution

spencer-skylight:
# Description
The conversion code has been updated to consume the containerized solution created by the cloud implementation team. Related documentation has also been added.

# Testing
In addition to standard unit tests, tests were additionally run against the containerized solution itself.  After installing Docker, the solution can be run from a shell using:

```bash
docker run -p 8080:8080 docker.io/gskathol/fhir-converter
```

An HL7 test can be performed using:
```python
import phdi.fhir.conversion

hl7_message = open("tests/assets/FileSingleMessageSimple.hl7").read()
phdi.fhir.conversion.convert_to_fhir(hl7_message,"http://0.0.0.0:8080/convert-to-fhir").json()
```

```python
import phdi.fhir.conversion

ccda_message = open("tests/assets/ccda_sample.xml").read()
phdi.fhir.conversion.convert_to_fhir(ccda_message,"http://0.0.0.0:8080/convert-to-fhir").json()
```

# Additional notes
A few TODOs were added to update documentation pointing to the containerized solution after it's been moved to the PHDI library.

### DIFF
--- a/phdi/fhir/conversion/convert.py
+++ b/phdi/fhir/conversion/convert.py
@@ -71,7 +71,7 @@ def convert_to_fhir(
     }
 
     if cred_manager:
-        access_token = cred_manager.get_access_token().token
+        access_token = cred_manager.get_access_token()
         headers["Authorization"] = f"Bearer {access_token}"
         response = http_request_with_reauth(
             cred_manager=cred_manager,
--- a/phdi/fhir/conversion/convert.py
+++ b/phdi/fhir/conversion/convert.py
@@ -5,6 +5,7 @@ import xml.etree.ElementTree as et
 from phdi.harmonization import standardize_hl7_datetimes
 from phdi.cloud.core import BaseCredentialManager
 from phdi.fhir.transport import http_request_with_reauth
+from phdi.transport.http import http_request_with_retry
 
 
 CCDA_CODES_TO_CONVERSION_RESOURCE = {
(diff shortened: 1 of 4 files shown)


phdi/fhir/conversion/convert.py
```diff
@@ -22,73 +23,77 @@
 
 def convert_to_fhir(
     message: str,
-    cred_manager: BaseCredentialManager,
-    fhir_url: str,
+    url: str,
+    cred_manager: BaseCredentialManager = None,
+    headers: dict = {},
```
DanPaseltiner, line 28: I like including `header` here.

phdi/fhir/conversion/convert.py
```diff
@@ -22,73 +23,77 @@
...
+            headers=headers,
+            data=data,
+        )
 
     if response.status_code != 200:
         raise Exception(
-            f"HTTP {str(response.status_code)} code encountered in $convert-data for a message"  # noqa
+            f"HTTP {str(response.status_code)} code encountered during conversion request"  # noqa
         )
     return response
```
DanPaseltiner, line 99: Do we want to raise an exception here, or just return the response so that the user can see what went wrong during conversion? The converter provides relatively robust messaging back to users when they make bad requests.

DanPaseltiner (changes requested):
Thanks for putting this together. This looks mostly good to me, but I think we should allow the user of this function to see the full response form the FHIR converter even when the status code is not 200. Our request validation with Pydantic gives useful feedback when users make bad requests. Additionally, when conversion fails we currently return whatever error messaging the MSFT converter CLI generates.

phdi/fhir/conversion/convert.py
spencer-skylight: I think it makes sense to return response regardless of status code. As an alternative, we could raise a requests.HTTPError that contains the response, so the caller can get to it if they need to.

IIRC @bryan-skylight might have been advocating for going the exception route, and @bamader might also have an opinion. What are your thoughts here?

phdi/fhir/conversion/convert.py
```diff
@@ -22,73 +23,77 @@
...
-            },
-            {
-                "name": "templateCollectionReference",
-                "valueString": conversion_settings.get("template_collection"),
-            },
-            {
-                "name": "rootTemplate",
-                "valueString": conversion_settings.get("root_template"),
-            },
-        ],
```
bamader, line 74: Is the fact that we can simplify this due to the converter that's been containerized? I love that we can eliminate these lines, just want to make sure a user's request will always be packaged appropriately to the container.

phdi/transport/http.py
```diff
@@ -48,6 +48,7 @@ def http_request_with_retry(
     )
     adapter = HTTPAdapter(max_retries=retry_strategy)
     http = requests.Session()
+    http.mount("http://", adapter)
```
bamader, line 51: What's the reason for this addition? Shouldn't we always be opting for secured channels when possible?

bamader (approved):
Good stuff here! I think once we get that response (or error holding a response) configured we'll be good to go!

phdi/fhir/conversion/convert.py
spencer-skylight: I thought about this more and decided that wrapping in an HTTPError wasn't exactly correct, since the HTTP request actually succeeded, it is just reporting an error that occurred during the conversion operation. So I created a ConversionError, with an http_response property. Curious to know what you think of this approach!

phdi/fhir/conversion/convert.py
```diff
@@ -22,73 +23,77 @@
...
       map to a specific supported template (Optional, default is No)
     :return: A requests.Response object
 
     """
+    # TODO Update documentation with a link to the containerized FHIR converter, once
+    # it's been ported over to the phdi repository.
+
     conversion_settings = _get_fhir_conversion_settings(message, use_default_ccda)
-    if conversion_settings.get("input_data_type") == "HL7v2":
+    if conversion_settings.get("input_type") == "hl7v2":
```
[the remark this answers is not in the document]
spencer-skylight: Lowercase is currently required by the containerized conversion service.

DanPaseltiner (approved):
Thanks for the pairing session. This looks like a good solution to me.

phdi/transport/http.py
spencer-skylight: The containerized conversion service only operates in http mode currently. The plan is to have cloud services handle https encryption, but since we also want to be able to run it against the container locally, we added the ability to connect via http. For on-prem solutions, this should be OK as long as the converter service is protected by the firewall.

PR #75, opened 2022-10-05
Release documentation

spencer-skylight:
# Description

Create documentation to define the initial release process at PHDI.


RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+  Minor version releases _may_ also include fixes.
+
+* **PATCH** versions introduce non-breaking bug fixes.
+
+  Releases that _only_ contain fixes are released as patches.
+
+
+## PHDI Release Process
+There are several steps to the release process.
```
bryan-skylight, line 31: I might recommend removing this line. Or perhaps add more to it? It seems distracting being here as a single sentence for some reason.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+  Releases that _only_ contain fixes are released as patches.
+
+
+## PHDI Release Process
+There are several steps to the release process.  
+
+
+### Define a Target Version
+The very first step in a release is defining a version number to release to. The previous section describes the release versioning scheme, and should be used to assign the version number for the next release. Your release will have the following parts:
```
bryan-skylight, line 35: Recommended rewording: The very first step in a release is defining the release's new version number.

Recommended reformatting: Move the parts about "(Major|Minor) versions will reset ..." up to the section about the version parts.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+* `MAJOR`: The major version number for the release. Major versions will reset `MINOR` and `PATCH` versions to 0.
+* `MINOR`: The minor version number. Minor versions will reset `PATCH` version to 0.
+* `PATCH`: The patch version number.
+
+The values identified will be referenced in the sections below.
```
bryan-skylight, line 40: Actually, now that I'm looking at this, I might delete this whole section as it seems to mostly be repeating what was stated in the first section (after we move the stuff about resetting version numbers). Maybe just combine the two sections with something like this:

# PHDI Release Documentation

## Release Methodology: Semantic Versioning
This document describes how the PHDI library, tools and related documentation are released to users. API documentation is published automatically with Sphinx and hosted via GitHub pages. PHDI updates are released to the Python Package Index (PyPI) according to the guidelines set out in [Semantic Versioning 2.0.0](https://semver.org/) with each release's version following the pattern of MAJOR.MINOR.PATCH. The following core tenets describe when each element of a release's version would be updated.

* **MAJOR** versions introduce breaking changes.

  A breaking change breaks backwards-compatibility with previous released versions. In other words, a breaking change is something that may cause a client's implementation to stop working when upgrading from a previous version. Common examples of breaking changes include:
  
    * Deleting a package or public functions/methods
    * Deleting public function parameters
    * Changing a function name
    * Changing the name or order of required parameters
    * Adding new required parameters

Major version releases _may_ also include non-breaking enhancements and fixes. When major versions are released, the minor and patch version numbers reset to 0.

* **MINOR** versions introduce new non-breaking functionality.
  Releases with enhancements that do not break backwards compatibility require a minor version update. Common examples of non-breaking changes include:
  
    * Adding a package, module, or method
    * Adding optional parameters

  Minor version releases _may_ also include fixes. When minor versions are released, the patch version number resets to 0.

* **PATCH** versions introduce non-breaking bug fixes.

  Releases that _only_ contain fixes are released as patches.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Major Version Release Process
+In order to support patching old major version releases without forcing users to upgrade, new major versions involve creating a release branch. The following steps should be followed when a breaking change is merged into `main`, requiring a new major version.
+
+First, in GitHub, create a new branch representing the **old/existing major version**. The new branch should be based on the commit prior to the breaking change that triggers a new version. The naming convention for the new branch is:
+
+```
```
bryan-skylight, line 47: Recommend making this inline code instead of a code block.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+vMAJOR
+```
+
+Where `MAJOR` is the current major version, prior to introducing the breaking change. The following commands may be used to create the new branch:
```
bryan-skylight, line 51: This would then get added to the previous sentence so it would look like this:

The naming convention for the new branch is `vMAJOR` where `MAJOR` is the current major version, prior to introducing the breaking change.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+```bash
```
bryan-skylight, line 53: I'm not sure we should add the commands for creating the new branch since that can cause issues if we don't explain the need to be on `main`, etc. I think referring to the commit hash helps alleviate that, but I don't think the instructions are clear enough as currently written to ensure it's done right.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+git branch vMAJOR COMMIT-HASH
+```
+
+* `MAJOR` is the major version number of the **old/existing major version**
+* `COMMIT-HASH` is the commit hash of the commit prior to the breaking merge commit (if the breaking commit has not been merged to `main` and can just specify `main`)
+
+
+### PyPI Release
+#### Authenticating with PyPI
+PHDI recommends that users authenticate with PyPI using an API token. To download a token, log in to PyPI, and access [project settings](https://pypi.org/manage/project/phdi/settings/). Select "Create a token for phdi" and create a phdi-specific token. Save a phdi-specific token to a file named `.pypitoken` in your project directory. When publishing to PyPI as described later in this section, this API token will be used to authenticate.
```
bryan-skylight, line 63: typo: "log into PyPI, and" --> "log into PyPI and"

Recommended rewording: "Save a phdi-specific token" --> "Save the token you just created"

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Prepare to Publish
+The steps below depend on defining a full version number using semantic versioning. See the release methodology section above for help assigning a version to your project.Release Methodology: Semantic Versioning
```
bryan-skylight, line 66: I'd recommend removing "Release Methodology: Semantic Versioning". The doc isn't long enough to need to reference which section a user should review.

bryan-skylight (changes requested):
Thanks for putting this all together! Reading through it I'm sure I'd be able to publish a new version of PHDI on my own, which is great! The next time we need to release a new version, I'd like to test this by trying to publish to PyPI using only the instructions in this doc. 

I did have some changes I'd like to see made though. Mostly just reformatting and fixing some typos.

RELEASE.md
spencer-skylight: It makes sense to me to consolidate details about major/minor/patch versions to the semantic versioning section. I hesitate to remove the section completely since I think assigning the values for the version parts is part of the process and should be mentioned in the process section. But I scaled back the text to refer to the previous section for details on how the values should be assigned. Hopefully this makes sense as good middle ground.

spencer-skylight:
Comments have been addressed, and a few additional edits made:
* Updated section headers to that describe actions to perform to be consistent, in present tense.
* Added a section on creating a new GitHub release.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+* `MAJOR` is the major version number of the **old/existing major version**
+* `COMMIT-HASH` is the commit hash of the commit prior to the breaking merge commit (if the breaking commit has not been merged to `main` and can just specify `main`)
+
+#### Create a GitHub Release
+A GitHub release ties together a few pieces of information in the release process:
+1) A tag for the commit that comprises the release
+2) A short description/name for the release
+3) A list of changes in the release
+
+In order to create a new release in GitHub, open a [new release GitHub form](https://github.com/CDCgov/phdi/releases/new). First, select a target branch - this will either be `main` or a the previous version branch if the release is for an older version. Create (or select) a tag named `vMAJOR.MINOR.PATCH`, replacing the release parts as appropriate with the correct release version values.
```
bryan-skylight, line 55: typo: "this will either be `main` or _a the_ previous version" --> "this will either be `main` or the previous version"

bryan-skylight (approved):
NICE! There's a single typo that I found, but once that's corrected this is good to go.

PR #77, opened 2022-10-05
Bugfix cu#30zf9qg

bryan-skylight:
# Description
A bug was identified in which we required the "use" of a Resource field be specified, but the STU3 and R4 specifications for FHIR do not list "use" as a required data point. This PR addresses this issue by modifying `get_field` to make the `use` parameter optional. In doing so, "use" now becomes an optional filtering criteria, as well as a means for indicating priority. If a user wants only "official" names, then they can set the `use` parameter to `"name"` and the `require_use` parameter to `True`.  However, if they'd _prefer_ to get the home phone of a patient, but will accept any number in the record, then they can set `use` to `"mobile"` and `require_use` to `False`. Lastly, if they have no specific use they care about, `use` can be set to `None` and `require_use` will be ignored.

The function signature has also been updated to reflect that the return value can be one of several data types. This violates one of our design principles, which states that functions should return one (two if `None` can be returned) data type, but we view this as a necessity for the sake of simplifying things for our end users. 

# What this doesn't do
This does not account for more complex data types that FHIR supports, nor does it support field extraction deeper than the the top level. For example, a user cannot extract _only_ the `Address.line` field, and must instead extract the whole address. Supporting this nested extraction would require implementing FHIRPath, which is a bridge we will cross later.

# Feedback requested
The key question I had was around how to handle the various return types, which was settled via a Slack conversation. Other than that, the common key questions are still outstanding:

1. Are we missing any edge cases?
2. Are the tests covering what they need to cover?
3. Does the logic make sense?
4. Is this usable by our users?

# Tests
`test_utils.py` needed to be updated to reflect the new function signature, but all tests pass and the `test_get_field` test already tests for an option `use` parameter in its attempt to extract "telecom".


phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
     :return: The first instance of the field value matching the desired
       use, or a default field value if a match couldn't be found
     """
     if field == "":
         raise ValueError("Field must be a defined, non-empty string")
-    if use == "":
-        raise ValueError("Use must be a defined, non-empty string")
     if field not in resource:
-        raise KeyError(f"Given resource does not contain a key called {field}")
+        raise KeyError(f"This resource does not contain a field called {field}")
```
BradySkylight, line 55: This may be nit-picky, especially since there are different meaning for these terms (field, element, etc... ) depending upon the context (FHIR vs. coding, etc..).  It may be good to follow the FHIR definition of terms here?  ie: "This resource does not contain an element called {field}".  I could go either way and depends upon what would be better for the user/audience.

phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
-    :param use: The 'use' the field must have to qualify
-    :param default_field: (optional) The index of the field type to treat as
-      the default return type if no field with the requested use case is
-      found; if not supplied, use the first data available
+    :param index: (optional) The nth element of the field to return. Note that the index
+    starts at 0, not 1. If the index is greater than the number of elements in the
+    field, the last element will be returned. Defaults to returning the first element.
+    :param use: (optional) The 'use' the field must have to qualify for selection
+    :param require_use: (optional) Indicates if the 'use' is required. If True, then
+    if no elements of the specified field have that use, none will be returned. If
```
m-goggins, line 46: Nit but slightly more understandable to me: If True _and_ if no elements of the specified field have that use, none will be returned

phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
+    False, then if no elements of the specified field have that use the default index
```
m-goggins, line 47: If False _and_

BradySkylight (approved):
This all looks good to me.  I didn't have one comment/recommendation, but isn't 'required' as it's a bit nitpicky.  

To answer the questions:

1. Are we missing any edge cases?
I think so other than the one concern discussed and resolved during stand-up.
2. Are the tests covering what they need to cover?
Yes, they are clear and straightforward.
3. Does the logic make sense?
Yes, and I like your Python wizardry!
4. Is this usable by our users?
Yes, I think this is a simple enough change and the function/method is simple to understand and use.

phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
+    will be returned. This parameter is ignored if no use is specified.
     :return: The first instance of the field value matching the desired
       use, or a default field value if a match couldn't be found
     """
     if field == "":
         raise ValueError("Field must be a defined, non-empty string")
-    if use == "":
-        raise ValueError("Use must be a defined, non-empty string")
     if field not in resource:
```
m-goggins, line 54: I'm wondering if it would be helpful to the user to also include the use in the error message (if use is included). For example, someone inputs a legitimate field and use, but the particular field they specify is not in the specified use and is instead in another section of the resource?

spencer-skylight (changes requested):
A few comments, functionally it looks good.

I'd suggest including a few other output datatype in the code coverage since it was a focus of the changes made.  There's also a test that I'd expect to fail, asserting an exception on a missing use parameter. That should be updated to assert it returns the first entry I think.

I'm not sure why the checks aren't running for this PR - maybe just a GitHub hiccup? I'm happy to pair or dig in further on that if needed.

phdi/fhir/utils.py
```diff
@@ -71,7 +77,7 @@ def get_one_line_address(address: dict) -> str:
     if len(address) == 0:
         return ""
     raw_one_line = " ".join(address.get("line", []))
-    raw_one_line += f" {address.get('city', '')}, {address.get('state', '')}"
-    if address.get("postalCode", "") != "":
+    raw_one_line += f" {address.get('city')}, {address.get('state')}"
```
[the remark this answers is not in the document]
bryan-skylight: You're absolutely right, and I'm not sure how this change got made because I didn't make tweaks to this line of code. 😅 I'll change it back to what it was before, because `None` should definitely not be the default.

phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
 
 
-def get_field(resource: dict, field: str, use: str, default_field: int = 0) -> str:
+def get_field(
+    resource: dict,
+    field: str,
+    index: int = 0,
+    use: str = None,
+    require_use: bool = True,
+) -> Union[Dict, List, str, None]:
```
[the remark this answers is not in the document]
bryan-skylight: Should we just change this to `Any` then?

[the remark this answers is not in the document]
spencer-skylight: Sounds fine to me! A little less specific, but also less room for error, and I think it makes sense based on the number of possible return types.

spencer-skylight (commented):
One more quick comment - as long as you're modifying test cases, you might consider changing the ones that should throw exceptions to use `pytest.raises`. Right now, if the function calls _don't_ raise the expected exception, the test could still succeed when we'd want it to fail.

phdi/fhir/utils.py
bryan-skylight: Given that this error is related to the field we're extracting (e.g. a user wants to extract "name" from a Lab resource and it's not there) and not the element (e.g. The Patient's given name is "Sam") we should refer to this as the field and not the element, right? Or do I have those backwards?

phdi/fhir/utils.py
bryan-skylight: Fixed in the latest commit.

phdi/fhir/utils.py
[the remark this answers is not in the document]
bryan-skylight: Done!

phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
-    the requested use case can be found, instead return a specified
-    default value for the field.
+    formatted JSON dict. Optionally, a particular "use" case of a
+    given field (such as name or address), can be provided such that
+    only instances with that use case are considered. Use case here
+    refers to the FHIR-based usage of classifying how a value is used
+    in reporting. For example, find the first name for a patient that
+    has a "use" of "official" (meaning the name is used for official
+    reports). If no instance of a field with the requested use case
+    can be found, instead return a specified default value for the field.
```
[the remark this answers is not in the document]
bryan-skylight: Great call out! I've updated the docstring and would appreciate your feedback on the new language. Going to resolve this comment in favor of creating a new one should additional changes need to be made.

phdi/fhir/utils.py
```diff
@@ -18,46 +18,52 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
     :param field: The field to extract
-    :param use: The 'use' the field must have to qualify
-    :param default_field: (optional) The index of the field type to treat as
-      the default return type if no field with the requested use case is
-      found; if not supplied, use the first data available
+    :param index: (optional) The nth element of the field to return. Note that the index
+    starts at 0, not 1. If the index is greater than the number of elements in the
+    field, the last element will be returned. Defaults to returning the first element.
+    :param use: (optional) The 'use' the field must have to qualify for selection
+    :param require_use: (optional) Indicates if the 'use' is required. If True, then
```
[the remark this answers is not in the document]
bryan-skylight: Done!

phdi/fhir/utils.py
m-goggins: In retrospect, I think I was confusing use and field. I was thinking of a scenario where someone includes a non-empty field, e.g. "name" and a non-empty resource "case", but name is not in the case resource. I see that is taken care of in the line just below this comment 😀

phdi/fhir/utils.py
```diff
@@ -18,60 +18,71 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
-    :param use: The 'use' the field must have to qualify
-    :param default_field: (optional) The index of the field type to treat as
-      the default return type if no field with the requested use case is
-      found; if not supplied, use the first data available
+    :param resource: A FHIR-formatted resource.
+    :param field: The field to extract.
+    :param index: The nth element of the field to return. If the index is greater than
+    the number of elements in the field, the last element will be returned. If the index
+    is less than 1, the first element will be returned. Defaults to returning the first
+    element.
```
bamader, line 41: Super small nit: spacing on the param docstring lines after the first should be prefaced with two spaces

phdi/fhir/utils.py
```diff
@@ -18,60 +18,71 @@ def find_entries_by_resource_type(bundle: dict, resource_type: str) -> List[dict
...
+    :param use: The 'use' the field must have to qualify for selection.
+    :param require_use: If True and no elements of the specified field have that
+    use, none will be returned. If False and no elements of the specified field
+    have that use, the nth element as indicated by the index parameter will be returned.
+    This parameter is ignored if no use is specified.
```
bamader, line 46: Spacing thing here too

bamader, line 46: I believe according to the style guide, the `index`, `use`, and `require_use` docstring lines should have second sentences of the form `Default: True.` specifying their default values.

bamader (approved):

phdi/fhir/utils.py
bryan-skylight: Good call out! Most recent commit address this and the comment you left below.

phdi/fhir/utils.py
bryan-skylight: This has been added in the most recent commit. Thanks for flagging!

PR #80, opened 2022-10-11
Move FHIR converter container code from GCP repo to phdi repo

nickclyde:
Resolves https://app.clickup.com/t/3nn7jw1

This adds a `containers/` directory at the root of the repo. Each subfolder of that directory will contain code for building a Docker container. To start off, I've copied the code over from the GCP repo for the FHIR converter container.

I've also added a step to the `test` workflow to test the container code, as well as a new workflow, `buildContainers`, which will build the container images using Docker, and push them to this repo's container registry, so that the images will display on the repo homepage, and can be downloaded by users from `ghcr.io`.


.github/workflows/buildContainers.yaml
```diff
@@ -0,0 +1,68 @@
+name: Build containers
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - nick/fhir-converter-container
```
DanPaseltiner, line 8: Should this just be `main`

DanPaseltiner: Also, should this run on every push to `main`, or should we restrict it to only when there are changes to `containers/`? Given that we are going to build from PyPi and not GitHub directly changes elsewhere in the repo won't affect the image.

.github/workflows/test.yaml
```diff
@@ -61,6 +60,51 @@ jobs:
...
+      - name: Install Pytest
+        run: pip install pytest
+      - name: Install dependencies
+        working-directory: ./containers/${{matrix.container-to-test}}
+        run: |
+          pip install -r requirements.txt
+      - name: Run unit tests for containers
+        working-directory: ./containers/${{matrix.container-to-test}}
+        run: |
+          python -m pytest
```
DanPaseltiner, line 110: Should we make building containers dependent on theses tests passing? No need to waste time build images for broken services.

DanPaseltiner (changes requested):
This looks great. I have a couple questions on our CI/CD that I'd like to work through before merging. Thanks Nick!

.github/workflows/buildContainers.yaml
nickclyde: I only added this branch so that I can test the workflow and get it working - will remove it to be just main before merging! But that is a good call about restricting by path, I will add that in.

.github/workflows/test.yaml
nickclyde: Good idea! I can add that in 👍

DanPaseltiner (approved):
Looks great!

.github/workflows/test.yaml
nickclyde: On second thought, I think the filter of only running on main will effectively accomplish this, since we should never merge to main if tests are failing, so no need to explicitly set dependencies.

containers/fhir-converter/FHIR-Converter-Installation-And-Usage-Guide.md
```diff
@@ -0,0 +1,121 @@
+# Microsoft FHIR Converter CLI Installation Guide
+This document provides a guide for installing the [Microsoft FHIR Converter](https://github.com/microsoft/FHIR-Converter) as a Command Line Interface (CLI) tool on Windows, MacOS, and Linux systems, as well as a brief introduction to using the converter.
+
+## Running via Docker
+To run the FHIR Converter in a Docker container, follow these steps:  
+  
+  1. From the `containers/fhir-converter` directory, run:  
+  `docker build -t fhir-converter .`
```
spencer-skylight, line 8: On my Fedora system (uses podman backend for docker commands), I'm getting the following error running this command.
```
DotLiquids/ValidateTests.cs(216,153): warning SA1028: Code should not contain trailing whitespace [/build/FHIR-Converter/src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Microsoft.Health.Fhir.Liquid.Converter.UnitTests.csproj]
Utilities/TemplateUtilityTests.cs(139,93): warning SA1013: Closing brace should be preceded by a space. [/build/FHIR-Converter/src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Microsoft.Health.Fhir.Liquid.Converter.UnitTests.csproj]
/usr/share/dotnet/sdk/3.1.424/Microsoft.Common.CurrentVersion.targets(4651,5): error MSB3030: Could not copy the file "/build/FHIR-Converter/src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/VXU_V04.liquid" because it was not found. [/build/FHIR-Converter/src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/Microsoft.Health.Fhir.TemplateManagement.UnitTests.csproj]
    20 Warning(s)
    1 Error(s)

Time Elapsed 00:00:18.60
Error: error building at STEP "RUN dotnet build -c Release -o output": error while running runtime: exit status 1
```
Navigating to [that file](https://github.com/microsoft/FHIR-Converter/tree/main/src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles) in the repository, it seems to be prefixed with .wh.

I tried this a while back on my Ubuntu laptop, and the same thing happened (native docker I believe). Files prefixed with .wh. are called "whiteout" files. Here's some [documentation](https://manpages.ubuntu.com/manpages/jammy/en/man5/aufs.5.html) about it. I've done some research, but still don't fully understand what's happening.

I've been able to work around this by checking out locally, removing the whiteout file completely. For some reason my base system does not treat the file specially, and removing it works just like any other file. This makes me believe this has something to do with the docker/podman storage driver that's treating the .wh. files differently. 

Anyhow, I updated the DockerFile to copy the working copy from my local machine rather than checking it out from GitHub, and was able to build the solution that way.

I'm not sure if this needs to hold things up, but I wanted to communicate and document the issue. I can also create an issue so we can track it.

spencer-skylight (changes requested):
I had some trouble building this on a Linux system which we may want to address now or later.  I'm happy to pair on it for troubleshooting.

Could you also update the following documentation in the SDK code that references this? I've got some placeholders for this documentation in the [conversion code](https://github.com/CDCgov/phdi/blob/73fb819ebf8f7a6766c7d52ff72634841de4aa19/phdi/fhir/conversion/convert.py#L59) and [conversion tutorial](https://github.com/CDCgov/phdi/blob/73fb819ebf8f7a6766c7d52ff72634841de4aa19/tutorials/conversion-tutorial.md?plain=1#L13).

containers/fhir-converter/FHIR-Converter-Installation-And-Usage-Guide.md
nickclyde: @spencer-skylight I just pushed a commit that adds a line to the Dockerfile which removes that `.wh.` prefix from the file. Can you pull the latest and try building again?

spencer-skylight: @nickclyde , sorry no luck. Here's the error on the updated branch:
```
(phdi) [spence@atlantis fhir-converter]$ docker build --no-cache -t fhir-converter .
[1/2] STEP 1/6: FROM mcr.microsoft.com/dotnet/sdk:3.1 AS build
[1/2] STEP 2/6: RUN git clone https://github.com/microsoft/FHIR-Converter.git --branch v5.0.4 --single-branch /build/FHIR-Converter
Cloning into '/build/FHIR-Converter'...
Note: checking out '4003b2f4b54157f6146230a9d09e54d8e3ce5064'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

--> 6ea8b8032e4
[1/2] STEP 3/6: WORKDIR /build/FHIR-Converter
--> a4c4234c8aa
[1/2] STEP 4/6: RUN mv ./src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid ./src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/VXU_V04.liquid
mv: cannot stat './src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid': No such file or directory
Error: error building at STEP "RUN mv ./src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid ./src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/VXU_V04.liquid": error while running runtime: exit status 1
```

nickclyde: Yeah, since we are able to build this in a GitHub workflow okay and push to GitHub's container registry, I think we're okay to merge for now and take a look at this issue later as tech debt. Thanks for calling this out though! Such a strange issue.

PR #84, opened 2022-10-14
Create generate search urls function

spencer-skylight:
# Description
This adds a function `_generate_search_urls` that accepts a schema, and outputs a set of initial search URLs.


phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,57 @@ def _generate_search_url(
...
+      * table_2A: search_string_2A
+      * ...
+    * dataset_B
+      * table_1B: search_string_1B
+      * table_2B: search_string_2B
+      * ...
+    * ...
+
+    :param schema: The schema to parse and create search_strings
+    :raises ValueError: If any table does not contain a `search_string` entry
```
bamader, line 239: nit: I think our style guide says we need periods at the end of these, right? (I'm still getting used to it too lol)

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,57 @@ def _generate_search_url(
...
+    :return: A dictionary containing search URLs.
+    """
+    since = None
+    count = None
```
bamader, line 243: I don't think this is actually hurting anything, but since `since` and `count` get defined in lines 254-55 to be either dataset metadata values or `*_top` values, are these lines necessary?

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,57 @@ def _generate_search_url(
...
+
+    metadata = schema.get("metadata", {})
+    since_top = metadata.get("since")
+    count_top = metadata.get("count")
+
+    for dataset_name, dataset in schema.get("datasets", {}).items():
+        dataset_metadata = dataset.get("metadata", {})
+
+        since = dataset_metadata.get("since", since_top)
+        count = dataset_metadata.get("count", count_top)
```
bamader, line 255: Nit: This is going to sound really dumb, but can we define these two variables (as well as `since_top` and `count_top` in lines 248-49) in the order that `_generate_search_url` accepts them? I totally believe that we're passing them in correctly, but seeing them defined in one order and passed in another is really making me second guess that.

tests/assets/valid_schema.yaml
```diff
@@ -0,0 +1,129 @@
+metadata:
+  count: 1000
+datasets:
+  dataset A:
+    metadata:
+      since: "2020-01-01T00:00:00"
+    tables:
+      table 1A:
+        metadata:
+          search_string: Patient
```
bamader, line 10: Out of curiosity, why do we call this a search string instead of a resource? Could it ever be something other than a resource? If not, we might want to consider renaming it to avoid introduction of another term of art.

tests/assets/valid_schema.yaml
```diff
@@ -0,0 +1,129 @@
...
+            selection_criteria: first
+            new_name: report_conclusion
+            pandas_datatype: object
+
+
+
+    
+
+
+
```
bamader, line 129: lol nit: maybe only blank line here 😂

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -277,3 +279,73 @@ def test_generate_search_url():
...
+        == f"{test_search_url_3}"
+        + f"{urllib.parse.quote('&_count=10&_since=2022-01-01T00:00:00', safe='?&=')}"
+    )
+
+
+@mock.patch("phdi.fhir.tabulation.tables._generate_search_url")
+def test_generate_search_urls(patch_generate_search_urls):
+    patch_generate_search_urls.side_effect = (
+        lambda search, count, since: f"{search}||{count}||{since}"
+    )
```
bamader, line 315: I want to make sure I understand what's going on here because we've run into trouble with tests before: what exactly does the `.side_effect` property do? I know it works on the mocked object/function, and my current understanding is that it's the "function" that executes in place of a mocked function (whereas `.return_value` is what the mocked function would theoretically return), but is that right?

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -277,3 +279,73 @@ def test_generate_search_url():
...
+
+    schema = yaml.safe_load(
+        open(
+            pathlib.Path(__file__).parent.parent.parent / "assets" / "valid_schema.yaml"
+        )
+    )
```
bamader, line 321: Just a note here (and nothing at all you did wrong): we defined a `test_schema.yaml` in the assets directory, and comparing it to this one, the two look like they have different formatting. So much so that when we try to `yaml.safe_load` test_schema, we have to have an additional line in the tests that says `schema = loaded_schema["my_table"]` (line 48 in the `test_apply_schema_to_resource` test for example). Is there a canonical way we're expecting users to format these YAML files? In either case it might be worth standardizing how we expect to write those so that we can test loading files in the same way.

bamader (commented):
Code functionality looks good to me. Just a few comments, questions, and nits, but no major functionality that needs changed.

tests/assets/valid_schema.yaml
m-goggins: My understanding is that search_string can specify more than just the resource, e.g., `Patient?birthdate=2000-01-01T00:00:00` to get Patient resources whose birthdates are 2000-01-01

m-goggins (approved):

tests/fhir/tabulation/test_fhir_tabulation.py
spencer-skylight: Yes, the purpose of this is to ensure we are calling `_generate_search_url` with specific parameter values, but does not assert what the actual result of that call should be. The purpose of this is to target test cases on the specific function they are meant to test, not its sub-functions (which are tested separately elsewhere).

tests/assets/valid_schema.yaml
spencer-skylight: That's correct. Further down in the file, there is a search_string with a query parameter attached. It was malformed, and has now been fixed, so thanks for pointing that out!

tests/assets/valid_schema.yaml
spencer-skylight: LOL - fixed!

bamader (approved):
Just a documentation change for the `generate_search_urls` function to reflect the new schema structure, but approving so that once that's done you can merge because everything else looks good

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,48 @@ def _generate_search_url(
...
+    for each table, in the following structure:
+    * dataset_A
+      * table_1A: search_string_1A
+      * table_2A: search_string_2A
+      * ...
+    * dataset_B
+      * table_1B: search_string_1B
+      * table_2B: search_string_2B
+      * ...
+    * ...
```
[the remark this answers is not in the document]
spencer-skylight: Good catch - updated.

tests/assets/valid_schema.yaml
spencer-skylight: This has been further updated with changes to a new schema structure (as outlined in our design document).

PR #85, opened 2022-10-14
Add tabulate function and tests

bamader:
This PR adds functionality to tabulate data that has been extracted from the FHIR server. This will eventually replace the part of the `generate_all_tables` function that incrementally appends resource-level values into a list of data.


phdi/fhir/tabulation/tables.py
```diff
@@ -81,6 +86,37 @@ def apply_schema_to_resource(resource: dict, schema: dict) -> dict:
...
+            resource.get("resource", {}), schema
+        )
+        if values_from_resource != {}:
+            data.append(values_from_resource)
+
+    # Could use DataFrame.from_records, but .json_normalize accounts for
+    # any nested dicts if they happen to be returned and automatically
+    # uses key/column indexing
+    data = pd.json_normalize(data)
+    data = data.replace(r"^\s*$", np.nan, regex=True)
```
spencer-skylight, line 116: I spoke with Bryan about this late last week and decided we don't actually want to convert to pandas at this stage - just return the list of dicts. I think you would just need remove these lines - do you think the NaN handling could wait until we add support for `include_unknowns`?

phdi/fhir/tabulation/tables.py
```diff
@@ -48,12 +48,14 @@ def _apply_selection_criteria(
...
+    and a schema. The keys of the created dict are the "new names"
+    for the fields in the given schema, and the values are the
+    elements of the given resource that correspond to these fields.
+    Here, `new_name` is a property contained in the schema that
+    specifies what a particular variable should be called. If a
+    schema can't be found for the given resource type, the raw
+    resource is instead returned.
 
     :param resource: A FHIR resource on which to apply a schema.
     :param schema: A schema specifying the desired values to extract,
```
m-goggins, line 61: Add :return: param

phdi/fhir/tabulation/tables.py
```diff
@@ -81,6 +84,46 @@ def apply_schema_to_resource(resource: dict, schema: dict) -> dict:
...
+    headers = None
+    for resource in extracted_data.get("entry", []):
+        values_from_resource = apply_schema_to_resource(
+            resource.get("resource", {}), schema
+        )
+
+        # apply_schema_to_resource ensures that all desired table fields
+        # exist in the first element
+        if headers is None:
+            headers = sorted(list(values_from_resource.keys()))
```
m-goggins, line 116: For my own understanding, why are we sorting the headers?

phdi/fhir/tabulation/tables.py
bamader: Sorry, what is this in reference to? The `apply_schema_to_resource` function that this re-formatted comment applies to already has a `:return:`, and the function I added in `tabulate_data` was written with a `:return:` as well.

phdi/fhir/tabulation/tables.py
m-goggins: Got it, thank you for explaining. I think you're definitely right about `sorted() `adding clarity. I could be wrong on this, but I was under the impression that `.keys()` returns returns the list in insertion order as of [Python 3.7](https://docs.python.org/3/library/stdtypes.html?highlight=keys#dict-views)(?).

For 2, I am wondering if users are expecting the data to come back in roughly the same order that they would see in a resource rather than alphabetical. I am okay with leaving it as alphabetical until we have user research, but maybe we could add it to our list of research questions for this BB? 

side note: Should we have a doc to store research questions as we make modifications to a BB?

bamader: I looked up the return type and found `set()` but it very well could be list (version changes and all that).

Re: 2: I personally think alphabetical is the best ordering for the end-user (particularly given that FHIR's nested structure could lead to differing placement of some nested fields in unexpected place), but I think you raise a good point that users might have a different expectation. I would love to get that added to some research docs, but in the mean time I vote for leaving it at alphabetical.

m-goggins: Sounds good to me!

m-goggins (approved):

spencer-skylight (changes requested):
The changes with the new schema we discussed have been merged! This is ready to refactor.

phdi/fhir/tabulation/tables.py
```diff
@@ -81,6 +84,46 @@ def apply_schema_to_resource(resource: dict, schema: dict) -> dict:
...
+    Turns data extracted from a FHIR server into a tabular format for
+    downstream processing. Accepts a schema to apply to the extracted
+    data, which is transformed into a list of lists. The first entry
+    in the returned list is a list of headers for the new table, and
+    subsequent lists are each an extracted data element represented
+    in a tabular format.
+
+    :param extracted_data: The content response from the FHIR server
+      holding retrieved data to tabulate.
+    :param schema: The schema of columns and values to apply to the
```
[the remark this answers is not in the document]
bamader: I actually rewrote the code in a way that allows `tabulate_data` to just take the whole schema and let `apply_schema_to_resource` do the work of sifting through particular columns. That way we have a single workhorse function responsible for identifying resource types, columns, and the like from the schema as they apply to FHIR resources.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,89 @@ def _apply_selection_criteria(
...
+        if not resource_type:
+            raise ValueError(
+                "Each table must specify resource_type. "
+                + f"resource_type not found in table {table_name}."
+            )
+
+        # We only care about the schema that matches the resource
+        if resource.get("resourceType", "") == resource_type:
+            for field in table.get("columns", []).keys():
+                col_in_table = field.lower().strip().replace(" ", "_")
```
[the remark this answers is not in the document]
bamader: I would bet there's a pretty rare output type that _shouldn't_ just have lower case and underscores used as the default. I think we could change this in the future if we needed, but I would honestly say from an analyst's perspective it's probably best practice to use lowercase column names in a table and to avoid white spaces.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,89 @@ def _apply_selection_criteria(
...
+
+        if not resource_type:
+            raise ValueError(
+                "Each table must specify resource_type. "
+                + f"resource_type not found in table {table_name}."
+            )
+
+        # We only care about the schema that matches the resource
+        if resource.get("resourceType", "") == resource_type:
+            for field in table.get("columns", []).keys():
```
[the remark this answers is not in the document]
bamader: Great calls on both points 😎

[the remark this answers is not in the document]
bamader: Yeah we probably should. The way I've been thinking about it mentally is that a "field" is what the property is called when it's associated with a FHIR resource or otherwise not yet in a tabular format. It becomes a "column" when put into a table or otherwise tabulated. I don't know if we need to write that anywhere in the code, but at the very least that kind of usage should probably be something we make consistent.

phdi/fhir/tabulation/tables.py
```diff
@@ -98,8 +152,8 @@ def generate_table(
     :param cred_manager: The credential manager used to authenticate to the FHIR server.
     """
     output_path.mkdir(parents=True, exist_ok=True)
-    for resource_type in schema:
-
+    for _, table in schema.get("tables", {}).items():
```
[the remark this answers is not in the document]
bamader: I believe that would make `table` a _key_ to the dictionary of tables instead of a value. So to access an actual table in the schema's table set, I would have to write `schema["tables"]["table"]`. I went with the `_, table` approach coupled with `.items()` because the underscore means the keys to `schema["table"]` won't be saved at all, but the values will still be stored in `table` so I don't have to access anything else nested.

phdi/fhir/tabulation/tables.py
```diff
@@ -164,11 +218,9 @@ def generate_all_tables_in_schema(
 
     schema = load_schema(schema_path)
 
-    for table in schema.keys():
-        output_path = base_output_path / table
-        generate_table(
-            schema[table], output_path, output_format, fhir_url, cred_manager
-        )
+    for _, table in schema.get("tables", {}).items():
```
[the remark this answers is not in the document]
bamader: same as above comment

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
spencer-skylight: OK, you're probably right.

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
spencer-skylight: You're absolutely right. What about using `schema.get("tables", {}).values()`?

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
spencer-skylight: By this way of thinking, it should probably be called a "field" in the schema. I'm thinking we should put it to the SDK team to debate and make a decision. Thoughts?

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,89 @@ def _apply_selection_criteria(
...
+        resource_type = table.get("resource_type")
+
+        if not resource_type:
+            raise ValueError(
+                "Each table must specify resource_type. "
+                + f"resource_type not found in table {table_name}."
+            )
+
+        # We only care about the schema that matches the resource
+        if resource.get("resourceType", "") == resource_type:
```
spencer-skylight, line 77: I think we could technically have two tables defined in the same schema for the same resource, which would cause problems for this approach.  For example, what if we want to have a table for lab observations, and another for growth chart observations? I think it would be better to leave table selection/identification to the calling function.

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
spencer-skylight: By this way of thinking, it should probably be called a "field" in the schema. I'm thinking we should put it to the SDK team to debate and make a decision. Thoughts?

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
bamader: Sure, I can do that

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
bamader: Oh I guess I think of it as more columnar-related in the schema. I think calling it column in the schema makes sense more from an intuitive perspective, but the schema is used to define the _columns_ that will occur in the output table. Maybe what this conversation is revealing is that we might need to have a larger conversation with the SDK team as you proposed, but that might also be something Bryan is involved in for nomenclature and user accessibility reasons--would you be okay not counting this then as a blocker to this PR?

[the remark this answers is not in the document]
spencer-skylight: Nope, not a blocker. I like thinking on next steps.

phdi/fhir/tabulation/tables.py
bamader: That's a really fair observation. I spent some time thinking about a programmatic splitting solution to this problem last night before I realized you were probably talking about something simpler. I thought you were asking for an `apply_schema_to_resource` function that could programmatically and intelligently decide which of two observation tables an observation resource should belong to, and that was giving me fits because it would involve a lot of pre-specifying work on the user that would defeat the point of doing this programmatically (I also didn't see any reasonably effective way to do that). But, what I think you're _actually_ asking for is that if a schema contains multiple tables for the same resource, grab all of the fields associated with all of those tables, not just the first occurring table (i.e. delete the `break` statement on line 93). This is way more effective and lets the calling function and user decide how to split up those fields after they're retrieved.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
+
+        if resource_type == "":
+            raise ValueError(
+                "Each table must specify resource_type. "
+                + f"resource_type not found in table {table_name}."
+            )
+
+        # We only care about the parts of the schema that match the resource
+        if resource.get("resourceType", "") == resource_type:
+            for column_name, column in table.get("columns", {}).items():
```
bryan-skylight, line 78: Again, here I'd recommend changing to `column_name` and `column_params`

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
+                # Use FHIR-path to identify desired value
+                path = column["fhir_path"]
+                parse_function = _get_fhirpathpy_parser(path)
+                value = parse_function(resource)
+
+                if len(value) == 0:
+                    data[col_in_table] = ""
+
+                else:
+                    selection_criteria = column["selection_criteria"]
```
bryan-skylight, line 90: Shouldn't we use `.get()` here to be consistent? What if the schema doesn't specify it?

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
+                + f"resource_type not found in table {table_name}."
+            )
+
+        # We only care about the parts of the schema that match the resource
+        if resource.get("resourceType", "") == resource_type:
+            for column_name, column in table.get("columns", {}).items():
+                col_in_table = column_name.lower().strip().replace(" ", "_")
+
+                # Use FHIR-path to identify desired value
+                path = column["fhir_path"]
```
bryan-skylight, line 82: We should use `.get()` here as well to ensure we don't get an unexpected `KeyError` if it's not present.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
+                parse_function = _get_fhirpathpy_parser(path)
+                value = parse_function(resource)
+
+                if len(value) == 0:
+                    data[col_in_table] = ""
```
bryan-skylight, line 87: Why are we using strings as our default empty value? Shouldn't this be `None`?

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
+
+                else:
+                    selection_criteria = column["selection_criteria"]
+                    value = _apply_selection_criteria(value, selection_criteria)
+                    data[col_in_table] = str(value)
```
bryan-skylight, line 92: Again here, why are we converting everything to string? If this is a carryover from the issues we were having with the Parquet writer, I'd advocate forcing the `load` function (or whatever we're going to call the function that persists the data) to deal with that. We shouldn't be preemptively passing back data that masks the datatype of a value.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
 
 def apply_schema_to_resource(resource: dict, schema: dict) -> dict:
```
bryan-skylight, line 49: This function should be renamed in my opinion. There's more to the schema than just the selection criteria, and this function is only extracting the value from the resource. I think we should rename it to something that reflects that.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
-        value = parse_function(resource)
-
-        if len(value) == 0:
-            data[resource_schema[field]["new_name"]] = ""  # pragma: no cover
-        else:
-            selection_criteria = resource_schema[field]["selection_criteria"]
-            value = _apply_selection_criteria(value, selection_criteria)
-            data[resource_schema[field]["new_name"]] = str(value)
+    for table_name, table in schema.get("tables", {}).items():
+        resource_type = table.get("resource_type", "")
```
bryan-skylight, line 68: This approach seems to reinforce the idea that a table will contain only a single resource, which isn't true. We'll need to update the schema structure and rethink this approach.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
+    return data
+
+
+def tabulate_data(extracted_data: dict, schema: dict) -> List[List]:
+    """
+    Turns data extracted from a FHIR server into a tabular format for
+    downstream processing. Accepts a schema to apply to the extracted
+    data, which is transformed into a list of lists. The first entry
+    in the returned list is a list of headers for the new table, and
+    subsequent lists are each an extracted data element represented
```
bryan-skylight, line 103: I don't think this accurately depicts what is returned. Each subsequent list should be a row, right? Meaning it would hold many extracted data elements? It might make more sense to say ", and each subsequent list contains the data extracted from a FHIR resource, according to the fields specified in the schema, which represent the rows of data" or something like that.

phdi/fhir/tabulation/tables.py
```diff
@@ -262,12 +313,13 @@ def _generate_search_urls(schema: dict) -> dict:
...
+    Removes resources from a FHIR response if the resource contains a
+    null value for fields where include_nulls is False, as specified
+    in the schema.
 
     :param response: List of resources returned from FHIR API.
-    :param schema_columns: Dictionary of columns to include in tabulation that specifies
-      which columns should include_nulls.
+    :param schema_columns: Dictionary of columns to include in tabulation
+      that specifies which columns should include_nulls.
     :param return: List of resources with removed nulls.
```
bryan-skylight, line 323: Can we make this more descriptive? Something like `:param return: A list of FHIR resources that meet the include_null specification of the schema`

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -66,17 +76,68 @@ def test_apply_schema_to_resource():
...
+    # Check all columns from schema present
+    assert tabulated_data[0] == [
+        "first_name",
+        "last_name",
+        "patient_id",
+        "phone_number",
+    ]
+
+    # Check all records in data bundle present
+    assert len(extracted_data["entry"]) + 1 == len(tabulated_data)
```
bryan-skylight, line 113: nit: Can we change this to the following?

```python
assert len(extracted_data["entry"]) == len(tabulated_data[1:])
```

Both solutions achieve the same goal, but I think this more clearly demonstrates what we're testing.

bryan-skylight (changes requested):
This is a great start, but I've got some things that I'd like to see changed before we merge this. Namely, I think we need to revisit the schema structure because I think something is getting lost in translation.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
 def apply_schema_to_resource(resource: dict, schema: dict) -> dict:
     """
-    Creates and returns a dictionary of data based on a FHIR resource and a schema. The
-    keys of the created dict are the "new names" for the fields in the given schema, and
-    the values are the elements of the given resource that correspond to these fields.
-    Here, `new_name` is a property contained in the schema that specifies what a
-    particular variable should be called. If a schema can't be found for the given
-    resource type, the raw resource is instead returned.
+    Creates and returns a dictionary of data based on a FHIR resource
+    and a schema. The given schema should define a table for each
```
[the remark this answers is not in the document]
bamader: Spencer and I talked on Tuesday when you were out and actually decided to push forward with the assumption that table = resource as a desirable one. There were a lot of angles covered, but some big points of the discussion came down to one of the following: 1) it being the user's / caller's responsibility to execute the joins and unions they want to see on their data after the data has been tabulated, or 2) if that's something we want to handle, this function as presently scoped isn't the right place to try to handle unions and joins and so it's actually advantageous from a "get things working" perspective to limit each table to holding the fields for a single resource. FWIW I also do think the schema structure has changed somewhat since you've been out so we probably need to go over that too.

I think more generally we need to have a larger conversation around what tabulation and schema work will look like because I feel like I've been given a new moving target for "schema" and "tabulation" every day this PR has been open. We're all clearly operating with some sort of different vision for what the work should be so I think we need to define that better before this PR work continues.

phdi/fhir/tabulation/tables.py
```diff
@@ -166,26 +320,165 @@ def generate_all_tables_in_schema(
...
+                # that's what the anchor will reference
+                reference_dicts[table_name][rtype][resource.get("id", "")] = resource
+
+            if rtype in resource_directions["reverse"]:
+                if rtype not in reference_dicts[table_name]:
+                    reference_dicts[table_name][rtype] = {}
+
+                # Reverse pointers are more involved: need to figure out what
+                # resource this points to
+                ref_loc = directions_by_table[table_name]["reverse"][rtype]
```
spencer-skylight, line 468: If `directions_by_table[table_name][direction][resource_type] is made into a list as suggested above, this would become:
```python
                for ref_loc in directions_by_table[table_name]["reverse"][resource_type]:
```

phdi/fhir/tabulation/tables.py
```diff
@@ -166,26 +320,165 @@ def generate_all_tables_in_schema(
...
+
+    # Build up connections table by table, since one resource could be
+    # used in multiple different tables
+    reference_dicts = {}
+    for table_name in directions_by_table.keys():
+        reference_dicts[table_name] = {}
+
+    for entry in extracted_data.get("entry", []):
+        resource = entry.get("resource", {})
+        rtype = resource.get("resourceType", "")
```
spencer-skylight, line 445: nit: Should we make a more descriptive variable name?

phdi/fhir/tabulation/tables.py
```diff
@@ -50,36 +50,190 @@ def _apply_selection_criteria(
...
+        headers = [c[0] for c in column_items]
+        data[table_name] = [headers]
+        anchor_type = (
+            schema.get("tables", {}).get(table_name, {}).get("resource_type", "")
+        )
 
+        # Second pass over just the anchor data, since that
+        # defines the table's rows
+        for anchor_id, anchor_resource in (
+            ref_dicts.get(table_name, {}).get(anchor_type, {}).items()
```
spencer-skylight, line 165: It's possible that we could have included resources that are of the anchor resource type that we would not want to build rows for. There is a field in the response `entry.search.mode` that I think we'll need to use to make sure we are selecting only the primary search results.

phdi/fhir/tabulation/tables.py
```diff
@@ -50,36 +50,190 @@ def _apply_selection_criteria(
...
+        ):
+            row = []
+
+            for _, column_params in column_items:
```
spencer-skylight, line 169: Nit: You could use column_item.values() here to avoid the throwaway variable.

phdi/fhir/tabulation/tables.py
```diff
@@ -50,36 +50,190 @@ def _apply_selection_criteria(
...
+                path_to_use = column_params["fhir_path"]
+                resource_to_use = anchor_resource
+
+                # Determine if we need to make a lookup in our
+                # first-pass reference mapping
+                if "reference_location" in column_params:
+                    [direction, ref_path] = column_params["reference_location"].split(
```
spencer-skylight, line 176: Nit: This might be a good spot to refactor code into another internal function to make this slightly more modular.

phdi/fhir/tabulation/tables.py
```diff
@@ -50,36 +50,190 @@ def _apply_selection_criteria(
...
+        # Get the columns from the schema so we always iterate through
+        # them in a consistent order
+        column_items = (
+            schema.get("tables", {}).get(table_name, {}).get("columns", {}).items()
+        )
+        headers = [c[0] for c in column_items]
+        data[table_name] = [headers]
+        anchor_type = (
+            schema.get("tables", {}).get(table_name, {}).get("resource_type", "")
+        )
```
spencer-skylight, line 160: Just a minor refactor suggestion for readability.
```suggestion
    for table_name, table_params in schema["tables"].items():
        # Get the columns from the schema so we always iterate through
        # them in a consistent order
        column_items = table_params.get("columns", {}).items()
        headers = [column_name for column_name, _ in column_items]
        data[table_name] = [headers]
        anchor_type = table_params.get("resource_type", "")
```

phdi/fhir/tabulation/tables.py
```diff
@@ -166,26 +320,165 @@ def generate_all_tables_in_schema(
...
+                )
+                referenced_resource_type = column_params.get("fhir_path").split(".")[0]
+                if direction == "forward":
+                    directions_by_table[table_name][direction].add(
+                        referenced_resource_type
+                    )
+                else:
+                    directions_by_table[table_name][direction][
+                        referenced_resource_type
+                    ] = ref_path
```
spencer-skylight, line 384: edge case - what if the same resource type is referenced twice from the same resource type? For instance:

```yaml
resource_type: Practitioner
columns:
  Interpreted Reports:
    fhir_path: DiagnosticReport.code.text
    reference_location: reverse:DiagnosticReport:resultsInterpreter
  Performed Reports: 
    fhir_path: DiagnosticReport.code.text
    reference_location: reverse:DiagnosticReport:performer
```

Thinking maybe you could make this element reference an array of `ref_path`s?
```suggestion
                    if directions_by_table[table_name][direction].get(referenced_resource_type) is None:
                        directions_by_table[table_name][direction][referenced_resource_type] = []
                    directions_by_table[table_name][direction][
                        referenced_resource_type
                    ].append(ref_path)
```

spencer-skylight (changes requested):
The code looks great - I like the structure you've created to meet the requirements! Most of the non-nitpicky stuff is just edge case handling.

The only other thing I had to point out is for @m-goggins 's invalid value handling code, I think she will need to do almost everything that `tabulate_data` does with the exception of build the `data` dictionary that gets returned at the end. Instead she'll just drop entries in the `extracted_data` parameter. The way the code is written now, I think she would need to copy and paste a lot of the tabulation logic, so we probably should refactor this to allow for better code reuse to support her changes.

phdi/fhir/tabulation/tables.py
```diff
@@ -50,36 +50,190 @@ def _apply_selection_criteria(
...
+
+                else:
+                    selection_criteria = column["selection_criteria"]
+                    value = _apply_selection_criteria(value, selection_criteria)
+                    data[col_in_table] = str(value)
+
+    return data
+
+
+def tabulate_data(extracted_data: dict, schema: dict) -> dict:
```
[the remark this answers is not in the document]
bamader: Yep you're absolutely right. Removing all referenced adjectives.

phdi/fhir/tabulation/tables.py
```diff
@@ -166,26 +320,165 @@ def generate_all_tables_in_schema(
...
+    for the column to be found in a resource that either a) references
+    a given resource, or b) is referenced *by* the given resource.
+    Since each table in the schema is defined with an  "anchor" resource
+    (the main type of resource determining the number of rows in the
+    table), referenced resources of type A can be labeled "backward"
+    pointers and referenced resources of type B can be labeled "forward"
+    pointers. This mapping is used to efficiently group and aggregate
+    related resource data for tabulation.
+
+    :param schema: A user-defined schema featuring some number
```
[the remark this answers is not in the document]
bamader: Yeah this makes sense. I'll check out how we refer to it across this file and standardize.

phdi/fhir/tabulation/tables.py
```diff
@@ -50,36 +50,190 @@ def _apply_selection_criteria(
 
 def apply_schema_to_resource(resource: dict, schema: dict) -> dict:
```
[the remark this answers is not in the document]
bamader: Yep! We no longer have need of this function. I would actually love to delete it, along with the code for `generate_schema_table` and `generate_all_tables`, because they're cluttering up this file, but I don't want to do so without some okay from @bryan-skylight and some notice to the cloud implementation workstream.

phdi/fhir/tabulation/tables.py
```diff
@@ -166,26 +320,165 @@ def generate_all_tables_in_schema(
...
+            if rtype in resource_directions["reverse"]:
+                if rtype not in reference_dicts[table_name]:
+                    reference_dicts[table_name][rtype] = {}
+
+                # Reverse pointers are more involved: need to figure out what
+                # resource this points to
+                ref_loc = directions_by_table[table_name]["reverse"][rtype]
+                ref_path = ref_loc.replace(":", ".") + ".reference"
+                parse_function = _get_fhirpathpy_parser(ref_path)
+                referenced_anchor = parse_function(resource)[0]
```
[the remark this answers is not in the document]
bamader: Good call. I totally forgot to just sub this code for the helper function I wrote anyway.

phdi/fhir/tabulation/tables.py
bamader: Sure thing 👍

phdi/fhir/tabulation/tables.py
bamader: Actually here, I don't think we can. A few lines above, we get the list of tuples for the schema in `column_items` so that we can extract the headers and then process the schema's columns in the same order. If we replaced the header extraction with a `.keys()` and this with a `.values()`, there's no guarantee that both sets come back in the same order because of the way Python stores dictionaries. So although we do have a throwaway here, I think it actually is necessary to preserve the ordering we grabbed.

phdi/fhir/tabulation/tables.py
bamader: I like this! Will implement.

phdi/fhir/tabulation/tables.py
bamader: Yeah that's a good idea, I'll break this out and add some testing around it

phdi/fhir/tabulation/tables.py
bamader: The code can actually already account for this! The part that you've commented on is _just_ responsible for creating the directions of association between the various resource types. If one resource references an anchor multiple times (in the way you illustrated), it actually doesn't matter that we're only processing one of those ref_paths. The `build_reference_dicts` just assembles the resources themselves that are connected to the anchor in an aggregated fashion--the `ref_path` is just used to figure out what the connection to the anchor is. This means that when we actually get to `tabulate`, because we're iterating back over the _schema columns_ rather than the ref_paths, the proper value can still be looked up and extracted. On the other hand, if we were to implement this change and make the reverse directions hold a list of ref_paths, the `build_reference_dicts` function would interpret those as two separate resources, which would lead to every resource linked to anchor being added to the reference dictionary twice. This in turn would cause the tabulate function to double report every value that was a reverse reference.

To back up everything that I'm saying, I just modified the test cases I have locally and ran them through. I asked for another reverse reference Observation (to make sure nothing was excluded or double counted) in the schema, and then without changing algorithm logic at all, checked the output. The code successfully added the new column with the new information from the additional observations without double counting anything. So, I think this case is already handled and not a concern with the algorithmm.

bamader:
In addition to the re-review, we also have to decide if we want to delete `apply_schema_to_resource` and the related `generate_table` functions here as well. I would be in favor of removing the former but leaving the latter until the write functionality is also in place.

phdi/fhir/tabulation/tables.py
spencer-skylight: We created a ticket to address this at a later date:
https://app.clickup.com/t/3rgqa07

spencer-skylight (approved):
LGTM!

spencer-skylight:
> In addition to the re-review, we also have to decide if we want to delete apply_schema_to_resource and the related generate_table functions here as well. I would be in favor of removing the former but leaving the latter until the write functionality is also in place (not least because not only do we not use the apply_schema_to_resource function for anything productive any more, but maintaining is also forcing us to maintain cluttered and out dated tests assets).

My vote is to restore `apply_schema_to_resource` to it's original code, and delete all old code as a part of the [ticket](https://app.clickup.com/t/3pc6e55) to re-implement `generate_all_tables_in_schema`. While I don't think it's strictly necessary to back out the changes this PR made to `apply_schema_to_resource`, but doing so would make things cleaner.

phdi/fhir/tabulation/tables.py
```diff
@@ -48,35 +48,88 @@ def _apply_selection_criteria(
...
-        parse_function = _get_fhirpathpy_parser(path)
-        value = parse_function(resource)
-
-        if len(value) == 0:
-            data[resource_schema[field]["new_name"]] = ""  # pragma: no cover
-        else:
-            selection_criteria = resource_schema[field]["selection_criteria"]
-            value = _apply_selection_criteria(value, selection_criteria)
-            data[resource_schema[field]["new_name"]] = str(value)
+    for table_name, table in schema.get("tables", {}).items():
```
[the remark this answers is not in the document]
bryan-skylight: This conversation has been marked as resolved, but neither of the following two conditions were met:

1. The change was implemented, or
2. An explanation of why the change won't be implemented is provided.

I know we've never officially declared those as criteria for any kind of official process, but I think it's good practice to not simply resolve comments without one of those two criteria taking place.

phdi/fhir/tabulation/tables.py
bryan-skylight: Same comment as above about resolved conversations with no indication of why.

phdi/fhir/tabulation/tables.py
bryan-skylight: Same comment as above about unresolved comments.

phdi/fhir/tabulation/tables.py
bryan-skylight: Same comment above about unresolved comments.

phdi/fhir/tabulation/tables.py
bryan-skylight: Same comment as above about unresolved comments.

bryan-skylight (approved):
So, I haven't done a full analysis of the PR because GitHub seems to really struggle to show the appropriate information when PRs get this big (97 comments and 20 commits for, effectively, 2 files). If issues arise, we can fix them in the future. but this PR is too big and I just want to get it closed.

m-goggins (approved):
Looks good to me per the conversation on slack about submitting and re-adjusting as needed.

PR #87, opened 2022-10-17
Add drop_null function and tests

m-goggins:
This PR adds functionality to drop resources from tabulation process if the user specifies to not include null values.


phdi/fhir/tabulation/tables.py
```diff
@@ -181,3 +181,35 @@ def _get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
...
+    Removes resources from FHIR response if the resource contains a null value for
+    fields where include_nulls is False, as specified in the schema.
+
+    :param response: List of resources returned from FHIR API.
+    :param return: List of resources with removed nulls.
+    """
+
+    # Identify fields to drop nulls
+    nulls_to_drop = [
+        schema[key]["new_name"]
```
spencer-skylight, line 197: nit: could you rename `key` to `column` to describe what the dictionary index actually represents?

phdi/fhir/tabulation/tables.py
```diff
@@ -181,3 +181,35 @@ def _get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
       at `fhirpath_expression`.
     """
     return fhirpathpy.compile(fhirpath_expression)
+
+
+def drop_null(response: list, schema: dict):
```
spencer-skylight, line 186: I suggested something to Brandon as well, but could you rename `schema` to `schema_columns`. Otherwise it's confusing what would actually be passed in here. `schema` makes it sound like the whole schema, but it's actually just a part of it.

phdi/fhir/tabulation/tables.py
```diff
@@ -181,3 +181,35 @@ def _get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
...
+    """
+    Removes resources from FHIR response if the resource contains a null value for
+    fields where include_nulls is False, as specified in the schema.
+
+    :param response: List of resources returned from FHIR API.
```
spencer-skylight, line 191: Remember to document the `schema` param.

phdi/fhir/tabulation/tables.py
```diff
@@ -181,3 +181,35 @@ def _get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
...
+
+    # Identify indices in List of Lists to check for nulls
+    indices_of_nulls = [response[0].index(field) for field in nulls_to_drop]
+
+    # Check if resource contains nulls to be dropped
+    for resource in response[1:]:
+        # Check if any of the fields are none
+        if "" not in resource:
+            continue
+        else:
```
spencer-skylight, line 210: Can you explain why this is needed? Since I think `in` will loop over all values in resource, I don't think this will help with performance.

spencer-skylight (changes requested):
Just a few comments.  I think you have a few things that need to be manually merged from main as well.

phdi/fhir/tabulation/tables.py
m-goggins: Ah you're right. It's not needed.

spencer-skylight (approved):
LGTM!

PR #88, opened 2022-10-18
Dan/upload fhir bundle endpoint

DanPaseltiner:
This PR adds the `/fhir/transport/http/upload_bundle_to_fhir_server` endpoint and refactors our handling of environment variables to use the Pydantic `Settings` class. Addresses this [ticket](https://app.clickup.com/t/2yv08ze).


containers/phdi-ingestion/app/config.py
```diff
@@ -0,0 +1,22 @@
+from functools import lru_cache
+from pydantic import BaseSettings
+from typing import Optional, Literal
+
+
+class Settings(BaseSettings):
+    credential_manager: Optional[Literal["azure", "gcp"]]
+    salt_str: Optional[str]
+    fhir_url: Optional[str]
```
BradySkylight, line 9: This is awesome!

containers/phdi-ingestion/app/routers/fhir_transport_http.py
```diff
@@ -1,11 +1,78 @@
...
+    input = dict(input)
+    required_values = ["credential_manager", "fhir_url"]
+    search_result = search_for_required_values(input, required_values)
+    if search_result != "All values were found.":
+        response.status_code = status.HTTP_400_BAD_REQUEST
+        return search_result
+
+    input["credential_manager"] = credential_managers[input["credential_manager"]]
+
+    fhir_server_response = upload_bundle_to_fhir_server(**input)
```
BradySkylight, line 50: Can you give me more information about how this is all working and the use of **?

containers/phdi-ingestion/app/routers/fhir_transport_http.py
```diff
@@ -1,11 +1,78 @@
...
+from phdi.cloud.azure import AzureCredentialManager
+from phdi.cloud.gcp import GcpCredentialManager
+from phdi.fhir.transport import upload_bundle_to_fhir_server
 
 router = APIRouter(
     prefix="/fhir/transport/http",
     tags=["fhir/transport"],
 )
 
+credential_managers = {"azure": AzureCredentialManager, "gcp": GcpCredentialManager}
```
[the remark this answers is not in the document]
DanPaseltiner: I don't think this is quite like an enum, at least based on my understanding of them. I think of an enum as a variable with only a certain set of pre-defined allowable values.  The use of `Literal` on line 21 makes `"credential_manager"'  like an enum.

Here things are little different. The problem I'm trying to solve on line 16 is that depending on what value the user provides for `"credential_manager"` we need to instantiate the appropriate credential manager class (e.g. `AzureCrendentialManager` if they provide a value of `"azure"' for `"credential_manager"'. We could solve this with an `if/elif/else` block, but that is a lot of lines and a little cumbersome to maintain as we develop new cred manager classes. Instead, line 16 creates a dictionary with keys corresponding to the allowable values of `"credential_manager"` (see line 21). The value associated with each of these keys is the specific credential manager class that `"credential_manager"` codes for. Then on line 48 we use this dictionary to easily get the correct credential manager class based on the value of `"credential_manager"` without having to use and `if` block. Kind of neat that the objects stored in Python dicts can can be classes?

I hope this helps some. Totally happy to discuss more here, or get on a call.

containers/phdi-ingestion/app/routers/fhir_transport_http.py
DanPaseltiner: On line 48 we use the `credential_managers` dict created on line 16 to replace the string value of the `"credential_manager"` key in input with the actual class. If the user specified `"credential_manager": "azure"` line 48 would change this key value pair to `"credential_manager": AzureCredentialManager`. This is important for calling `upload_bundle_to_fhir_server` on line 50.

Using `**` "unpacks" a dictionary and provides each key:value pair as an argument for the function. For example, let use the simple function below:

```
def simple_function(a, b):
    return a + b
```

We can call this function like this,
```
#Option 1
simple_function(1, 2)
```

or like this:

```
#Option 2
simple_functions(a=1, b=2)
```

Additionally, we can also call it by unpacking the values of dictionary using ** like this:

```
#Option 3
my_dict = {"a": 1, "b":2}
simple_function(**my_dict)
```

Option 3 is equivalent to option 2. Here are [some docs](https://python-reference.readthedocs.io/en/latest/docs/operators/dict_unpack.html) on the unpack operator for dictionaries.

PR #89, opened 2022-10-18
Add drop_invalid function and tests

m-goggins:
This PR adds functionality and tests for dropping unknown resources from FHIR responses when tabulating data; it is virtually the same as the drop_null PR (#87). I will need to do some manual merges to resolve conflicts.


tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -277,3 +278,38 @@ def test_generate_search_url():
         == f"{test_search_url_2}"
         + f"{urllib.parse.quote('?_count=10&_since=2022-01-01T00:00:00', safe='?&=')}"
     )
+
+
+def test_drop_unknown():
+
+    schema = yaml.safe_load(
+        open(
+            pathlib.Path(__file__).parent.parent.parent / "assets" / "test_schema.yaml"
```
spencer-skylight, line 287: Could you code this against `valid_schema.yaml`, which has the new schema structure? You may also need to update the drop_null logic as well.

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,33 @@ def _generate_search_url(
...
+        if not schema_columns[column]["include_unknowns"]
+    ]
+
+    # Identify indices in List of Lists to check for unknowns
+    indices_of_unknowns = [response[0].index(field) for field in unknowns_to_drop]
+
+    # Check if resource contains unknowns to be dropped
+    for resource in response[1:]:
+        for i in indices_of_unknowns:
+            if not resource[i]:
```
spencer-skylight, line 247: I think this would also capture null-string resources.  This should probably be `if resource[i] is None:`.

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,33 @@ def _generate_search_url(
...
+        schema_columns[column]["new_name"]
+        for column in schema_columns.keys()
+        if not schema_columns[column]["include_unknowns"]
+    ]
+
+    # Identify indices in List of Lists to check for unknowns
+    indices_of_unknowns = [response[0].index(field) for field in unknowns_to_drop]
+
+    # Check if resource contains unknowns to be dropped
+    for resource in response[1:]:
```
spencer-skylight, line 245: This will loop over all resources which will be expensive.  I think we should only looping over data if `len(indices_of_unknowns)>0`... That way, if we know we don't want to drop any data, we don't have to scan the dataset.

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -277,3 +278,38 @@ def test_generate_search_url():
...
+
+    # Keeps all resources because include_nulls all False
+    responses_no_nulls = drop_unknown(
+        fhir_server_responses_no_unknowns, schema["my_table"]["Patient"]
+    )
+    assert len(responses_no_nulls) == 3
+    assert responses_no_nulls[1][3] == fhir_server_responses_no_unknowns[1][3]
+
+    # Drop null resource
+    fhir_server_responses_1_null = [
```
spencer-skylight, line 305: Additional suggested test cases:
* Columns that should not drop nulls do not cause rows to be dropped for unknown values.
* Null (empty string) values are not dropped.

spencer-skylight (changes requested):
Looking good! I had a few small fixes and suggested optimizations.

phdi/fhir/tabulation/tables.py
m-goggins: I see your point. My understanding was that this function would only ever be called if we already knew from the schema that the user wanted to drop at least 1 column if unknown, but maybe that assumption is wrong?

spencer-skylight: That's a fair point. I think we can do that in our code, but the user could also make use of the function directly and we're not sure what they will do. It's probably reasonable to expect they would do the same check, but also an easy change for us to make and it could improve perception of performance in PHDI. Since it's simple and I'd argue the pros outweigh the cons, I'd probably just add it, but I don't think it's a blocker.

tests/fhir/tabulation/test_fhir_tabulation.py
m-goggins: Will do!

tests/fhir/tabulation/test_fhir_tabulation.py
m-goggins: I understand the 2nd bullet and implemented, but I'm not understanding the 1st one. Is that for the drop_null function; don't drop unknowns when we intend to only drop nulls?

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,34 @@ def _generate_search_url(
...
+    ]
+
+    # Identify indices in List of Lists to check for unknowns
+    indices_of_unknowns = [response[0].index(field) for field in unknowns_to_drop]
+
+    # Check if resource contains unknowns to be dropped
+    if len(indices_of_unknowns) > 0:
+        for resource in response[1:]:
+            for i in indices_of_unknowns:
+                if resource[i] is None:
```
bryan-skylight, line 248: This is a nice, clean implementation, and I really like it! I think this line, though, would effectively make this function the same as `drop_null`, right? I think this function is the more complicated one because we need a set of known values that constitute "unknown" such as "UNK", "UNKNOWN", "Did not ask", "DNA", etc. I'm wondering if we add a parameter that allows a STLT to specify what they consider to be "unknown" and use those values to filter.

tests/assets/valid_schema.yaml
```diff
@@ -0,0 +1,51 @@
+incremental_query_count: 1000
```
bryan-skylight, line 1: I believe Spencer just merged a PR that includes this file as well, so I'd recommend making sure to merge `main` into your branch when you make updates.

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -277,3 +278,56 @@ def test_generate_search_url():
...
+
+    # Keeps all resources because include_unknowns all False
+    responses_no_unknowns = drop_unknown(
+        fhir_server_responses_no_unknowns,
+        schema.get("tables").get("table 1A").get("columns"),
+    )
+    assert len(responses_no_unknowns) == 3
+    assert responses_no_unknowns[1][3] == fhir_server_responses_no_unknowns[1][3]
+
+    # Drop null resource
```
bryan-skylight, line 305: I'm not sure that `drop_unknown` should drop resources if a field is `None`. See my comment above about this.

bryan-skylight (changes requested):
Thanks for putting this together. I really like your implementation for dropping resources; very clean! I do have some questions about the approach though which I think we'll need to clear up before merging.

phdi/fhir/tabulation/tables.py
```diff
@@ -219,3 +219,34 @@ def _generate_search_url(
         return search_url_prefix
 
     return "?".join((search_url_prefix, urlencode(query_string_dict, doseq=True)))
+
+
+def drop_unknown(response: list, schema_columns: dict):
```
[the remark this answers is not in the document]
m-goggins: Makes sense to me and I will keep this in mind for future functions

phdi/fhir/tabulation/tables.py
m-goggins: It is very similar to drop_null but not exactly the same. After doing some research and checking with Spencer, we determined that an unknown value in a FHIR server would return as None (or really just not present, which would become None in a dictionary) whereas a null value would return as an empty string. 

To your second point, I think that is really important. Perhaps we could have an additional field in the schema where users can list the unknown values for a particular column?

tests/assets/valid_schema.yaml
m-goggins: Yeah, I forgot to do that until afterwards.

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,37 @@ def _generate_search_urls(schema: dict) -> dict:
     return url_dict
 
 
-def drop_null(response: list, schema_columns: dict):
+def drop_invalid(data: list, schema_columns: dict):
```
spencer-skylight, line 301: Ideally the declared data type should be more specific: `List[list]`. Also add declared data type for return value to the method signature.

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,37 @@ def _generate_search_urls(schema: dict) -> dict:
...
     """
-    Removes resources from FHIR response if the resource contains a null value for
-    fields where include_nulls is False, as specified in the schema.
+    Removes resources from FHIR response if the resource contains an invalid value for
```
spencer-skylight, line 303: Could you add a little more to the documentation about how `invalid_values` is defined and used?

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,37 @@ def _generate_search_urls(schema: dict) -> dict:
...
-
-    # Identify fields to drop nulls
-    nulls_to_drop = [
-        schema_columns[column]["new_name"]
+    # Identify columns to drop invalid values
+    invalid_columns_to_drop = [
+        column
         for column in schema_columns.keys()
-        if not schema_columns[column]["include_nulls"]
+        if not schema_columns[column]["include_invalid"]
```
spencer-skylight, line 316: I think it _should_ be safe to check `schema_columns[column]` directly here because of the way you are looping over `schema_columns`. However, should we allow a default value? i.e., `if not schema_columns[column].get("include_invalid",True)`

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,37 @@ def _generate_search_urls(schema: dict) -> dict:
...
     ]
 
-    # Identify indices in List of Lists to check for nulls
-    indices_of_nulls = [response[0].index(field) for field in nulls_to_drop]
+    # Identify indices in List of Lists to check for invalid values
+    indices_of_invalids = {}
+    for column in invalid_columns_to_drop:
+        indices_of_invalids[data[0].index(column)] = schema_columns[column][
```
spencer-skylight, line 322: I think we'll only try to pull "invalid_values" when "include_invalid" is False. In this case, we'd expect "invalid_values" to be set to something. If the schema didn't set "invalid_values", I think the code would raise a generic `KeyError`. Do you think we should check for this case and raise an error that is more descriptive to the user, like...
```python
try:
  indices_of_invalids[data[0].index(column)] = schema_columns[column]["invalid_values"]
except KeyError as ke:
  raise KeyError(f"Schema column {column} must define 'invalid_values' because 'include_invalid' is set to False") from ke
```

spencer-skylight (commented):
I think this is a great change in approach!

Something to consider: if we just apply filtering when "invalid_values" is set, we could probably avoid having the "include_invalid" boolean setting. Does keeping "include_invalid" make things more clear to the user?

m-goggins:
> I think this would be a good point if we were only handling user-generate schemas, but I am curious about how we might implement this in the web version. My assumption was that users would have to check a box if they want to drop invalid values and then something would prompt them to enter the invalid values. If we proceed in that direction, I think we should also make the change from "include_invalid" to "drop_invalid" in the schema. It makes more sense to me to say "Yes, I want to drop invalid values, here are the invalid values I want to drop" rather than "No, I do not want to include invalid values, but here are the ones I want to drop." Does that make sense?

phdi/fhir/tabulation/tables.py
m-goggins: Yes, I like this approach.

tests/fhir/tabulation/test_fhir_tabulation.py
spencer-skylight: Let's say we don't want to drop patients without a first name (i.e. include_invalid = True). We should make sure that patients without a first name are not dropped. The code will not drop them obviously since we are using invalid_values, which isn't read or used when include_invalid = True. But I thought we may just want to add it as an edge test case.

phdi/fhir/tabulation/tables.py
m-goggins: Yes, that's a good catch and definitely something we want to alert users to. I'll make the change.

spencer-skylight: I did some research about this earlier today, and found that adding the `as ke` and `from ke` to this is unnecessary. You could still consider raising an exception with a more detailed description. Pretty nit-picky, I wouldn't consider this a blocker.

spencer-skylight:
> I think this would be a good point if we were only handling user-generate schemas, but I am curious about how we might implement this in the web version. My assumption was that users would have to check a box if they want to drop invalid values and then something would prompt them to enter the invalid values. If we proceed in that direction, I think we should also make the change from "include_invalid" to "drop_invalid" in the schema. It makes more sense to me to say "Yes, I want to drop invalid values, here are the invalid values I want to drop" rather than "No, I do not want to include invalid values, but here are the ones I want to drop." Does that make sense?

This is a good point. I do think in the web version it _probably_ makes more sense to pair a checkbox with a list of values rather than just having a list labeled "invalid values". It's not clear in that case what they actually do. I'm good with just keeping the two separate settings.

spencer-skylight (changes requested):
So close! Sorry 2 minor things:
* Could you move the invalid test case (drop_invalid=true without invalid_values) to a new schema file? As of now, valid_schema.yaml is technically invalid.
* Add a new row where patient last name is None and make sure it is _not_ filtered since drop_invalid = false for the last name column (this will take care of my last remaining comment as well).

m-goggins:
I moved the invalid schema to a new test file called `invalid_schema_no_invalid_values` and I believe the test starting on line 415 of the test file (under `User-specified values are not dropped if drop_invalid is False`) takes care of the second test you mentioned.

spencer-skylight (approved):
LGTM!

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,44 @@ def _generate_search_urls(schema: dict) -> dict:
...
     """
-    Removes resources from FHIR response if the resource contains a null value for
-    fields where include_nulls is False, as specified in the schema.
+    Removes resources from FHIR response if the resource contains an invalid value for
+    fields where the field's value is invalid, as specified in the invalid_values field
+    in a user-defined schema. Users may provide invalid values as a list, including
+    empty string values ("") and None/null values (null).
 
-    :param response: List of resources returned from FHIR API.
+    :param data: List of resources returned from FHIR API.
```
bamader, line 308: Is this function intended to be used before or after the data is tabulated? Right now, the input `data` is a list of lists, which suggests that it's post-tabulation, but the comment for `data` says it's a list of resources returned from the FHIR API, which actually provides a bundle (a dict).

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,44 @@ def _generate_search_urls(schema: dict) -> dict:
...
     :param schema_columns: Dictionary of columns to include in tabulation that specifies
-      which columns should include_nulls.
-    :param return: List of resources with removed nulls.
+      which columns should include_invalid.
+    :param return: List of resources with invalid resources.
```
bamader, line 311: Should this be "...with invalid fields"? Resources shouldn't have invalid resources, right?

phdi/fhir/tabulation/tables.py
```diff
@@ -298,32 +298,44 @@ def _generate_search_urls(schema: dict) -> dict:
...
     ]
 
-    # Identify indices in List of Lists to check for nulls
-    indices_of_nulls = [response[0].index(field) for field in nulls_to_drop]
+    # Identify indices in List of Lists to check for invalid values
+    indices_of_invalids = {}
+    for column in invalid_columns_to_drop:
+        try:
+            indices_of_invalids[data[0].index(column)] = schema_columns[column][
+                "invalid_values"
```
bamader, line 325: Is this assuming conversion to a pandas data frame? I think Bryan said he wants to explicitly avoid `DataFrame` conversion at this point to avoid overhead.

bamader (changes requested):
Mostly there. There's a couple points of clarification in one of the docstrings that I think needs cleaned up. Also, I think we need Bryan to weigh in on whether this is an appropriate point to use pandas, as with a previous PR he's said he'd like to avoid using it to minimize overhead on users.

phdi/fhir/tabulation/tables.py
m-goggins: Yes! I changed to "List of resources with valid fields"

phdi/fhir/tabulation/tables.py
m-goggins: No, this operates without Pandas, working on the list of lists. Here I am creating a list of indices (in a list of data) to check for invalid fields. So if we have data that look like:
`
data = [
    ["Patient ID", "First Name", "Last Name", "Phone Number"],
    ["some-uuid", "John", "Doe", "123-456-7890"],
    ["some-uuid2", None, "Lastname", None],
    ["some-uuid2", "Firstname2, "Lastname2", 123-456-7890],
]
`

And I know from the schema that the user wants to drop resources if the Phone Number column is invalid (Null/None), from the code above I would know to only check index 3 for Phone Number in each of the rows of data after the headers one. Does that make sense?

phdi/fhir/tabulation/tables.py
m-goggins: No, this function is intended to happen within `extract_data_from_fhir_search_incremental`, which I am now seeing will return a list of dictionaries rather than a list of lists, as you pointed out. 


 The thinking behind inclduing this as part of  `extract_data_from_fhir_search_incremental` is that we want the functions to be modular so if someone only wants to extract data and they provide a list of invalid_values, they should be able to extract it as they want without going through tabulation. It seems that this function needs to be updated to work on a list of dictionaries rather than a list of lists.

I am still reviewing your tabulation code, and I am now seeing

phdi/fhir/tabulation/tables.py
bamader: That makes perfect sense to me! My question around pandas was based on a completely boneheaded assumption on my part--I saw a `.index()` method and was like oh that must be a data frame because nothing else in Python has _ever_ implemented a method with _that_ name 😝 . So yeah, that's on me and the code here is totally fine.

phdi/fhir/tabulation/tables.py
bamader: Ahh I see. Yeah, the data won't become a a list of lists until after tabulation (and even then, it'll actually be a dictionary where the keys are table names as defined in the schema and the values are lists of lists), so we should update this function to work on either a list of dictionaries (representing a list of resources) _or_ a FHIR bundle (which we know will always have a field `entry` which holds a list of resources).

m-goggins: After looking into the extraction via querying and considering that LAC has not indicated a need for working with FHIR-formatted data, we're moving to drop invalids after tabulation for now and revisit drop invalid values after extraction. I think my latest changes address the expected data coming out of the tabulation work.

tests/assets/valid_schema.yaml
```diff
@@ -8,30 +8,22 @@ tables:
...
           - "Unknown"
         selection_criteria: first
       Last Name:
-        fhir_path: Patient.name.family
-        drop_invalid: false
-        invalid_values:
-          - null
-          - ""
-          - "Unknown"
+        fhir_path: Patient.name.famil
```
spencer-skylight, line 23: You are missing a `y` here.

phdi/fhir/tabulation/tables.py
```diff
@@ -356,44 +356,58 @@ def _generate_search_urls(schema: dict) -> dict:
...
+            column
+            for column in schema.get("tables").get(table).get("columns")
+            if (
+                schema.get("tables")
+                .get(table)
+                .get("columns")
+                .get(column)
+                .get("invalid_values")
+            )
+        ]
```
spencer-skylight, line 388: `get()` won't throw an exception if nothing is found at that index, but it will return None. I'd suggest using a direct index for anything that you expect will always succeed, and give a default if you are handling a case that you later expect to be valued.
```suggestion
    for table in schema.get("tables",{}):
        invalid_columns_to_drop[table] = [
            column
            for column in schema["tables"][table].get("columns", {})
            if (
                schema["tables"][table]["columns"][column].get("invalid_values")
            )
        ]
```

This should
* avoid a TypeError: NoneType object is not iterable error if "tables" or "columns" are empty
* Use a slightly more readable syntax when we (developers) are confident a dict index exists
* throw a KeyError if table or column doesn't exist (which they always should since you are iterating over them)
* .get("invalid_values") will evaluate to None if it isn't values which works fine when checking for presence of that configuration

phdi/fhir/tabulation/tables.py
```diff
@@ -356,44 +356,58 @@ def _generate_search_urls(schema: dict) -> dict:
...
+    for table in invalid_columns_to_drop.keys():
+        indices_of_invalids[table] = {}
+        for column in invalid_columns_to_drop[table]:
+            indices_of_invalids[table][data[table][0].index(column)] = (
+                schema.get("tables")
+                .get(table)
+                .get("columns")
+                .get(column)
+                .get("invalid_values")
+            )
```
spencer-skylight, line 402: Similar suggestion as the dictionary indexing described above...

spencer-skylight (changes requested):
A few minor things to review in comments, and I think you can also delete the drop_nulls function. It should be replaced by this, and is no longer covered by unit tests.

spencer-skylight (approved):
LGTM!

m-goggins:
This PR might be in the same boat as the tabulate_data one, where it has many out-of-scope changes now. I am wondering if we would consider moving it forward as is and opening new tickets/PRs for further modifications. @bryan-skylight @bamader

phdi/fhir/tabulation/tables.py
```diff
@@ -690,3 +690,50 @@ def _generate_search_urls(schema: dict) -> dict:
...
+    """
+    Removes resources from tabulated data if the resource contains an invalid value, as
+    specified in the invalid_values field in a user-defined schema. Users may provide
+    invalid values as a list, including empty string values ("") and
+    None/null values (null).
+
+    :param data:  A dictionary mapping table names to lists of lists. The first list in
+        the data value is a list of headers serving as the columns, and all subsequent
+        lists are rows in the table.
+    :param schema: A schema of columns and values to apply to the
```
bryan-skylight, line 705: nit: we need a standardized description for the schema parameter. I've opened a [ticket in ClickUp](https://app.clickup.com/t/3qve7j5) for this as it's out of scope for this PR.

phdi/fhir/tabulation/tables.py
```diff
@@ -690,3 +690,50 @@ def _generate_search_urls(schema: dict) -> dict:
...
+    # Identify indices in List of Lists to check for invalid values
+    indices_of_invalids = {}
+
+    for table in invalid_columns_to_drop.keys():
+        indices_of_invalids[table] = {}
+        for column in invalid_columns_to_drop[table]:
+            indices_of_invalids[table][data[table][0].index(column)] = schema["tables"][
+                table
+            ]["columns"][column]["invalid_values"]
+
```
bryan-skylight, line 729: If I'm reading these two `for` loops correctly, it appears that we're looping over the columns in each table in the schema and identifying which columns should be dropped twice. The first loop will give us an object that looks something like the following.

```python
invalid_columns_to_drop = {"table1A": ["col1", "col3"], "table2A": [], "table3A": ["col4"]}
```

The next `for` loop is getting the invalid values for any given column. This gives us something like the following:

```python
# note that the indexes are off by 1 from the column names
indicies_of_invalids = {"table1A": {0: ["unknown"], 2: ["unknown", "unk", None]}, "table3A": {3: [None]}}
```

Both of these for loops require looping over the schema to get the information, so I'm wondering if we can just combine them into something like this:

```python
invalid_values_by_column_index = {}
for table in schema.get("tables"):
    columns = schema["tables"][table]["columns"]
    invalid_values_by_column_index[table] = {i: columns[col].get("invalid_values") for i,col in enumerate(columns) if columns[col].get("invalid_values", [])}
```

This would give us something that looks like this:

```python
invalid_values_by_column = {"table1A": {0: ["unknown"], 2: ["unknown", "unk", None]}, "table3A": {3: [None]}
```

This approach should make the code more performant (not that I think that'll be a huge issue here) and a little cleaner, and it should work with the third `for` loop below.

phdi/fhir/tabulation/tables.py
```diff
@@ -690,3 +690,50 @@ def _generate_search_urls(schema: dict) -> dict:
...
+    # Check if resource contains invalid values to be dropped
+    for table in data.keys():
+        if len(indices_of_invalids[table]) > 0:
+            for resource in data[table][1:]:
+                for i in indices_of_invalids[table].keys():
+                    if resource[i] in indices_of_invalids[table][i]:
+                        data[table].remove(resource)
+                        break
+
```
bryan-skylight, line 738: Using the above approach, this code block would change to:

```python
for table in data.keys():
    if len(invalid_values_by_column_index[table]) > 0:
        for resource in data[table][1:]:
            for index, invalid_values in invalid_values_by_column_index[table].items():
                if resource[index] in invalid_values:
                    data[table].remove(resource)
                    break
```

bryan-skylight (approved):
This looks good to me. Like I said, I really like your approach here. I have on nit-picky change I'd like to have implemented before merging (the spacing in the docstring), but otherwise it's good to go.

With that said, I also offered an alternative approach to the one you implemented. It's the same concept, and still your code, it just cleans it up a bit and removes one of the `for` loop blocks. Feel free to implement it if you'd like, or don't. I think it makes the code a little more clear and clean, but it doesn't change anything functionally.

phdi/fhir/tabulation/tables.py
m-goggins: I think it only loops over the schema once; the second time it's looping over the `invalid_columns_to_drop` list that came out of the first loop. But overall, I agree with your point that we only need to loop once and can grab all the information we need. I'm not sure what the schemas for STLTs will look like but better to be on the safer side.

phdi/fhir/tabulation/tables.py
m-goggins: Modified following your approach above

bamader (approved):
Looks great! Sorry for being a bum on this for so long.

PR #90, opened 2022-10-20
Create the geocoding container

BradySkylight:
This PR includes:
1- The container code to geocode addresses within a bundle or patient resource utilizing the PHDI SDK
2- Allows for smarty, census, or all geocoding methods - if all is selected then census and smarty are both completed
3- Tests to verify all the different possible pathways - success, errors, exceptions


nickclyde (approved):
This looks great! Awesome work Brady!

containers/phdi-ingestion/app/routers/fhir_geospatial.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+    if input.get("geocode_method") in ["census", "all"]:
+        geocode_client = CensusFhirGeocodeClient()
+
+    # Here we need to remove the parameters that are used here
+    #   but are not required in the PHDI function in the SDK
+    del input["geocode_method"]
+    del input["auth_id"]
+    del input["auth_token"]
```
DanPaseltiner, line 63: I think should use the `.pop()` method here instead of `del` because if these keys are not present, which I believe would be the case when `"geocode_method"=="census"`, we will raise an error (example below).

```
Python 3.10.6 (main, Aug  8 2022, 19:28:58) [Clang 13.1.6 (clang-1316.0.21.2.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> test = {}
>>> del test["missing-key"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'missing-key'
>>> test.pop("missing-key", None)
>>> 
```

containers/phdi-ingestion/app/routers/fhir_geospatial.py
```diff
@@ -0,0 +1,65 @@
...
+        search_result = search_for_required_values(input, required_values)
+        if search_result != "All values were found.":
+            response.status_code = status.HTTP_400_BAD_REQUEST
+            return search_result
+        geocode_client = SmartyFhirGeocodeClient(
+            auth_id=input.get("auth_id"), auth_token=input.get("auth_token")
+        )
+
+    if input.get("geocode_method") in ["census", "all"]:
+        geocode_client = CensusFhirGeocodeClient()
```
DanPaseltiner, line 57: I'm a little confused about our use of `"all"` here. The docs state that when `"gecode_method" == "all"` we will geocode with both Smarty and the US Census. It looks to me that the current behavior is:

1. If `"gecode_method" == "smarty"` we will make a geocoding client for Smarty and use it appropriately.
2. If `"gecode_method" == "census"`  we will make a geocoding client for Census and use it appropriately.
3. If `"gecode_method" == "all"` we will first make a Smarty client, overwrite it with a Census client and then only geocode with the Census.

If I'm missing something subtle here I apologize. 

Additionally, I don't think we really need the `"all"` option, although it is a cool idea. Since the geocoding process (as currently written) replaces values in  `Patient.address` we would simply replace the result of work from the first geocoder with the result from the second. What do you think about this combined with modifying `GeocodeAddressInBundleInput`?

```suggestion
    input = dict(input)

    if input.get("geocode_method")  == "smarty" :
        required_values = ["auth_id", "auth_token"]
        search_result = search_for_required_values(input, required_values)
        if search_result != "All values were found.":
            response.status_code = status.HTTP_400_BAD_REQUEST
            return search_result
        geocode_client = SmartyFhirGeocodeClient(
            auth_id=input.get("auth_id"), auth_token=input.get("auth_token")
        )

    elif:
        geocode_client = CensusFhirGeocodeClient()
```

containers/phdi-ingestion/app/routers/fhir_linkage_link.py
```diff
@@ -31,7 +31,7 @@ async def add_patient_identifier_in_bundle_endpoint(
     Add a salted hash identifier to every patient resource in a FHIR bundle using. If
     a salt is not provided in the request the value of the 'SALT_STR' environment
     variable will be used. In the case where a salt is not provided and 'SALT_STR' is
-    not defined and HTTP 500 status code is returned.
+    not defined an HTTP 500 status code is returned.
```
DanPaseltiner, line 35: Thank you for catching this! Definitely a typo I introduced.

containers/phdi-ingestion/tests/test_config.py
```diff
@@ -9,17 +8,23 @@ def test_get_settings_success():
...
 
 
 def test_get_settings_failure():
     os.environ["CREDENTIAL_MANAGER"] = "some-unknown-cred-manager"
-
     get_settings.cache_clear()
+
     with pytest.raises(ValidationError):
         get_settings()
+    os.environ.pop("CREDENTIAL_MANAGER", None)
```
DanPaseltiner, line 30: Nice!

containers/phdi-ingestion/tests/test_fhir_geospatial.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_geocode_bundle_success_census():
+    test_request = {"bundle": test_bundle, "geocode_method": "census"}
+    expected_response = copy.deepcopy(test_bundle)
+    expected_response["entry"][0]["resource"]["address"][0]["street"] = "123 Main St."
+    actual_response = client.post(
+        "/fhir/geospatial/geocode/geocode_bundle", json=test_request
+    )
+    assert actual_response.json() == expected_response
```
DanPaseltiner, line 38: Is this test actually making a call to the US Census API? I'm not sure we want our unit tests to depend on services external to PHDI. Also, if the US Census makes a change to their addresses cleaning process could it cause this test to fail? What about mocking the geocoder client and testing for how it was called?

containers/phdi-ingestion/tests/test_fhir_geospatial.py
```diff
@@ -0,0 +1,115 @@
...
+        "auth_id": "",
+        "auth_token": "",
+    }
+    os.environ["AUTH_ID"] = "test_id"
+    os.environ["AUTH_TOKEN"] = "test_token"
+    get_settings.cache_clear()
+    with pytest.raises(Exception):
+        client.post("/fhir/geospatial/geocode/geocode_bundle", json=test_request)
+    os.environ.pop("AUTH_ID", None)
+    os.environ.pop("AUTH_TOKEN", None)
```
DanPaseltiner, line 115: Do we need to test the success case for Smarty?

DanPaseltiner (changes requested):
Thanks for this Brady! I think there are a couple things we should change before merging. Happy for discussion and/or push back though!

containers/phdi-ingestion/app/routers/fhir_geospatial.py
```diff
@@ -0,0 +1,65 @@
...
+) -> dict:
+    """
+    Given a FHIR bundle and a specified geocode method, with any required
+    subsequent credentials (ie.. SmartyStreets auth id and auth token),
+    geocode all patient addresses across all patient resources in the bundle.
+
+    If the geocode method is smarty or all, then the auth_id and auth_token parameter
+    values will be used.  If they are not provided in the request then the values
+    will be obtained via environment variables.  In the case where smarty is the geocode
+    method and auth_id and auth_token are not supplied an HTTP 500 status code will
```
DanPaseltiner, line 36: Looks like we are actually returning a 400. This might be a mistake I originally introduced.

containers/phdi-ingestion/app/routers/fhir_geospatial.py
BradySkylight: That makes sense and not sure how I didn't catch that.  I'll update this.

containers/phdi-ingestion/app/routers/fhir_geospatial.py
BradySkylight: I had considered not adding the 'all' option as well and I did miss the fact that it's not making a call to the actual client until the end.  So, major miss on my part.  Now thinking about it, a user could potentially make two calls to the same geocoder if they wanted both...this is less efficient, but, we can always add the 'all' ability later down the road.

containers/phdi-ingestion/tests/test_fhir_geospatial.py
BradySkylight: Yes, I need to mock this and I need to make a mocked test for Smarty as well.

DanPaseltiner: The FHIR upload tests should have examples of very similar mocking that might be helpful.

containers/phdi-ingestion/tests/test_fhir_geospatial.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+client = TestClient(app)
+
+
+test_bundle = json.load(
+    open(pathlib.Path(__file__).parent / "assets" / "single_patient_bundle.json")
+)
+
+
```
DanPaseltiner, line 17: I think we are still missing a testing for a Smarty success.

DanPaseltiner (approved):
Approving now so I'm not blocking while on vacation, but I think we still need a test for a Smarty success.

containers/phdi-ingestion/tests/test_fhir_geospatial.py
BradySkylight: Nope, it's there on line 46-58

PR #91, opened 2022-10-20
Tabulate extraction incremental function

m-goggins:
This PR creates the `extract_data_from_fhir_search_incremental` function which generates a single batch of data along with a "next" URL from a FHIR search in order to support iteratively performing search calls and supporting incremental ETL processing.

One thing to note: I had to switch the use of `fhir_server_get` to `http_request_with_reauth` in order to test the code on the local fhir server Spencer created. The primary reason for this switch is that `fhir_server_get` requires a credential manager and will not except None; the local fhir server does not uses credentials. Spencer had talked about opening a separate ticket/PR to address the credential manager issues separately but I wanted to give reviewers a heads up before looking at the code.


phdi/fhir/tabulation/tables.py
```diff
@@ -184,6 +186,42 @@ def _get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
...
+    :param search_url: The URL to a FHIR server with search criteria.
+    :param cred_manager: The credential manager used to authenticate to the FHIR server.
+    :return: Tuple containing single page of data as a dictionary and the next URL.
+    """
+
+    # TODO: Modify fhir_server_get (and http_request_with_reauth) to function without
+    # mandating a credential manager. Then replace the direct call to
+    # http_request_with_reauth with fhir_server_get.
+    # response = fhir_server_get(url=full_url, cred_manager=cred_manager)
+    response = http_request_with_reauth(
```
spencer-skylight, line 206: I'd make a few changes to this call...
* Right now, cred_manager is passed in as None, ignoring the parameter value of cred_manager. I think we should pass in the argument value (which defaults to None) so that if a user does need to use a credential manager, they can.
* The headers parameter should be passed in with an empty dict rather than a dummy authorization token.

phdi/fhir/transport/http.py
```diff
@@ -20,7 +20,7 @@ def http_request_with_reauth(
     fails with an authorization error (HTTP status 401), obtains a new token using the
     `cred_manager`. If the original request had an Authorization header, replaces
     it with the new token and re-initiates
-    :func:`phdi.transport.http.http_request_with_retry`.
+    :func:`phdi.transport.http  .http_request_with_retry`.
```
spencer-skylight, line 23: I think these extra spaces may have been a typo?

spencer-skylight (changes requested):
Looking good! Just a few minor suggested tweaks.

phdi/fhir/tabulation/tables.py
m-goggins: Yes, good call!

phdi/fhir/transport/http.py
m-goggins: Yep!

PR #92, opened 2022-10-24
Implement extract_data_from_fhir_search

spencer-skylight:
# Description
This PR adds a function that calls `extract_data_from_fhir_search_incremental` recursively until all paginated data has been retrieved from the FHIR server via search.

# Additional Changes
* Changed signature of `extract_data_from_fhir_search_incremental` to return `List[dict], str` rather than `List[Response], str`.
* Reordered some code to put internal/helper methods at the bottom of the file.

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -7,8 +7,6 @@ from functools import cache
 from typing import Any, Callable, Literal, List, Tuple
 from urllib.parse import parse_qs, urlencode
 
-from requests import Response
-
 from phdi.cloud.core import BaseCredentialManager
 from phdi.fhir.transport import fhir_server_get, http_request_with_reauth
 from phdi.tabulation.tables import load_schema, write_table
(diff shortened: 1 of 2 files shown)


m-goggins (commented):
Looks good to me

m-goggins (approved):

phdi/fhir/tabulation/tables.py
```diff
@@ -222,6 +207,32 @@ def extract_data_from_fhir_search_incremental(
...
+
+    results, next = extract_data_from_fhir_search_incremental(
+        search_url=search_url, cred_manager=cred_manager
+    )
+
+    while next is not None:
+        incremental_results, next = extract_data_from_fhir_search_incremental(
+            search_url=next, cred_manager=cred_manager
+        )
+        results.extend(incremental_results)
```
bamader, line 231: More for my own information than a requested change, but once tabulate and write are done I'm assuming the plan will be to get rid of this `results.extend()` call in favor of incrementally processing and writing the page of data that we retrieved?

bamader (approved):
LGTM!

phdi/fhir/tabulation/tables.py
spencer-skylight: This function will actually keep the iterative request and `.extend` call. There will be another function we add after the write logic is complete to do the incremental calls that you're describing.

PR #94, opened 2022-10-24
Adapt to new schema structure

spencer-skylight:
# Description
Update our tabulation module to reflect recently agreed upon changes to schema structure

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -293,7 +293,7 @@ def _generate_search_urls(schema: dict) -> dict:
         if query_params is not None and len(query_params) > 0:
             search_string += f"?{urlencode(query_params)}"
 
-        count = table_metadata.get("incremental_query_count", count_top)
+        count = table_metadata.get("results_per_page", count_top)
         since = table_metadata.get("earliest_update_datetime", since_top)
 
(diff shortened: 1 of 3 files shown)


phdi/fhir/tabulation/tables.py
```diff
@@ -273,25 +273,28 @@ def _generate_search_urls(schema: dict) -> dict:
     """
     url_dict = {}
 
-    count_top = schema.get("incremental_query_count")
-    since_top = schema.get("earliest_update_datetime")
+    schema_metadata = schema.get("metadata", {})
+    count_top = schema_metadata.get("results_per_page")
+    since_top = schema_metadata.get("earliest_update_datetime")
```
m-goggins, line 278: I think `earliest_update_datetime` is missing in the schema_metadata, but is included in the table 1A metadata. I guess `_generate_search_urls` can always rely on the default but could be worth checking

spencer-skylight: This is by design... I wanted to test a few variations of variables which could be set at multiple levels:
* set at top level
* set at table level
* not set at all

Does this make sense?

m-goggins: Ah yes, that makes sense. I just wasn't sure since it's in the main code and not the test code.

m-goggins (approved):

PR #96, opened 2022-10-25
Implements functionality to read in schemas as JSON as well as YAML

bryan-skylight:
# Description
This PR adds the ability for users to upload their schema as a JSON file instead of YAML. It relies on the last 5 characters of the file path to determine if it's JSON or YAML, and raises a `ValueError` if it's any other file extension. The code also raises an error if it fails to parse the JSON.

# Tests
I updated the tests to account for the new file type as well as the new Exception types.

# Feedback Requested
Testing for valid YAML seems pretty hard given that YAML is a dumpster fire of a data format. Any one have ideas on how we might check for that? We'd likely want that in a new ticket/PR related to schema validation, but wanted to raise it here in case anyone has a quick fix.

### DIFF
--- a/phdi/tabulation/tables.py
+++ b/phdi/tabulation/tables.py
@@ -22,12 +22,11 @@ def load_schema(path: pathlib.Path) -> dict:
     :raises JSONDecodeError: If a JSON file is provided with invalid JSON.
     :return: A dict representing a schema read from the given path.
     """
-    path = str(path)
     try:
         with open(path, "r") as file:
-            if path.endswith(".yaml"):
+            if path.suffix == ".yaml":
                 schema = yaml.safe_load(file)
-            elif path.endswith(".json"):
+            elif path.suffix == ".json":
                 schema = json.load(file)
             else:
                 ftype = path.split(".")[-1].upper()
--- a/phdi/tabulation/tables.py
+++ b/phdi/tabulation/tables.py
@@ -29,7 +29,7 @@ def load_schema(path: pathlib.Path) -> dict:
             elif path.suffix == ".json":
                 schema = json.load(file)
             else:
-                ftype = path.split(".")[-1].upper()
+                ftype = path.suffix.replace(".", "").upper()
                 raise ValueError(f"Unsupported file type provided: {ftype}")
         return schema
     except FileNotFoundError:
--- a/phdi/tabulation/tables.py
+++ b/phdi/tabulation/tables.py
@@ -1,30 +1,46 @@
 import csv
 import os
 import pathlib
+import yaml
+import json
 import pyarrow as pa
 import pyarrow.parquet as pq
-import yaml
 
 from pathlib import Path
 from typing import Literal, List, Union
 
 
-def load_schema(path: str) -> dict:
+def load_schema(path: pathlib.Path) -> dict:
     """
     Given the path to a local YAML file containing a data schema,
     loads the file and return the resulting schema as a dictionary.
     If the file can't be found, raises an error.
 
     :param path: The file path to a YAML file holding a schema.
-    :raises Exception: If the file to be loaded could not be found.
+    :raises ValueError: If the provided path points to an unsupported file type.
+    :raises FileNotFoundError: If the file to be loaded could not be found.
+    :raises JSONDecodeError: If a JSON file is provided with invalid JSON.
     :return: A dict representing a schema read from the given path.
     """
+    path = str(path)
     try:
         with open(path, "r") as file:
-            schema = yaml.safe_load(file)
+            if path.endswith(".yaml"):
+                schema = yaml.safe_load(file)
+            elif path.endswith(".json"):
+                schema = json.load(file)
+            else:
+                ftype = path.split(".")[-1].upper()
+                raise ValueError(f"Unsupported file type provided: {ftype}")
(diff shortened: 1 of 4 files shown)


tests/tabulation/test_tables.py
```diff
@@ -19,10 +20,20 @@ def test_load_schema():
         open(pathlib.Path(__file__).parent.parent / "assets" / "test_schema.yaml")
     )
 
-    # Test invalid schema file path
-    with pytest.raises(Exception):
+    # Invalid schema file path
+    with pytest.raises(FileNotFoundError):
         load_schema("invalidPath")
 
+    # Invalid JSON
```
spencer-skylight, line 27: This looks good, but should we also add tests for a valid json schema?

spencer-skylight (commented):
This looks great! Just a few minor suggestions, some of which should probably be completed in another ticket if we decide to prioritize them.

In terms of yaml validation, pyyaml has a yaml.YAMLError that is raised when there are parsing issues. Should we add a test case for that (perhaps in another ticket/PR since this one is centered around JSON)?

I was also researching how we might validate the resulting schema structure (whether it came from yaml or json). I found this: https://pypi.org/project/schema/ 

Do you think that might be helpful? If so, we could create a ticket to add a `_validate_schema` function that can be called from `load_schema`...

phdi/tabulation/tables.py
```diff
@@ -1,30 +1,46 @@
...
+            else:
+                ftype = path.split(".")[-1].upper()
+                raise ValueError(f"Unsupported file type provided: {ftype}")
         return schema
     except FileNotFoundError:
-        raise Exception("Could not find path to given file")
+        raise FileNotFoundError(
+            "The specified file does not exist at the path provided."
+        )
+    except json.decoder.JSONDecodeError as e:
```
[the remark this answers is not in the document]
spencer-skylight: I think you're right, it's fine as is.  I created [a gist](https://gist.github.com/spencer-skylight/504c36f1dd9d08c9b39bb15749486d88) to test things out which will hopefully illustrate what I was trying to say. My concern was the user's code calling this function might not be able to see details about the deeper source of the error (in this case somewhere within `json.load`). But actually, the [default behavior](https://docs.python.org/3/library/exceptions.html#exception-context) in python is to copy the original raised exception's `__context__` to the re-raised exception.

Leaving this comment open to increase visibility on my reasoning, but I think this can probably be closed.

spencer-skylight (approved):
Looks good.

phdi/tabulation/tables.py
```diff
@@ -1,30 +1,46 @@
...
     If the file can't be found, raises an error.
 
     :param path: The file path to a YAML file holding a schema.
-    :raises Exception: If the file to be loaded could not be found.
+    :raises ValueError: If the provided path points to an unsupported file type.
+    :raises FileNotFoundError: If the file to be loaded could not be found.
+    :raises JSONDecodeError: If a JSON file is provided with invalid JSON.
     :return: A dict representing a schema read from the given path.
     """
+    path = str(path)
```
bamader, line 25: Why change the parameter signature in the function definition if we're going to cast it to a string anyway?

phdi/tabulation/tables.py
```diff
@@ -1,30 +1,46 @@
...
     try:
         with open(path, "r") as file:
-            schema = yaml.safe_load(file)
+            if path.endswith(".yaml"):
+                schema = yaml.safe_load(file)
+            elif path.endswith(".json"):
+                schema = json.load(file)
+            else:
+                ftype = path.split(".")[-1].upper()
+                raise ValueError(f"Unsupported file type provided: {ftype}")
```
bamader, line 34: How would you feel about raising this error before we ever enter the `try/except` block? It seems logically odd to me to attempt an open if we haven't validated that we support the file type at the path's location.

bamader (commented):
Some questions around exception handling and a path cast that I'd like answered but won't block around; actual functionality looks good to me.

phdi/tabulation/tables.py
bryan-skylight: Yeah, I hear ya. I could go either way. Putting it at the top has the benefit of early failing. As it's currently implemented, it fits in with the code block that's already tasked with evaluating the file extension of the provided path.

I don't feel strongly about this either way and am happy to change it if you think it makes more sense.

phdi/tabulation/tables.py
bamader: Ah gotcha. Yeah I have no problem passing in a path object, it was the cast and the fact that it used to be a string that caught me off guard (I think we pass the file as a path object everywhere else anyway so I'll pretend I asked the variant question that you came up with in the first place 😝 )

phdi/tabulation/tables.py
bamader: I don't really feel strongly about it either, moreso just wanted to raise the logic of it. I understand the idea of keeping it with code that's already tasked with extension evaluation though. Sounds like no strong feelings = no reason to move it.

phdi/tabulation/tables.py
bryan-skylight: I just added in the change to use `.suffix` instead of converting to string and using `.endswith`.

PR #97, opened 2022-10-25
Fix failing tests and tests that were running too long

BradySkylight:
This PR addresses the following issues:

1- In the container.phdi-ingestion section of PHDI the test_geospatial.py there were two tests that were passing, but were making actual calls to the Smarty third-party application, slowing the tests and not properly utilizing mocking capabilities.  The mocking has been updated so the unit test still performs proper testing, but no longer calls Smarty.
2- In the container.fhir-converter section of PHDI there were two tests that were expecting to find a path and file that weren't available on local machines.  To fix this, the Path capabilities were mocked, which fixed the issue.


nickclyde (approved):
LGTM!

PR #98, opened 2022-10-26
Create upload blob to cloud storage function and container

BradySkylight:
This PR includes the following:
1- The function within the ingestion_container to upload a blob to cloud storage
   a- this can be used in the cases where the convert to fhir fails or if the upload to fhir server fails to store the errors in a separate cloud storage
2- Tests to verify this functionality
3- Necessary code to expose this container function
4- Tests for the new utils function (get_cloud_provider) as well as the new util function get_cloud_provider.
5- Tests for the config.py and main.py


containers/phdi-ingestion/app/config.py
```diff
@@ -9,6 +9,9 @@ class Settings(BaseSettings):
     fhir_url: Optional[str]
     auth_id: Optional[str]
     auth_token: Optional[str]
+    cloud_provider: Optional[Literal["azure", "gcp"]]
+    bucket_name: Optional[str]
+    file_name: Optional[str]
```
DanPaseltiner, line 14: Given that the name of the file to be written will likely change every time the endpoint is called I'm not sure we need to support it as an env var.

containers/phdi-ingestion/app/main.py
```diff
@@ -17,6 +18,7 @@
 app.include_router(fhir_geospatial.router)
 app.include_router(fhir_linkage_link.router)
 app.include_router(fhir_transport_http.router)
+app.include_router(cloud_write_to_storage.router)
```
DanPaseltiner, line 21: What about making the route more generic like `/cloud` with a `/write_to_storage` endpoint?

DanPaseltiner (changes requested):
This is looks good to me. Just a couple small things I think we should change.

containers/phdi-ingestion/app/config.py
BradySkylight: This is a great point and reminds me of something else I should do.  The filename should be different each time and maybe not even set by the user, but we create it by default?  Or, if we do allow the user to have a say in the filename, then we should append something to the end of it to make it unique.  Thoughts?

BradySkylight: Great minds think alike!

DanPaseltiner: I still think we should remove it as an environment variable and require that a user always provide a file name.

DanPaseltiner (approved):

PR #100, opened 2022-10-28
Create extract data from schema function

spencer-skylight:
# Description
This PR creates a new function that accepts a tabulation schema, and extracts resources based on the derived search string.

# Additional Changes
The method signature for `extract_data_from_fhir_search_incremental` was updated to specify the correct return type, and the return docstring was also updated.

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -236,7 +236,7 @@ def extract_data_from_fhir_search(
 
 def extract_data_from_schema(
     schema: dict, fhir_url: str, cred_manager: BaseCredentialManager = None
-) -> Dict[str, dict]:
+) -> Dict[str, List[dict]]:
     """
     Performs a full FHIR search for each table in `schema`, and returns a dictionary
     mapping the table name to corresponding search results.
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -4,7 +4,7 @@ import random
 import pathlib
 
 from functools import cache
-from typing import Any, Callable, Literal, List, Tuple
+from typing import Any, Callable, Dict, Literal, List, Tuple
 from urllib.parse import parse_qs, urlencode
 
 from phdi.cloud.core import BaseCredentialManager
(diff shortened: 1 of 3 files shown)


phdi/fhir/tabulation/tables.py
```diff
@@ -173,15 +173,16 @@ def generate_all_tables_in_schema(
 
 def extract_data_from_fhir_search_incremental(
     search_url: str, cred_manager: BaseCredentialManager = None
-) -> Tuple[dict, str]:
+) -> Tuple[List[dict], str]:
```
bamader, line 176: I think my only suggested change is that we standardize all of the return types of all of the extract functions to be the same thing. I.e. any function whose output should be passed to tabulation as input should have the exact same format, so that tabulation can be written with one input type in mind. Currently, that format is a bundle (complete with `entry` field), but this one looks like the first parameter is a list whose tuples already have the resources extracted from the bundle. Given that we discussed needing the `entry.search.mode` parameter, I think this should probably stay as a bundle of data for each page so that tabulate can access the search parameter and continue processing everything as a bundle.

bamader (changes requested):
Almost there, I think we just need to align on what the output type of our extractions is going to be (bundle or just a list of resources).

phdi/fhir/tabulation/tables.py
spencer-skylight: Good call! The common denominator returned by extract functions is `List[dict]`, and I updated the `extract_data_from_schema` to report the right return value. 

The incremental function also returns the `next`, and the "from schema" function indexes results by table name but in all cases the actual results are managed as a `List[dict]`. Does this make sense and look good to you?

m-goggins (approved):

phdi/fhir/tabulation/tables.py
spencer-skylight: I added a ticket to ensure extract and tabulate have a consistent format to pass data between them:
https://app.clickup.com/t/3rgp6w4

bamader (approved):
Looks great, and very clear ticket and AC for instructions on reconciling the two formats moving forward!

PR #101, opened 2022-10-31
Fix the requirements doc that is causing a build to fail

BradySkylight:


DanPaseltiner (approved):
I almost flagged this my original review, but figured you had moved away from the PyPi release intentionally.

PR #102, opened 2022-10-31
Tabulation harmonize data formats

m-goggins:
This PR harmonizes the data format returned from extract functions (`extract_data_from_fhir_search_incremental` and `extract_data_from_fhir_search` to match the format expected by tabulate_data. The data output of the extraction functions and the data input of the tabulation functions are now an entry returned from a FHIR server, as a List[dict] where each dict in the list is a resource. 

I am somewhat unsure of the docstring language we should use to describe the output and input data. Also, because I had started this off of the tabulate_data branch before it had been merged into main, some of Brandon's changes there appear in the commit log.


bamader:
Going to review this soon, but did the tabulate PR not squash before merging? I feel like that should've all been condensed into one commit before it landed in main.

m-goggins:
I made a branch off of your tabulate-data branch before you merged it into main...and this is the result. I'm unsure if this is the result of something I did (very likely) or if this is what will happen as we go forward with the stacked PRs.

bryan-skylight:
> I made a branch off of your tabulate-data branch before you merged it into main...and this is the result. I'm unsure if this is the result of something I did (very likely) or if this is what will happen as we go forward with the stacked PRs.

I believe if you run the following (assuming you don't have any outstanding changes in your current branch), it should clean this up.

```bash
git checkout main
git pull origin main
git checkout tabulation-harmonize-data-formats
git merge main
git push origin tabulation-harmonize-data-formats
```

bamader (changes requested):
The signature for the extract looks right, but I think we need to change the signature for `tabulate` proper (I think that was Spencer's intention with the ticket--it could be in Slack, but I think there's a case he made in the ticket for why tabulate accepting a bundle isn't logically ideal). There are three areas of changes I can see this entailing:

1. `_build_reference_dicts` currently takes a _bundle_ of data, but we want it to take the list of entries. This should be a simple matter of changing the function signature and then altering the for loop over the entries.
2. Similar case as above, but with `tabulate_data` itself. Here, I don't think we have to change any loops within the code, just in the function signature.
3. The tests for the above two functions. Right now, we have `extracted_data = json.load(blah blah)`, which leaves `extracted_data` as a bundle. We'll just have to turn this into a list of entries before passing it into the respective functions.

Once these changes are in I think the AC will be fully done!

m-goggins:
Argh, in one of my git mishaps, I seem to have walked back on the changes that would have addressed Brandon's comments. I think they should all be in there now!

phdi/fhir/tabulation/tables.py
```diff
@@ -505,8 +505,8 @@ def extract_data_from_fhir_search_incremental(
 
     :param search_url: The URL to a FHIR server with search criteria.
     :param cred_manager: The credential manager used to authenticate to the FHIR server.
-    :return: Tuple containing single page of data as a list of FHIR resources
-      and the next URL.
+    :return: Tuple containing single page of data as a list of dictionaries and the
+        next URL.
```
bamader, line 509: nittiest of nits: I think there's two too many spaces here in this second line

bamader (approved):
LGTM!

PR #105, opened 2022-11-02
Fixed Azure Cred Manager and Cloud Storage Provide Class instantiation issue

BradySkylight:
Utils was only passing back a class and never instantiated an instance of the classes.  Azure also requires certain parameters when constructing a Provider and Cred Manager class.  I have added those parameters to all the functions where necessary, as well as the proper checks.  I also added tests and fixed the existing failing tests with this updated functionality.

### DIFF
--- a/containers/phdi-ingestion/app/routers/cloud_storage.py
+++ b/containers/phdi-ingestion/app/routers/cloud_storage.py
@@ -17,6 +17,7 @@ class WriteBlobToStorageInput(BaseModel):
     cloud_provider: Optional[Literal["azure", "gcp"]]
     bucket_name: Optional[str]
     file_name: str
+    storage_account_url: Optional[str] = None
 
 
 @router.post("/write_blob_to_storage", status_code=200)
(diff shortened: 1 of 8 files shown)


containers/phdi-ingestion/app/routers/cloud_storage.py
```diff
@@ -17,6 +17,7 @@ class WriteBlobToStorageInput(BaseModel):
     cloud_provider: Optional[Literal["azure", "gcp"]]
     bucket_name: Optional[str]
     file_name: str
+    storage_account_url: Optional[str] = None
```
DanPaseltiner, line 20: Really a nit, but what do think about this? I'd argue for setting the default value to an empty string instead of `None` so that it will be of type string which is consistent with our typing hinting.

```suggestion
storage_account_url: Optional[str] = ""
```

Here is a demonstration of the type difference between `None` and `""`.

```
Python 3.10.6 (main, Aug  8 2022, 19:28:58) [Clang 13.1.6 (clang-1316.0.21.2.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> type(None)
<class 'NoneType'>
>>> type("")
<class 'str'>
```

containers/phdi-ingestion/app/utils.py
```diff
@@ -77,7 +77,9 @@ def search_for_required_values(input: dict, required_values: list) -> str:
     return message
 
 
-def get_cred_manager(cred_manager: str) -> Any:
+def get_cred_manager(
+    cred_manager: str, location_url: str = None
+) -> BaseCredentialManager:
```
DanPaseltiner, line 82: Thanks for improving the specificity of the return type. Nice!

DanPaseltiner (approved):
This is awesome! Thanks for fixing this. One little nit I think we should address, but I'll admit it is a little pedantic and should not be blocking.

PR #106, opened 2022-11-02
Merge inc loc  derive include search parameters

spencer-kathol:
# Description
This change is part of an effort to support multi-resource tables in the PHDI tabulation framework.  It provides foundational support for augmenting search parameters with `_include` and `_revinclude` directives to ensure that data required for multi-resource joins is available in search results.

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -722,8 +722,12 @@ def _merge_include_query_params_for_location(
       result in the original input parameter if called with a variable
       or the return value.
     """
+
     direction, field_location = reference_location.split(":", 1)
 
+    # Search term is _include for forward searchs, _revinclude for reverse searches.
+    # In addition, we must add an :iterate modifier if the reference is relative to
+    # another included resource
     query_param_name = None
     if direction == "forward":
         query_param_name = "_include" if relates_to_anchor else "_include:iterate"
@@ -737,8 +741,15 @@ def _merge_include_query_params_for_location(
 
     query_param_includes = query_params.get(query_param_name)
 
+    # Handle the case where the search term (_include or _revinclude)
+    # is not specified or is specified as a list.
     if query_param_includes is None:
         query_param_includes = []
+        query_params[query_param_name] = query_param_includes
+    elif isinstance(query_param_includes, str):
+        # Convert query_param_includes from str to list, and make sure
+        # the query_params dict references the new object.
+        query_param_includes = [query_param_includes]
         query_params[query_param_name] = query_param_includes
 
     if field_location not in query_param_includes:
--- a/tests/assets/reference_location_test_schema.yaml
+++ b/tests/assets/reference_location_test_schema.yaml
@@ -61,6 +61,6 @@ tables:
     columns:
       Patient ID:
         fhir_path: Patient.id
-      Case Organizaiton:
+      Case Organization:
         fhir_path: Practitioner.id
         reference_location: [reverse:Composition:subject, forward:Composition:custodian] 
--- a/tests/fhir/tabulation/test_fhir_tabulation.py
+++ b/tests/fhir/tabulation/test_fhir_tabulation.py
@@ -805,7 +805,7 @@ def test_merge_include_query_params_for_location():
     )
 
     assert schema_table["metadata"].get("query_params", {}) == {
-        "_include": ["Observation.subject"]
+        "_revinclude": ["Observation.subject"]
     }
 
     # Test forward reference with existing _include query parameter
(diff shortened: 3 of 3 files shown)


bamader:
So there's a couple things I see going on in this PR.

First, there's yet another schema being introduced for test purposes. While this isn't necessarily bad (in fact, we probably should have multiple schema varieties to test against), I noticed that this schema changes the ordering of things like `metadata`, `resource_type`, etc., creating a new nested structure. If we land this PR, we should change that to reflect the agreed upon, standardized layer structure defined in the `tabulation_schema` asset (which is our currently agreed upon _gold standard_ for what a schema should look like).

Second is a much larger issue. The functionality in this PR is (imo) divorced from what we said we wanted out of tabulation right now. This is for a couple of reasons in my eyes:

1. the function modifies a `query_params` dictionary in place without any indication of where this should be called or when it should be used. As I understand it, the `query_params` object is used to set the search string URL that we'll make as an API call to the FHIR server. The query params tell the server how to collect and collate resources for us, as well as specifying any conditions on the resources we expect to receive back. Because this function is couched as "merging" query params, it implies that some of `_include` and `_revinclude` are already in the query params, which makes me question why we need a merging function to begin with. I think the core idea of building the `include` and `revinclude` parameters into what will eventually generate the search string is a good thing, and maybe this PR can be re-scoped to that, but in its present state I am somewhat confused by the intended use of this function.

2. the function and schema makes some assumptions that don't hold for what we've said about tabulation in its current state. There are example tests included in this PR that indicate the possibility that a `reference_location` for a referenced resource column is a _list_, rather than a single string value. That possibility was to this point never discussed--while I _think_ I understand that this refers to resource chaining (more on that latter though), tabulation as an algorithm right now does not support this kind of behavior. We made the conscious design decision to force a single string parameter so that a user would be deliberate about what reference columns they wanted to specify. Implementing a chaining operation would require a likely substantial refactor of the tabulation work we otherwise just successfully landed. On first glance, though, this PR doesn't make it seem like that's the case, but because there are no tests for how tabulation would change when dealing with chained list reference data, the fact that we're not seeing errors is a result of omission rather than success.

3. most importantly, I think this PR runs countermanding to a decision we made around resource chaining and linking. When we were developing tabulation, I had a conversation with Spencer around one example where chaining might pop up. "Let's say I want a table compiling diagnostics on some patients, where Patient is my anchor resource, I want the DiagnosticReport ID of any reports related to that patient, and I want the InterpretingProvider who created those reports all in a row. Patient will define the rows, DiagnosticReport ID will be a list of reverse references, and InterpretingProvider will be a reference off of Diagnostic Report rather than Patient." Spencer's response illustrated the difficulty of chaining resources like this: "The interpreting provider would not be tied to the ID of the specific diagnostic report, it would just be a big list of report IDs in one column, and a big list of interpreting providers for those reports in the other. To get more usable data in this scenario, I think the better option would be to create another table against DiagnosticReport, and include the subject field as a column in that table that could be joined to a patient table post extraction. The DiagnosticReport table could contain a column with interpreting providers." Now, it's possible Spencer's research uncovered that this `_iterate` parameter could be used to constrain the chained reference so that this didn't happen; I tried to research this yesterday and didn't come to a conclusive stand one way or the other. But the bigger issue is that I was under the impression that to handle nested reference layers, it was actually easier to just ask the user to create two simple tables and join them post-extraction and tabulation. If we want to change that, it's possible tabulation in its current form could be used to address this, but this gets back to the difficulty in handling a list of reference locations outlined in point 2 above.

All of this goes to say two things: I think before we try to move forward with modifying/reviewing/implementing this PR, we need to revisit the design decisions underpinning a) what tabulation is intended to do, and b) what tabulation is _not_ intended to do (i.e. its limitations) to determine if resource chaining is in scope. I think we then need to come to a consensus on a) the types of chaining that we'll accept, and b) how the user will represent that chaining in the schema. Once we address these points, _then_ I think we can move forward with the code here. I think there's just a lot of big ideas in this code that aren't as tested and documented as I'd like given our drive to finish tabulation this sprint, so we need some team input on those thoughts before jumping down the rabbit hole.

bamader:
After discussion with the team, we'll be closing this PR. The concerns discussed above are very real and valid, and for the time being--when we're trying to finish up our tabulate procedures by the end of a sprint with just a few days left--a lot of this work is more complicated and out of scope than we need to make it. Instead, we'll be taking the following approach:

1. open up a new PR with the minimum functionality needed to embed `include` and `revinclude` in FHIR server search strings
2. later, after we've begun working with LAC for a while and have more understanding of the complexity of their use cases, we'll revisit this problem from the "FHIR" side rather than the tabulation side. We'll map out chained reference use cases we've seen in practice, understand how the FHIR server returns those resources, and then implement a multi-tiered approach to tabulation depending on the complexity of the case at hand (i.e. if a user wants a no reference or straightforward include table, just tabulate that ourselves; if they have chained references, make multiple FHIR calls, create some simple intermediary tables, then do the joins ourselves and return the result to the user).

PR #107, opened 2022-11-02
Add write data feature

bamader:
This PR introduces the updated functionality to write tabulated data to one of several output types. Input data is expected to be in a list of lists format (which means that coming out of tabulation, the user will need to call write on each of the tables in the returned dict--this allows them to specify different table names and output file destinations for each table).


phdi/tabulation/tables.py
```diff
@@ -42,6 +43,89 @@ def load_schema(path: pathlib.Path) -> dict:
...
+        # Create requested table if it's not already in the database
+        headers = tuple(tabulated_data[0])
+        cursor.execute(f"CREATE TABLE IF NOT EXISTS {db_tablename} {headers};")
+
+        # Format the programmatic values correctly by number of cols
+        # Need this ugly construction so that the binding placeholder qmarks
+        # don't have quotes around them individually when parsed by the SQL
+        # engine
+        hdr_placeholder = "({})".format(", ".join("?" * len(tabulated_data[0])))
+        tuple_data = [tuple([str(x) for x in row]) for row in tabulated_data[1:]]
```
m-goggins, line 113: Why are we converting to string? Would we not want to preserve the original data types before inserting into a table? 

I imagine this could also trigger an error if the table has been defined prior to using the tabulation BB and the datatypes are not strings.

phdi/tabulation/tables.py
```diff
@@ -42,6 +43,89 @@ def load_schema(path: pathlib.Path) -> dict:
...
+    :param db_tablename: The name of the table in the database to create
+      or write data to. Omit if `output_type` is not SQL. Default: `None`.
+    :param pq_writer: A pre-existing `ParquetWriter` object that can be
+      used to append data to a parquet format. Used in cases where
+      incremental writing to a parquet destination is desired. Omit if
+      `output_type` is not Parquet. Default: `None`.
+    """
+    if output_type == "csv":
+        write_headers = False if os.path.isfile(location + filename) else True
+        with open(location + filename, "a", newline="") as fp:
```
m-goggins, line 85: This is likely outside the scope of this ticket (and possibly the BB), but I'm wondering if you think users would only want to append _new_ data to an existing file. I'm not sure if there's a use case but maybe something like maintaining a list of patients who have gotten their flu shots over time. You wouldn't want to re-add data that's already in the existing file. Maybe this is a question for research but curious to know what you think as well.

tests/tabulation/test_tables.py
```diff
@@ -41,6 +44,175 @@ def test_load_schema():
...
+    file_format = "csv"
+
+    if os.path.isfile(file_location + output_file_name):  # pragma: no cover
+        os.remove(file_location + output_file_name)
+
+    # Batch 1 tests writing and creating brand new file
+    # Only one row actually written in first batch
+    write_data(batch_1, file_location, file_format, filename=output_file_name)
+    with open(file_location + output_file_name, "r") as csv_file:
+        reader = csv.reader(csv_file, dialect="excel")
```
m-goggins, line 74: Do we need `dialect-"excel"`?

tests/tabulation/test_tables.py
```diff
@@ -41,6 +44,175 @@ def test_load_schema():
...
+            for i in range(len(row)):
+                assert row[i] == str(batch_1[line][i])
+            line += 1
+        assert line == 2
+
+    # Batch 2 tests appending to existing csv
+    # Two more rows written here, make sure no duplicate header row
+    write_data(batch_2, file_location, file_format, filename=output_file_name)
+    with open(file_location + output_file_name, "r") as csv_file:
+        reader = csv.reader(csv_file, dialect="excel")
```
m-goggins, line 86: ditto

tests/tabulation/test_tables.py
```diff
@@ -41,6 +44,175 @@ def test_load_schema():
...
+        line = 0
+        for row in reader:
+            print(row)
```
m-goggins, line 89: Remove print statement unless necessary

tests/tabulation/test_tables.py
```diff
@@ -41,6 +44,175 @@ def test_load_schema():
...
+    res = cursor.execute("SELECT * FROM PATIENT").fetchall()
+    assert len(res) == 3
+    assert res == [
+        (
+            "Price929",
+            "Waltham",
+            "['obs1']",
+            "i-am-not-a-robot",
+        ),
+        ("Shepard", "Zakera Ward", "None", "no-srsly-i-am-hoomun"),
```
m-goggins, line 206: lol

tests/tabulation/test_tables.py
```diff
@@ -41,6 +44,175 @@ def test_load_schema():
...
+        ("None", "Faketon", "['obs2', 'obs3']", "None"),
+    ]
+    conn.close()
+
+    os.remove(file_location + db_file)
+
+
+# @TODO: REMOVE THIS FUNCTION ALONG WITH THE OLD GENERATE ALL TABLES CODE
```
m-goggins, line 214: Good call on marking these as needing to be deleted.

m-goggins (changes requested):
Looking pretty good to me. I made a few minor suggestions and commented on several things that are likely outside the scope of this PR but I wanted to get your thinking on them as well.

phdi/tabulation/tables.py
```diff
@@ -42,6 +43,89 @@ def load_schema(path: pathlib.Path) -> dict:
...
+        if pq_writer is None:
+            pq_writer = pq.ParquetWriter(location + filename, table.schema)
+        pq_writer.write_table(table=table)
+        return pq_writer
+
+    if output_type == "sql":
+
+        # We need a file-space cursor to operate on a connected table
+        conn = sql.connect(location + db_file)
+        cursor = conn.cursor()
```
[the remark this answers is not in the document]
bamader: We probably should, yeah. This is a good idea to support in the future, but since that relates more to a SQL feature than an intrinsic part of the writing functionality, I'm going to say that's out of scope for the current PR. I'll write a ticket for it as part of closing this PR.

phdi/tabulation/tables.py
bamader: Yeah, I hear you. Initially the conversion wasn't in this PR at all, and we just wrote the raw data into the table. But then when we merged in tabulation and had values start coming back as lists (i.e. lists of reverse pointers to a an anchor resource), everything broke. This is one of the only things I have to complain about SQL and sqlite3--we can't insert Python lists into tables as objects. We have to instead serialize them in some way, our options of which are stringifying (done here), json.dumps (seemed weird because we're not maintaining a dictionary), or converting to binary (the worst out of the three lol). I _think_ there are ways we can register decorators or automatic converters to serialize for us when writing, but it did seem to be a bit more complex and could involve some architecture conversations around how we want to handle this (so I'll ping @bryan-skylight for his thoughts on serializing). 

My instinct for right now is to leave this as a string (because it works and unblocks the write features), and then open a separate ticket for work around proper serializing/de-serializing data as its being written and read from the database file (because I think it'll involve some research and extra complexity).

phdi/tabulation/tables.py
bamader: I think your mind is in the right place with a pretty clear use case our end users might have, but my current understanding is that that falls outside the scope of the building blocks (or rather, the BB as write presently). Data maintenance and updating existing records are more akin to a style of database management we don't currently support; so, using the example you provided, I would expect a user's workflow to be something like: use BB to write and append incoming data to a record of all patients, write separate program or function to work through this file of existing records and add info where relevant in a new column. If we tried to use the BB (that we have currently) to do this, I think we'd end up backed into doing something really clunky, gross, and inefficient versus just asking the user to write such a function themselves. @bryan-skylight Any more thoughts on whether this is something we should support in the future? It's definitely not in scope for this PR, but it is a good question that we might want to tackle at some point.

tests/tabulation/test_tables.py
bamader: I think we do need this, yeah. A `csv.reader.Dialect` is the object that holds rules for interpreting quotes, double quotes, end of lines, and escape characters, so having this ensures that the results of the tests are consistent across operating systems that generate the test CSV (there are dialect objects for both Windows systems via `excel` and unix systems via, well, `unix`).

tests/tabulation/test_tables.py
bamader: lol likewise ditto

tests/tabulation/test_tables.py
bamader: For my own curiosity, which part of this 100% legitimate test data point earned the lol?

phdi/tabulation/tables.py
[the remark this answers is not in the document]
bamader: Love this. Great call 😎

tests/tabulation/test_tables.py
m-goggins: "no-srsly-i-am-hoomun"

phdi/tabulation/tables.py
m-goggins: Yeah, that makes sense to me for now. Maybe add a TODO item here to remind us to come back to this.

tests/tabulation/test_tables.py
bamader: Ah shoot I was hoping you might've recognized the game reference lol

tests/tabulation/test_tables.py
bamader: Actually yeah we probably should. We expect our users to be operating on government Windows machines, but that extra safety layer is smart.

m-goggins (approved):
Approve knowing that we have future word related to this function

PR #108, opened 2022-11-03
Incorporate drop_invalid into tabulate_data

m-goggins:
Small PR to incorporate the `drop_invalid` function into tabulation and update the tabulation_schema file. Note: `drop_invalid` and `tabulate_data` function are still tested separately.


bryan-skylight (approved):
We don't need to change any tests to reflect this change in the `tabulate_data` function? If not, then this is good to go.

m-goggins:
@bryan-skylight, no. I kept the full run of `drop_invalid` tests separate from the `tabulate_data` tests, although the `tabulate_data` test implicitly tests that invalid values are only dropped when they are present. i.e., Patient ID includes `null` and` ""` as invalid values in the schema but because they are not present in the tabulation test data, they are not dropped.

PR #109, opened 2022-11-03
remove drop_null from tabulation code

m-goggins:
Another small PR to purge `drop_null` from the code base since `drop_invalid` takes care of null values


bryan-skylight (approved):
Code deletion >> code addition

PR #110, opened 2022-11-03
Add entry search feature

bamader:
This PR adds the functionality specified in this ticket https://app.clickup.com/t/3rgqa07 . Previously, it was possible for the FHIR server to return referenced resources that happened to share the same type as a table anchor, and these referenced resources would then incorrectly generate new rows in the table. With the changes in this PR, this no longer happens. Now, only resources that match the _original_ FHIR server query criteria generate rows in the table, while "include" resources are used to fill out desired column values.

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -144,9 +144,16 @@ def tabulate_data(data: List[dict], schema: dict) -> dict:
 
         # Second pass over just the anchor data, since that
         # defines the table's rows
-        for anchor_resource in (
+        for anchor_resource, is_result_because in (
             ref_dicts.get(table_name, {}).get(anchor_type, {}).values()
         ):
+
+            # Resources that aren't matches to the original criteria
+            # don't generate rows because they were included via a
+            # reference
+            if is_result_because != "match":
+                continue
+
             row = []
 
             for _, column_params in column_items:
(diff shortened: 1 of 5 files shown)


tests/assets/FHIR_server_extracted_data.json
```diff
@@ -250,6 +250,9 @@
                     "reference": "doc2",
                     "type": "Practitioner"
                 }
+            },
+            "search": {
+                "mode": "match"
```
m-goggins, line 255: Does this information come back automatically when extracted? Just wondering why it wasn't included before.

phdi/fhir/tabulation/tables.py
```diff
@@ -144,9 +144,16 @@ def tabulate_data(data: List[dict], schema: dict) -> dict:
...
         # defines the table's rows
-        for anchor_resource in (
+        for anchor_resource, is_result_because in (
             ref_dicts.get(table_name, {}).get(anchor_type, {}).values()
         ):
+
+            # Resources that aren't matches to the original criteria
+            # don't generate rows because they were included via a
+            # reference
+            if is_result_because != "match":
```
m-goggins, line 154: If I'm understanding this correctly, if the `is_result_because` is not present in the ref_dicts, i.e., empty dictionary, then this will `continue` as well, right? Is there a chance that we would miss data? I think this is related to my other comment below on whether we expect `search` to always be included in the extracted data.

m-goggins (commented):
I think I understand what this PR is doing but I need more clarification around `search` and `mode` being included in the extracted data.

tests/assets/FHIR_server_extracted_data.json
bamader: Yes! The FHIR server automatically includes this information in all included resources of the requested query type (if it's applicable). The only reason it wasn't included in our sample data before was because a) I completely fabricated this example file, and b) I didn't know this was a thing until Spencer told me lol. So this information will be auto-populated for any anchor resource we query. And, if it's _not_ present in an ancillary resource, that means by definition the resource also isn't a result of the original query of interest, which is also telling information.

phdi/fhir/tabulation/tables.py
bamader: Good question. Based on my pretty lengthy dive into the documentation, no, there is _no_ chance of missing data. `entry.search.mode` will _always_ be populated in a FHIR server results return for _every_ resource related to the original query. That means that if a resource doesn't have an `entry.search` dictionary, it wasn't an original queried resource, hence can't be an anchor, hence can't generate a row.

tests/assets/FHIR_server_extracted_data.json
m-goggins: Oh awesome!

m-goggins (approved):
Looks good to me!

PR #115, opened 2022-11-07
Schema validation

m-goggins:
This PR covers the [ticket](https://app.clickup.com/t/3r5f1eg) for validating a user-specified schema. It ensures that users have appropriately specified their schemas in terms of structure and (some) content, and returns jsonschema Validation errors in cases where the schemas have not been properly specified.


phdi/tabulation/validation_schema.json
```diff
@@ -0,0 +1,86 @@
...
+                        ],
+                        "uniqueItems": true
+                    }
+                },
+                "selection_criteria": {
+                    "type": "string",
+                    "enum": [
+                        "first",
+                        "last",
+                        "random"
```
bryan-skylight, line 71: This should also include "all" for instances where the user wants all values returned (likely to be useful for fields like race and gender)

bryan-skylight (approved):
This is fantastic! Nice work! Once you make the change to include "all" in the selection criteria enum, I think this is ready to go.

PR #117, opened 2022-11-09
Update generate all tables

m-goggins:
This PR creates the `generate_tables` function, which replaces the `generate_table` and `generate_all_tables_in_schema` functions per this [ticket](https://app.clickup.com/t/3pc6e55), and associated tests. To orchestrate all of the tabulation components, the user must now provide an `output_data` dict that contains all of the necessary parameters to execute `write_data` per table. The output_data.json asset I created mocks this up.


phdi/fhir/tabulation/tables.py
```diff
@@ -622,3 +622,42 @@ def _get_reference_directions(schema: dict) -> dict:
...
+    desired location, according to the supplied schema.
+
+    :param schema_path: A path to the location of a schema config file.
+    :param output_data: A dictionary containing the parameters for writing each table.
+    :param fhir_url: A URL to a  FHIR server.
+    :param cred_manager: The credential manager used to authenticate to the FHIR server.
+    """
+
+    # Load schema
+    schema = load_schema(schema_path)
```
BradySkylight, line 644: Do we need any kind of check here to verify that the schema is not null/none after it returns from this function?

phdi/fhir/tabulation/tables.py
```diff
@@ -622,3 +622,42 @@ def _get_reference_directions(schema: dict) -> dict:
...
+
+    # Extract data from FHIR server
+    extracted_data = extract_data_from_schema(schema, fhir_url, cred_manager)
```
BradySkylight, line 647: Would it make sense to wrap this in a try-catch or is there error handling within this function itself?  ie... the fhir refuses connection, or the URL is bad, or there is no data returned?

phdi/fhir/tabulation/tables.py
m-goggins: We added the `validate_schema` function to `load_schema`, which should catch errors in the schema so that we don't return an invalid schema. Here's the [code](https://github.com/CDCgov/phdi/blob/main/phdi/tabulation/tables.py#L48-L64) for `validate_schema` in case you want to take a look/have suggestions.

BradySkylight: I thought as much.  I was just being too lazy to dive into that code in the load_schema!

phdi/fhir/tabulation/tables.py
m-goggins: I think we could probably add more error handling within the function itself (or in one of the several functions that are nested within this function). 

For the URL, it looks like `_generate_search_url` raises a value error if the search url is not correctly specified in the schema. 

For the data, the function is nested like: `extract_data_from_fhir_search` <- `extract_data_from_fhir_search_incremental` <- `http_request_with_reauth` <- http_request_with_retry`. If a token is old, the retry is triggered. I actually don't know what will happen if fhir refuses the connection - any idea?

We could catch the no data returned in `extract_data_from_fhir_search_incremental`, which I think is a good idea since it doesn't make sense to move on to tabulation without any data. I can create a separate ticket for that work item. I'm curious if @bamader has any input on this.

BradySkylight: This is a good discussion, for clarity, if I understand we have a few different checks that need to be made in order.  Some are already handle and others may not be.

BradySkylight: This is a good discussion, for clarity, if I understand we have a few different checks that need to be made in order. Some are already handle and others may not be.

1. Validate the URL - This already exists
2. FHIR Server refuses connection -> I'm pretty sure the FHIR server would return some kind of 400 error, though it may depend upon the implementation of the specific FHIR server.  I think if you have a connection to the FHIR server using the token along with the refresh option you specified, we should be good, as long as there is a check initially to ensure a successful connection with the FHIR server was successful.  I assume this is already in place.
3. No data returned during the extraction incremental process -> we can discuss this in more detail with Brandon and yourself, if we aren't already handling this.

phdi/fhir/tabulation/tables.py
```diff
@@ -625,3 +627,42 @@ def _get_reference_directions(schema: dict) -> dict:
                     ] = ref_path
 
     return directions_by_table
+
+
+def generate_tables(
+    schema_path: pathlib.Path,
+    output_data: dict,
```
bamader, line 634: nit: could we change this variable name? Since it's a collection of parameter options I don't think it should be called `_data`. I'd opt for something simple like `output_params`.

phdi/fhir/tabulation/tables.py
```diff
@@ -625,3 +627,42 @@ def _get_reference_directions(schema: dict) -> dict:
...
+) -> None:
+    """
+    Queries a FHIR server for information, and generates and stores the tables in the
+    desired location, according to the supplied schema.
+
+    :param schema_path: A path to the location of a schema config file.
+    :param output_data: A dictionary containing the parameters for writing each table.
+    :param fhir_url: A URL to a  FHIR server.
+    :param cred_manager: The credential manager used to authenticate to the FHIR server.
+    """
```
bamader, line 646: Could you add some more information in the docstring about the `output_params` dictionary? I think we should explain to the user what the broad format of that is (i.e. there's a dict inside for each table in the schema, and each dict has the parameters of the desired `write` call the user intends to make).

phdi/fhir/tabulation/tables.py
bamader: Good points here, all. I think we should be pretty robust in how we handle any communication from the FHIR server. Request with retry and with reauth should both (I think already, but if not we definitely need to implement this) raise appropriate errors if communication with the server doesn't work. I think we may want to raise a warning in `extract_incremental` if we get something with no data back, because there's definitely the possibility that one part of an incremental search returns no data while the other incremental searches did (in this case, I say warning just as a way to not break the process; if incremental search X has a warning for no data, we should continue processing X+1 and all the iterations between 1 and X if they have data). Then, in `extract_from_fhir_search`, I think once all the incrementals complete, if there's no data retrieved whatsoever, we throw an error and don't tabulate. So I think we'd have multiple layers of catching to figure things out depending on how severe the issue and how high up the calling chain.

phdi/fhir/tabulation/tables.py
```diff
@@ -625,3 +627,42 @@ def _get_reference_directions(schema: dict) -> dict:
...
+        # Write each table
+        write_data(
+            tabulated_data=tabulated_data,
+            directory=output_data[table_name].get("directory"),
+            filename=output_data[table_name].get("filename"),
+            output_type=output_data[table_name].get("output_type"),
+            db_file=output_data[table_name].get("db_file", None),
+            db_tablename=output_data[table_name].get("db_filename", None),
+            pq_writer=output_data[table_name].get("pq_writer", None),
+        )
```
bamader, line 668: So clean 😎

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -691,3 +694,53 @@ def test_extract_data_from_schema(patch_search, patch_gen_urls):
...
+    patients_path = os.path.join(
+        output_data["Patients"]["directory"], output_data["Patients"]["filename"]
+    )
+    assert os.path.exists(patients_path) is True
+
+    physical_exams_path = os.path.join(
+        output_data["Physical Exams"]["directory"],
+        output_data["Physical Exams"]["filename"],
+    )
+    assert os.path.exists(physical_exams_path) is True
```
bamader, line 746: Instead of just checking that the files exist (which we should do for sure), can we also open them and check that the information inside matches what we expect to be there? Because looking at the test_outputs files attached below, they actually look wrong to me (the same set of three rows of content gets repeated a bunch of times which shouldn't be happening anymore). Also, I would recommend we use the same code that we have in other tests to delete the output files after they're created and checked; I think we do it in the `test_write` functions, but that way we're not leaving behind any kruft or new files on a user's machine just by running the tests.

bamader (changes requested):
Good start to this work. I think we need some changes around testing because the outputs don't look right to me. Also a couple suggestions for docstring information.

phdi/fhir/tabulation/tables.py
m-goggins: This sounds good to me. I added a ticket to address these issues and will work on it later today.

BradySkylight: This sounds like a good and robust approach.  I like it!!

tests/fhir/tabulation/test_fhir_tabulation.py
m-goggins: I think the repeated data came from me testing the code multiple times and our `write_data` function blindly appending without checking existing data. This should be resolved with your suggestion to delete the file after testing. I've added checks to make sure the content of new test files matches the expected output.

bamader (approved):
This looks much better. Great work on this Marcelle!

PR #119, opened 2022-11-10
Fix extract data from fhir search incremental

m-goggins:
Quick PR to address an issue that popped up in testing the `generate_tables` function. We use `http_response_with_reauth` in `extract_data_from_fhir_search` which returns a response, e.g., 200. We had mocked the response as actual data from the FHIR server, which is what we would like to have. I have updated the function `extract_data_from_fhir_search` and associated tests to reflect that we now return data from the FHIR server and not just the response.


BradySkylight (approved):
This looks good to me.  Good find and I like the naming conventions used.

bamader (approved):
Seems straightforward and legit

BradySkylight (approved):
Nice!  Looks good.

BradySkylight (approved):

PR #122, opened 2022-11-14
Update drop_invalid signature

m-goggins:
The  `drop_invalid`  function currently expects the data it's given to be in the form of a dictionary of tables, where each table is one of the outputs of the tabulate_data  function. However, due to how tabulate needs to process data to be compatible with our extraction procedure, this signature needs to change. During extraction, we process results in one list of dicts _per table_, which makes much more sense given that's how the FHIR server sends back content responses to its API calls. We then tabulate the results one table at a time. The scope of this work is therefore updating the `drop_invalid`  function so that it operates on only a single table's worth of information, the result of a single call to `tabulate_data` .


phdi/fhir/tabulation/tables.py
```diff
@@ -10,43 +10,42 @@
...
-      tabulated data, including invalid_values if applicable.
-    :param return: A dictionary mapping table names to lists of lists, without resources
-        that contained invalid values. The first list in the data value is a list of
-        headers serving as the columns, and all subsequent lists are rows in the table.
+    :param table_name: Name of the table to drop invalid values.
+    :param schema: A schema of columns and values to apply to the tabulated data,
+        including invalid_values if applicable.
+    :param return: A list of lists, without resources that contained invalid values. The
+        first list in the data value is a list of headers serving as the columns, and
+        all subsequent lists are rows in the table.
```
BradySkylight, line 27: This description is great!  When you say a list of lists without resources, would it be safe to say (just for my own understanding) that it's a list of list with specific rows of data missing from the list due to the invalid values?

m-goggins: Yes, exactly!

phdi/fhir/tabulation/tables.py
```diff
@@ -10,43 +10,42 @@
...
-                        data[table].remove(resource)
-                        break
+    # Identify columns to drop invalid values for each table in schema
+    columns = schema["tables"][table_name]["columns"]
+    # Identify indices in List of Lists to check for invalid values
+    invalid_values_by_column_index[table_name] = {
+        i: columns[col].get("invalid_values")
+        for i, col in enumerate(columns)
+        if columns[col].get("invalid_values", [])
+    }
```
BradySkylight, line 38: I'm curious to understand more of how this works in python as it appears you are creating a list of dicts, unless I'm misunderstanding.  More so a dict of dicts maybe and using the table_name as the key for the value of a dict of invalid_values.  Correct?

phdi/fhir/tabulation/tables.py
```diff
@@ -10,43 +10,42 @@
...
 
+    # Check if resource contains invalid values to be dropped
+    if len(invalid_values_by_column_index) > 0:
+        for resource in data[1:]:
+            for index, invalid_values in invalid_values_by_column_index[
+                table_name
+            ].items():
+                if resource[index] in invalid_values:
+                    data.remove(resource)
+                    break
```
BradySkylight, line 48: Again, my assumption here is that 'resource' is 'row of data from the list'.  Correct?  Thanks for your patience with me.  :)

phdi/fhir/tabulation/tables.py
BradySkylight: This may be too nit-picky, and please push back, but would it make more sense to describe it as a 'row' or even 'resource_row'?

phdi/fhir/tabulation/tables.py
```diff
@@ -375,7 +374,7 @@ def _dereference_included_resource(
     # data didn't contain any of them, the referenced type
     # doesn't appear in the mapping
     if referenced_type not in ref_dicts[table_name]:
-        return None
+        return None  # pragma: no cover
```
BradySkylight, line 377: This just ensures that the line of code is not evaluated for test coverage, correct?

phdi/fhir/tabulation/tables.py
m-goggins: You're definitely on the right track; I will say this code is perhaps fairly pythonic (i.e., not a lot of lines) but it could be more readable with more code. It's making a dict of dicts, where the table_name is the top level key, and the second level of keys are the indices of the list where we should be checking for invalid values. The end result looks like this:
`invalid_values_by_column_index = {'Patients': {0: [None, ""], 1: [None, "", 'Unknown']}}`

So, in the `Patients` table when we're looking at data, we should check at index= 0 for `None` and empty strings whereas at index=1 we should check for `None`, empty strings, and `'Unknown'`.

We could simplify to exclude the table_name level of the dictionary to be something like the following if you think it's easier to understand:
`invalid_values_by_column_index = {0: [None, ""], 1: [None, "", 'Unknown']}`

Originally, we were dropping invalid values for multiple tables, hence the nesting. The only upside I see for keeping the `table_name` is that it could be easier for debugging if there was some sort of issue with the specified invalid values.

Let me know if that makes sense.

phdi/fhir/tabulation/tables.py
m-goggins: Yes, precisely. If you think it makes more sense to change the naming I am not particularly attached to `resource`.

phdi/fhir/tabulation/tables.py
m-goggins: Yep, exactly. It wasn't exactly related to this PR...but I made the change

phdi/fhir/tabulation/tables.py
m-goggins: Haha I will leave the FHIR-related nomenclature up to you! Do you think it is helpful to know that the data we're referring to is the resource data and not some row in the resource? 

We have a separate [ticket](https://app.clickup.com/t/3qve7j5) related to cleaning up the docstrings. I think this issue would be appropriate to tackle in that round of work rather than this PR since this return is also the same as the `tabulate_data` return and I want to make sure we are consistent everywhere.

BradySkylight (approved):
Thank you for humoring my questions!  This is great!

phdi/fhir/tabulation/tables.py
```diff
@@ -10,43 +10,42 @@
...
+    }
 
+    # Check if resource contains invalid values to be dropped
+    if len(invalid_values_by_column_index) > 0:
+        for resource in data[1:]:
+            for index, invalid_values in invalid_values_by_column_index[
+                table_name
+            ].items():
+                if resource[index] in invalid_values:
+                    data.remove(resource)
```
bamader, line 47: Are we sure this is iterator-safe? I'm a bit surprised this doesn't throw an error. Since the iteration is over the rows in the table, and we outright remove rows in the loop, I would've expected some sort of concurrency issue with modifying the iterated loop in place.

Also, for efficiency reasons, I might suggest the following approach: iterate over `i in range(len(data[1:]))` rather than over the rows, create a list of the row numbers that need to be dropped rather than dropping things mid-loop, and then after that first loop, loop over all the row numbers in that list of rows-to-drop and drop those indices (in reverse order). As written, this is actually an n^2 operation since `.remove()` has to loop through the data list to find the value that matches the given row, whereas if we make two loop overs, it's linear time since it's just 2n.

bamader (changes requested):
Mostly there, just some change suggestions around naming conventions and a loop safety / efficiency thing.

phdi/fhir/tabulation/tables.py
```diff
@@ -10,43 +10,42 @@
...
+    # Identify indices in List of Lists to check for invalid values
+    invalid_values_by_column_index[table_name] = {
+        i: columns[col].get("invalid_values")
+        for i, col in enumerate(columns)
+        if columns[col].get("invalid_values", [])
+    }
 
+    # Check if resource contains invalid values to be dropped
+    if len(invalid_values_by_column_index) > 0:
+        for resource in data[1:]:
```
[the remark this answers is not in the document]
BradySkylight: I was thinking the same thing about this.  I think row makes it more clear.

phdi/fhir/tabulation/tables.py
BradySkylight: This is a good catch.

phdi/fhir/tabulation/tables.py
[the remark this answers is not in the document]
m-goggins: Ha Brady called this out as well. I'll change this to row.

phdi/fhir/tabulation/tables.py
m-goggins: Good point! I made the changes you suggested except I looped over ` i in range(len(data))` so that the indices would match and I didn't have to do any additional manipulation when running through the data a second time.

phdi/fhir/tabulation/tables.py
```diff
@@ -10,42 +10,47 @@
...
+            for index, invalid_values in invalid_values_by_column_index[
+                table_name
+            ].items():
+                if data[i][index] in invalid_values:
+                    rows_to_remove.append(i)
+                    break
+
+    # Remove rows with invalid values
+    for idx, i in enumerate(rows_to_remove):
+        del data[i - idx]
```
bamader, line 53: Oh nice 😎

bamader (approved):
Nice work, thanks for updating those things!

PR #126, opened 2022-11-16
Add warning and error for empty FHIR response content during extraction

m-goggins:
This PR addresses this [ticket](https://app.clickup.com/t/3v6p7ta). When testing `generate_tables` we realized that there was no check for whether data was actually being returned from the FHIR server, i.e., returning empty data. I added two checks: 

1. In `extract_data_from_fhir_search_incremental` the user will see a warning if the particular search_url returns no data, but the code will continue to run incrementally on the off chance that there is other data that can be retrieved from the FHIR server. The warning includes the search_url to help the user debug.
2. In `extract_data_from_fhir_search`, an error will be raised if there is no data returned at all for a given table. This will stop the tabulation process for all tables but will prevent errors popping up in the `tabulate_data` function.


phdi/fhir/tabulation/tables.py
```diff
@@ -77,6 +79,12 @@ def extract_data_from_fhir_search(
         )
         results.extend(incremental_results)
 
+    # Check that results are not empty
+    if not results:
+        raise KeyError(
+            f"No data returned from server with the following query: {search_url}"
+        )
+
```
BradySkylight, line 87: This is great!

BradySkylight (approved):
This looks good and is an elegant solution to handle no data being returned from the FHIR Server.  Nice!

m-goggins:
Realized that a ValueError would be more appropriate.

PR #127, opened 2022-11-17
Add data-types to validation schema

nickbristow:
Support data types in validation schema. 

Limits the supported data types to `bool`, `float`, and `string`


tests/assets/valid_schema.yaml
```diff
@@ -18,9 +18,11 @@ tables:
           - ""
           - "Unknown"
         selection_criteria: first
+        data_type: string
       Last Name:
         fhir_path: Patient.name.family
         selection_criteria: first
+        data_type: string
```
m-goggins, line 25: Make sure to add these to the `valid_schema.json` file as well...and the `tabulation_schema.yaml` file. We have too many test files but just want to keep them consistent for now before paring down.

tests/assets/valid_schema.yaml
```diff
@@ -29,6 +31,9 @@ tables:
           - "Unknown"
           - "DNA"
         selection_criteria: first
+      Age:
+        fhir_path: Patient.age
+        data_type: number
   table 2A:
```
m-goggins, line 37: If `data_type` is required (from comment above), we should add data types for this table as well.

m-goggins (changes requested):
Looks pretty good to me! I just requested a few changes (to ensure that the `valid_schema.json` and `valid_schema.yaml` files are consistent and had a question about how `data_type` will be implemented.

phdi/tabulation/validation_schema.json
```diff
@@ -71,6 +71,14 @@
                         "random",
                         "all"
                     ]
+                },
+                "data_type": {
```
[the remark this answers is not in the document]
bamader: I think the goal here is to make it not required, and to assume the data should just be treated as a string if no data is specified. At least, that's the assumption I'd make for the `write_data` component of this (which this PR isn't supporting quite yet). Interested in thoughts on that though!

tests/assets/valid_schema.yaml
bamader: _biiiiiiig_ ➕  here for this

tests/assets/valid_schema.yaml
bamader: Also, regardless of if it's required or not, we should add some test cases to show what happens for fields that _don't_ specify a data_type. If it's not required, there should be no change or impact to the functionality, but if it is required, we want to see that throw the appropriate error.

bamader (changes requested):

phdi/tabulation/validation_schema.json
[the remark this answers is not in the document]
m-goggins: I think optional makes sense; just need to add the appropriate error for when a user does not specify the data type and is writing to an already-created SQL table that has other data types so that they know where to add data types to the schema.

tests/assets/valid_schema.yaml
BradySkylight: ++

nickbristow:
@bamader @m-goggins i think this is good to go now. I changed some of the tests and updated them. I also updated the click up to note the need on the other ticket for additional tests.

tests/fhir/tabulation/test_fhir_tabulation.py
```diff
@@ -104,9 +105,10 @@ def test_tabulate_data():
             "Price929",
             "555-690-3898",
             "907844f6-7c99-eabc-f68e-d92189729a55",
+            "165",
         },
-        {"65489-asdf5-6d8w2-zz5g8", "John", "Shepard", None},
-        {"some-uuid", "John ", None, "123-456-7890"},
+        {"65489-asdf5-6d8w2-zz5g8", "John", "Shepard", None, "1234"},
+        {"some-uuid", "John ", None, "123-456-7890", "123"},
```
m-goggins, line 111: Double checking that this is supposed to be a string and not a number as specified in the schema.

m-goggins, line 111: Just for my own clarification...when we update the functions (and not just the schema), these BuildingNumbers will be number type, right?

bamader (approved):
LGTM!

PR #130, opened 2022-11-21
Refactor generate_tables for incremental approach

m-goggins:
This PR refactors the `generate_tables` function to work incrementally, as outlined in the general approach of this [ticket](https://app.clickup.com/t/3pc6e55). Instead of extracting, tabulating, and writing each table into memory, we are extracting, tabualting, and writing each _incremental_ piece of each table so that we do not overwhelm memory in the case of a large table.

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -568,11 +568,8 @@ def _generate_search_urls(schema: dict) -> dict:
                 query_params = _merge_include_query_params_for_location(
                     query_params, column.get("reference_location", "")
                 )
-        print("query_params:", query_params)
         search_string = resource_type
         if query_params is not None and len(query_params) > 0:
-            s = urlencode(query_params, encoding="utf-8")
-            print(s)
             search_string += f"?{urlencode(query_params,True)}"
 
         count = table.get("results_per_page", count_top)
@@ -656,7 +653,7 @@ def generate_tables(
     desired location, according to the supplied schema.
 
     :param schema_path: A path to the location of a schema config file.
-    :param output_data: A dictionary of dictionaries containing the parameters for
+    :param output_params: A dictionary of dictionaries containing the parameters for
         writing each table specified in the schema. For each table in the schema, the
         nested dictionary must contain a directory, filename, and output_type at
         minimum. See `write_data` function for full writing specifications.
@@ -671,23 +668,17 @@ def generate_tables(
     search_urls = _generate_search_urls(schema=schema)
 
     for table_name, search_url in search_urls.items():
-        count = 0
-        print(table_name)
-        print("retrieving 1st incremental results")
         # Get 1st set of incremental results
         incremental_results, next = extract_data_from_fhir_search_incremental(
             search_url=urllib.parse.urljoin(fhir_url, search_url),
             cred_manager=cred_manager,
         )
-        print(urllib.parse.urljoin(fhir_url, search_url))
         # Tabulate data for 1st  set of incremental results
-        print("tabulating for 1st set of incremental results")
         tabulated_incremental_data = tabulate_data(
             incremental_results, schema, table_name
         )
 
         # Write 1st set of tabulated incremental data
-        print("writing data for 1st set of incremental results")
         write_data(
             tabulated_data=tabulated_incremental_data,
             directory=output_params[table_name].get("directory"),
@@ -699,20 +690,16 @@ def generate_tables(
         )
 
         while next is not None:
-            print("retrieving additional results")
             incremental_results, next = extract_data_from_fhir_search_incremental(
                 search_url=urllib.parse.urljoin(fhir_url, next),
                 cred_manager=cred_manager,
             )
-            print("next:", next)
-            print("tabulating additional results")
             # Tabulate data for each set of incremental results
             tabulated_incremental_data = tabulate_data(
                 incremental_results, schema, table_name
             )
 
             # Write each set of tabulated incremental data
-            print("writing additional results")
             write_data(
                 tabulated_data=tabulated_incremental_data,
                 directory=output_params[table_name].get("directory"),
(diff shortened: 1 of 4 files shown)


phdi/fhir/tabulation/tables.py
```diff
@@ -508,6 +509,15 @@ def _merge_include_query_params_for_location(
 
     direction, field_location = reference_location.split(":", 1)
 
+    # Convert field location to search parameter formatting
+    new_field_location = str()
```
bamader, line 513: I would just say `new_field_location = ""`

phdi/fhir/tabulation/tables.py
```diff
@@ -508,6 +509,15 @@ def _merge_include_query_params_for_location(
...
+    for letter in field_location.split(":")[1]:
+        if letter.isupper():
+            letter = "-" + letter.lower()
+        new_field_location += letter
+
```
bamader, line 518: We shouldn't modify an iterator in place. We could instead just append directly to `new_field_location`, e.g.

```
if letter.isupper():
    new_field_location += "-"
new_field_location += letter.lower()
```

This won't do anything to letters that are already lowercase but will successfully prepend hypens without risking an iterator issue.

phdi/fhir/tabulation/tables.py
```diff
@@ -508,6 +509,15 @@ def _merge_include_query_params_for_location(
 
     direction, field_location = reference_location.split(":", 1)
```
bamader, line 511: Since we do multiple splits on `field_location` later, let's just get the constituent parts here:

```
direction, field_location = reference_location.split(":", 1)
resource, element_path = field_location.split(":", 1)
```

Then below, we just make the for loop over the letters in `element_path`, and then recombine `field_location` as `field_location = resource + ":" + new_field_location`

phdi/fhir/tabulation/tables.py
```diff
@@ -572,10 +582,9 @@ def _generate_search_urls(schema: dict) -> dict:
                 query_params = _merge_include_query_params_for_location(
                     query_params, column.get("reference_location", "")
                 )
-
         search_string = resource_type
         if query_params is not None and len(query_params) > 0:
-            search_string += f"?{urlencode(query_params)}"
+            search_string += f"?{urlencode(query_params,True)}"
```
bamader, line 587: God this was so annoying to catch 😅

phdi/fhir/tabulation/tables.py
```diff
@@ -669,21 +678,48 @@ def generate_tables(
...
         write_data(
-            tabulated_data=tabulated_data,
+            tabulated_data=tabulated_incremental_data,
             directory=output_params[table_name].get("directory"),
             filename=output_params[table_name].get("filename"),
             output_type=output_params[table_name].get("output_type"),
             db_file=output_params[table_name].get("db_file", None),
             db_tablename=output_params[table_name].get("db_filename", None),
             pq_writer=output_params[table_name].get("pq_writer", None),
         )
```
bamader, line 704: We can actually condense this code to ditch the second set of repeated calls by preemptively setting next and kicking off the while loop:

```
for table_name, search_url in search_urls.items():
    next = search_url
    while next is not None:
        incremental_results, next = extract_data_from_fhir_search_incremental(
                search_url=urllib.parse.urljoin(fhir_url, next),
                cred_manager=cred_manager,
        )
        # Tabulate data for each set of incremental results
        [code just continues as normal]
```

tests/assets/tabulation_schema.yaml
```diff
@@ -1,5 +1,5 @@
 metadata:
-  incremental_query_count: 1000
+  results_per_page: 1000
```
bamader, line 2: One of these days we'll condense our asset files and actually standardize everything 👀

bamader (changes requested):
Great work, there's just a couple places we could condense code and be a bit more efficient.

phdi/fhir/tabulation/tables.py
BradySkylight: I like this.

tests/assets/valid_schema.yaml
```diff
@@ -32,7 +32,7 @@ tables:
   table 2A:
     resource_type: Observation
     query_params:
-      category: http://hl7.org/fhir/ValueSet/observation-category|laboratory
+      category: laboratory
```
BradySkylight, line 35: Do we know why Spencer was using the full URL value for this here?

bamader: Marcelle and I discussed this yesterday and we came to the conclusion that it was a page that enumerated the possible allowed values for an observation's category element. We then tested it with an actual FHIR server and the full URL didn't resolve to a successful query but truncating it to just `laboratory` did.

phdi/fhir/tabulation/tables.py
m-goggins: Yeah, good call.

phdi/fhir/tabulation/tables.py
m-goggins: Ha of course! My brain did not compute this but it is definitely better.

tests/assets/valid_schema.yaml
m-goggins: Yeah, I'm thinking that he would have changed it if it had been tested using the FHIR server and not mocked responses. As far as I can tell from the [documentation](https://build.fhir.org/searchparameter-registry.html), category should be a token type.

phdi/fhir/tabulation/tables.py
BradySkylight: Was this the solution to the issue where the url was not formatted correctly with a [''] scenario?

tests/assets/valid_schema.yaml
BradySkylight: Makes sense.

phdi/fhir/tabulation/tables.py
m-goggins: Yes haha

BradySkylight (approved):
With the additional changes, this looks good.  Nice work!

bamader (approved):
Thanks for making those changes, looks good!

PR #132, opened 2022-11-30
Fix bugs in gcpGloudStorageConnection

DanPaseltiner:
This PR addresses two problems that were discovered in `gcpGloudStorageConnection`. 

1. The `__storage_client` attribute is now initialized before we attempt to use it.
2. The `upload_object` method now supports both strings and dicts.

Tests have been updated as well to account for these changes.

### DIFF
--- a/containers/phdi-ingestion/requirements.txt
+++ b/containers/phdi-ingestion/requirements.txt
@@ -1,4 +1,4 @@
 fastapi
 uvicorn
-phdi @ git+https://github.com/CDCgov/phdi.git@dan/fix-GcpCloudStorageConnection
+phdi
 httpx
--- a/containers/phdi-ingestion/requirements.txt
+++ b/containers/phdi-ingestion/requirements.txt
@@ -1,4 +1,4 @@
 fastapi
 uvicorn
-phdi
+phdi @ git+https://github.com/CDCgov/phdi.git@dan/fix-GcpCloudStorageConnection
 httpx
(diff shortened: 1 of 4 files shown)


containers/phdi-ingestion/requirements.txt
```diff
@@ -1,4 +1,4 @@
 fastapi
 uvicorn
-phdi
+phdi @ git+https://github.com/CDCgov/phdi.git@dan/fix-GcpCloudStorageConnection
```
BradySkylight, line 3: Eventually this will have to be changed to just phdi, once these changes have been released to pypi.  Nice!

nickclyde, line 3: It might make sense to update our testing workflow to automatically insert the current branch here in an ephemeral workflow run instead of having to commit this change and then change it back later. I'd be happy to help figure that out if you agree

nickclyde (approved):
Looks good to me, besides the note about workflow dependency targeting.

containers/phdi-ingestion/requirements.txt
DanPaseltiner: Thanks for reminding me! Actually this should get changed back before merging. I did this so I could build a new version of the ingestion container to test in GCP.

bamader (approved):
Once the repo name is changed, this looks good--great catches again, Dan!

BradySkylight (approved):
Looks good.  Nice work!

PR #134, opened 2022-12-02
Support list data type in data tabulation

nickbristow:
Supports list types in data tabulation

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -196,18 +196,20 @@ def tabulate_data(data: List[dict], schema: dict, table_name: str) -> List[list]
       The first list is a list of headers serving as the columns,
       and all subsequent lists are rows in the table.
     """
-
+    print('tabulate data \n')
     if table_name not in schema.get("tables", {}):
         raise KeyError(f"Provided table name {table_name} not found in schema")
 
     # First pass: build mapping of references for easy lookup
     ref_directions = _get_reference_directions(schema)
+    data_types = _get_data_types(schema)
     ref_dicts = _build_reference_dicts(data, ref_directions)
 
     # Get the columns from the schema so we always iterate through
     # them in a consistent order
     table_params = schema["tables"][table_name]
     column_items = table_params["columns"].items()
+    print('column_items'+str(column_items))
     headers = [column_name for column_name, _ in column_items]
     tabulated_data = [headers]
     anchor_type = schema["tables"][table_name]["resource_type"]
@@ -330,6 +332,7 @@ def _build_reference_dicts(data: List[dict], directions_by_table: dict) -> dict:
     # Build up connections table by table, since one resource could be
     # used in multiple different tables
     reference_dicts = {}
+    print("*********tableName****"+str(directions_by_table))
     for table_name in directions_by_table.keys():
         reference_dicts[table_name] = {}
 
@@ -447,6 +450,7 @@ def _extract_value_with_resource_path(
     """
     parse_function = _get_fhirpathpy_parser(path)
     value = parse_function(resource)
+    print("resource"+str(resource))
     if len(value) == 0:
         return None
     else:
@@ -631,6 +635,44 @@ def _get_reference_directions(schema: dict) -> dict:
       how referenced resources relate to the anchor resource.
     """
 
+    directions_by_table = {}
+    for table_name, table_params in schema.get("tables", {}).items():
+        anchor_type = table_params.get("resource_type", "")
+        directions_by_table[table_name] = {
+            "anchor": anchor_type,
+            "forward": set(),
+            "reverse": {},
+        }
+
+        for column_params in table_params.get("columns", {}).values():
+            if "reference_location" in column_params:
+                [direction, ref_path] = column_params.get(
+                    "reference_location", ""
+                ).split(":", 1)
+                referenced_resource_type = column_params.get("fhir_path", "").split(
+                    "."
+                )[0]
+                if direction == "forward":
+                    directions_by_table[table_name][direction].add(
+                        referenced_resource_type
+                    )
+                else:
+                    directions_by_table[table_name][direction][
+                        referenced_resource_type
+                    ] = ref_path
+
+    return directions_by_table
+
+
+def _get_data_types(schema: dict) -> dict:
+    """
+    Creates a dictionary mapping indicating the data types that will be used
+    for the data.
+    :param schema: A user-defined schema, for one or more tables, that
+        maps a FHIR resource and element to a specified column in a table.
+    :return: A dict containing mappings, for each table, of
+      how referenced resources relate to the anchor resource.
+    """
     directions_by_table = {}
     for table_name, table_params in schema.get("tables", {}).items():
         anchor_type = table_params.get("resource_type", "")
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -461,8 +461,10 @@ def _convert_list_to_string(val: list) -> str:
     for i, v in enumerate(val):
         if isinstance(v, list):
             val[i] = _convert_list_to_string(v)
-        if isinstance(v, dict):
+        elif isinstance(v, dict):
             val[i] = _convert_dict_to_string(v)
+        elif type(v) != str:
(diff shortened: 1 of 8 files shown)


phdi/fhir/tabulation/tables.py
```diff
@@ -451,9 +450,28 @@ def _extract_value_with_resource_path(
...
+            return _convert_list_to_string(value)
+        if isinstance(value, dict):
+            return _convert_dict_to_string(value)
         return value
 
 
+def _convert_list_to_string(val: list) -> str:
+    for i, v in enumerate(val):
+        if isinstance(v, list):
+            val[i] = _convert_list_to_string(v)
```
bryan-skylight, line 463: This bring up an edge case that I hadn't considered before regarding nested arrays. Because this function unnests the lists, do we risk losing context around certain data elements that have been previously "grouped" together? This isn't a blocking question, but we should make sure to ask Brady once he's back online.

bryan-skylight (changes requested):
The code itself looks good, but the codecov bot is not pleased. Can you please update the tests to ensure we have proper coverage?

bryan-skylight:
> The code itself looks good, but the codecov bot is not pleased. Can you please update the tests to ensure we have proper coverage?

If you look [here](https://app.codecov.io/gh/CDCgov/phdi/pull/134) you'll see the three lines of code where we don't currently have coverage.

phdi/fhir/tabulation/tables.py
nickbristow: Yeah, i wanted to check with @bamader but he's been busy

nickbristow:
> > The code itself looks good, but the codecov bot is not pleased. Can you please update the tests to ensure we have proper coverage?
> 
> If you look [here](https://app.codecov.io/gh/CDCgov/phdi/pull/134) you'll see the three lines of code where we don't currently have coverage.

I've added additional tests to get this to pass

phdi/fhir/tabulation/tables.py
```diff
@@ -266,13 +265,13 @@ def tabulate_data(data: List[dict], schema: dict, table_name: str) -> List[list]
                     )
                     for r in resource_to_use
                 ]
+                values = _convert_list_to_string(values)
```
bamader, line 268: I have a pretty strong preference to not do string conversion during the value extraction part of tabulation and instead keep type conversion strictly to writing. Bryan asked in a different comment whether we'll lose any blocking/contextual information by un-nesting part of the data, and I believe the answer to this is yes. More generally, from a pedagogical perspective, we made the conscious choice some time ago that we wanted to keep each phase of tabulation as "pure" as possible from the perspective of modifying data. I.e. we don't want to unnecessarily convert data into a different type early in the process, because a) that step doesn't _need_ the conversion, and b) users already deal with problems about upstream modifications to their data so we don't want to impose more of that. Third and finally, making this change to convert data at the tabulation level breaks workflows that don't terminate in a write. If a user were working in an interactive session, for example (using the BB in a notebook), then they're no longer able to do analysis on a tabulated result involving lists because they're now strings. I think we should _only_ be changing this type in the write function.

phdi/fhir/tabulation/tables.py
```diff
@@ -451,9 +450,28 @@ def _extract_value_with_resource_path(
         return None
     else:
         value = _apply_selection_criteria(value, selection_criteria)
+        if isinstance(value, list):
+            return _convert_list_to_string(value)
+        if isinstance(value, dict):
+            return _convert_dict_to_string(value)
```
bamader, line 456: I don't actually know if we can encounter any dictionaries at the point of value extraction? Lists, yes, because reverse pointers can accumulate multiple references per anchor, but I don't actually know under what conditions you'd get a list post-value extraction. Did you find any cases?

phdi/fhir/tabulation/tables.py
bamader: Yeah I think this is a dangerous approach, see my longer comment above about moving conversion to write instead of tabulate

bamader (changes requested):
No need to repeat myself, my big concern is in the longer comment about what level we convert to string at

nickbristow:
@bamader changes have been made

phdi/tabulation/tables.py
```diff
@@ -102,6 +102,14 @@ def write_data(
       incremental writing to a parquet destination is desired. Omit if
       `output_type` is not Parquet. Default: `None`.
     """
+    for i in range(1, len(tabulated_data)):
+        tableList = tabulated_data[i]
```
bamader, line 106: You're gonna hate this lol but in python we avoid camel case variables like the plague. This needs to become `table_list`.

phdi/tabulation/tables.py
```diff
@@ -102,6 +102,14 @@ def write_data(
...
+        for n, v in enumerate(tableList):
```
bamader, line 107: Let's be more descriptive with the `v` in the enumeration (and I'd prefer the `n` as well). Since this is an iteration over a table row, I'd go with something like `for row, elt in enumerate(table_row)`.

phdi/tabulation/tables.py
```diff
@@ -102,6 +102,14 @@ def write_data(
...
+            if isinstance(v, list):
+                tabulated_data[i][n] = _convert_list_to_string(v)
+            if isinstance(v, dict):
+                tabulated_data[i][n] = _convert_dict_to_string(v)
```
bamader, line 111: It looks like the `_convert_dict_to_string()` function is just a one line cast to `str()`. If that's the case, let's delete the helper function and just call `str()` here.

phdi/tabulation/tables.py
```diff
@@ -155,3 +163,31 @@ def write_data(
 
         conn.commit()
         conn.close()
+
+
+def _convert_list_to_string(val: list) -> str:
+    """
+    Takes a parameter and checks that the value, if it is a list, uses recursion
+    to break down to the values to join into a string with a `,`.
```
bamader, line 171: We've got some rigid guidelines around docstring formatting and structure, so let's change this to the following: `Serializes a given list into a string, separating values with commas." (Rationale/explanation: We don't need tend to mention technical details in docstrings unless they're particularly relevant, so we can ditch the recursion point; there's no value checking on the input because we just assume that the input `val` is actually a list; small typos).

phdi/tabulation/tables.py
```diff
@@ -155,3 +163,31 @@ def write_data(
...
+
+    :param val: A list object that may contain other types of objects to be stringified
+    for use in CSV, SQL, etc
```
bamader, line 174: We can just say "a list," no need to mention it's an object. Also, all parameter and return docstring lines after the first should be prefaced with two spaces to simulate a half indent.

phdi/tabulation/tables.py
```diff
@@ -102,6 +102,14 @@ def write_data(
       incremental writing to a parquet destination is desired. Omit if
       `output_type` is not Parquet. Default: `None`.
     """
+    for i in range(1, len(tabulated_data)):
+        tableList = tabulated_data[i]
+        for n, v in enumerate(tableList):
+            if isinstance(v, list):
```
bamader, line 108: Can you add a comment before this line that "some elements may themselves contain lists, if selection_criteria = all is used"? This way it's clear in the code we're not just directly calling the join here.

bamader (changes requested):
Code logic looks good. Just some changes needed now to bring it into a pythonic style and in line with our repo guidelines.

phdi/tabulation/tables.py
nickbristow: i hate that

nickbristow:
@bamader changes made

phdi/tabulation/tables.py
bamader: It is the python (and superior 😉 ) way

bamader (approved):
Looks better to me. Assuming Bryan has no issues with the code or the styling this should be good.

bryan-skylight (approved):
LGTM!

I'd prefer that the "serialization" functionality be in its own function, but I don't think that's a hard requirement, and certainly not worth keeping this PR open for. We can revisit later as a refactor ticket if we so choose.

nickbristow:
Ok, closing!

PR #139, opened 2022-12-19
quick fix to re-use parquet writer for each increment of tabulated data

m-goggins:
Per @DanielPaseltiner's catch, I updated `generate_tables` to re-use the same parquet writer for each incremental result of tabulated data. With this update, we now check once to see if the user has passed in a writer and if yes, return the same writer for re-use on each subsequent incremental chunk of tabulated data.

### DIFF
--- a/phdi/fhir/tabulation/tables.py
+++ b/phdi/fhir/tabulation/tables.py
@@ -686,7 +686,7 @@ def generate_tables(
     search_urls = _generate_search_urls(schema=schema)
 
     for table_name, search_url in search_urls.items():
-        pq_writer = output_params[table_name].get("pq_writer", None)
+        pq_writer = None
         next = search_url
         while next is not None:
 
@@ -707,6 +707,8 @@ def generate_tables(
                 filename=output_params[table_name].get("filename"),
                 output_type=output_params[table_name].get("output_type"),
                 db_file=output_params[table_name].get("db_file", None),
-                db_tablename=output_params[table_name].get("db_filename", None),
+                db_tablename=output_params[table_name].get("db_tablename", None),
                 pq_writer=pq_writer,
             )
+        if pq_writer is not None:
+            pq_writer.close()  # pragma: no cover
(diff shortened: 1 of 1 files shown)


DanPaseltiner (commented):
Thanks for addressing this @m-goggins! I do wonder if we want users to be able to pass in their own writer, or if it would be better for us to just handle everything to do with the writer internally. If the user passes in a writer for a different file than the one specified in  `output_params[table_name]["filename")` things could get wonky fast. I wasn't part of the entire design process on this work so I will defer to you and @bamader on this. 

I also think that we should make sure to close the writer, if there is one on each iteration of the for loop to ensure we are closing the file connections we open.

m-goggins:
Yes, I see your point about the writer and think it makes sense to handle it internally. When we initially developed `write_data` we hadn't yet conceived of the `output_params` file. I'll make those changes unless @bamader has a different opinion.

bamader (changes requested):
I am absolutely for internal handling at this point. If our discovery brings up that users want to pass in their own writers, we can add that later, but major desire to avoid wonkiness with this right now. 

As for closing file connections, also agree with Dan on that. Since we're ostensibly re-using the same writer in each loop iteration, I think all we have to do is close the pq_writer at the end of the for loop before going on to the next table.

phdi/fhir/tabulation/tables.py
```diff
@@ -700,12 +701,12 @@ def generate_tables(
...
 
             # Write set of tabulated incremental data
-            write_data(
+            pq_writer = write_data(
                 tabulated_data=tabulated_incremental_data,
                 directory=output_params[table_name].get("directory"),
                 filename=output_params[table_name].get("filename"),
                 output_type=output_params[table_name].get("output_type"),
                 db_file=output_params[table_name].get("db_file", None),
                 db_tablename=output_params[table_name].get("db_filename", None),
```
DanPaseltiner, line 710: Should this be:
`db_tablename=output_params[table_name].get("db_tablename", None)` ?

m-goggins: Oops, yes! good catch

bamader (approved):
Looks good. Thanks again for catching this, Dan! Probably saved all of us some rather head-smashing logic down the road.

BradySkylight (approved):
Great catches and discussions!

PR #141, opened 2022-12-30
Add phdi-alerts container

nickclyde:
Resolves https://app.clickup.com/t/86696umzu

This PR adds a container which provides an API for sending alerts via SMS, Slack, or Microsoft Teams. The idea here is that we can configure this container as part of the pipeline to fire off alerts for failed conversions to FHIR, failed uploads to FHIR server, or anything else!

I have included a README with setup instructions. Currently the SMS endpoint is Azure specific, but we could theoretically add support for other SMS providers in the future.

Screenshots of successful alerts when I deployed this as a Container App in Azure and followed all the setup steps:  
  

IMG_D8A365015793-1
Screenshot 2022-12-30 at 10 22 07 AM
Screenshot 2022-12-30 at 10 22 29 AM


DanPaseltiner (changes requested):
I'm really excited to see this deployed. I would like to step up our documentation in the code so that the docs that FastAPI generates automatically are more complete. I think we should add the following:

1. Metadata for the FastAPI service.
2. Descriptions for each field in a Pydantic schema (these get pulled into the auto docs)
3. A doc string at a minimum for each endpoint function so that all the endpoints in the docs have a description. 
4. Add contents of the README to the FastAPI metadata `description` attribute. 

The tabulation service has examples on how to do all of this now. I'm happy to pair as well if you have any questions.

nickclyde:
@DanPaseltiner I think I covered everything you mentioned for improving docs, let me know if I missed anything!

DanPaseltiner (approved):
LGTM! Thanks for adding the additional documentation.

PR #142, opened 2023-01-03
Generate parquet schema based on passed data type

nickbristow:
Use passed data type to generate parquet data type. Default to string


tests/tabulation/test_tables.py
```diff
@@ -118,14 +120,30 @@ def test_write_data_parquet(patched_pa_table, patched_writer):
...
     table_to_use = tabulate_data(extracted_data, schema, "Physical Exams")
     batch_1 = table_to_use[:2]
-    batch_2 = [table_to_use[0]] + table_to_use[2:]
+    batch_2 = [table_to_use[0]] + table_to_use[2]
+    batch_3 = [table_to_use[0]] + table_to_use[3:]
     file_location = "./"
     output_file_name = "new_parquet"
     file_format = "parquet"
 
     # Batch 1 tests creating a new parquet file and returning a writer
```
m-goggins, line 129: I think we should also add an assert statement that we have returned the writer. Something like `pq_writer is not None`

tests/tabulation/test_tables.py
```diff
@@ -134,12 +152,35 @@ def test_write_data_parquet(patched_pa_table, patched_writer):
 
     # Batch 2 tests appending to existing parquet using previous writer
```
m-goggins, line 153: For this one, I would add an assertion that tests whether the additional data as has been added. Something like the length of the updated file's contents.

m-goggins (changes requested):
I think this looks good, though I will admit it has been a minute since I have looked at this section of tabulation. It would be helpful to have a docstring for the function you created and maybe a little more detail for the PR since I'm a little out of the loop.

phdi/tabulation/tables.py
```diff
@@ -181,3 +194,30 @@ def _convert_list_to_string(val: list) -> str:
         elif type(v) != str:
             val[i] = str(v)
     return (",").join(val)
+
+
+def _create_pa_schema_from_table_schema(
```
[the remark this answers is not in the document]
nickbristow: Yes done

tests/tabulation/test_tables.py
nickbristow: Yes, done

phdi/tabulation/tables.py
```diff
@@ -181,3 +198,91 @@ def _convert_list_to_string(val: list) -> str:
...
+                    pa_schema_arr.append(pa.field(name, pa.string()))
+    pa_schema = pa.schema(pa_schema_arr)
+    return pa_schema
+
+
+def _create_from_arrays_data(row_data: List) -> List:
+    """
+    Returns a list that is one array per column. Accepts list that is one
+      array per row.
+    :param row_data: A list that is made of multiple arrays
```
m-goggins, line 245: ditto `:return:`

phdi/tabulation/tables.py
```diff
@@ -181,3 +198,91 @@ def _convert_list_to_string(val: list) -> str:
...
+    col_data = []
+    for row in row_data:
+        for i, data in enumerate(row):
+            if (len(col_data) - 1) < i:
+                col_data.append([])
+            col_data[i].append(data)
+    return col_data
+
+
+def _create_parquet_data(data: List[List], pq_schema: pa.Schema) -> List[List]:
```
m-goggins, line 258: ditto docstring

tests/tabulation/test_tables.py
```diff
@@ -146,6 +173,244 @@ def test_write_data_parquet(patched_pa_table, patched_writer):
...
+        batch_2,
+        file_location,
+        file_format,
+        output_file_name,
+        pq_writer=pq_writer,
+        schema=schema,
+        table_name=table_name,
+    )
+
+    # Batch 3 tests appending to existing parquet using previous writer
```
m-goggins, line 221: Do we need Bacth 3? How is it different than batch 2 above on line 210?

m-goggins (changes requested):
A couple of return statements and docstrings need to be added, but the tests look much more robust and I'm glad we got them figured out last week 😄

tests/tabulation/test_tables.py
m-goggins: Gotcha, that makes sense!

phdi/tabulation/tables.py
```diff
@@ -181,3 +199,101 @@ def _convert_list_to_string(val: list) -> str:
...
+    """
+    Returns a list with the data being modified to the data types specified by the
+      pyarrow schema.
+    :param data: A list of lists with multiple objects.
+    :param pq_schema: A pyarrow schema file which references the data in the data file.
+    :return: A list of lists.
```
m-goggins, line 268:
```suggestion
    :return: A list of lists with data types specified in the pyarrow schema.
```

m-goggins (approved):
Made one suggestion but it looks good to me!

phdi/tabulation/tables.py
```diff
@@ -122,7 +124,22 @@ def write_data(
...
     if output_type == "parquet":
-        table = pa.Table.from_arrays(tabulated_data[1:], names=tabulated_data[0])
+        if schema and table_name:
+            pq_schema = _create_pa_schema_from_table_schema(
+                schema, tabulated_data[0], table_name
+            )
+        else:
+            pq_schema = None
+        parquet_data = _create_parquet_data(tabulated_data, pq_schema)
+        # parquet_data = tabulated_data
```
bamader, line 134: nit: remove comment

bamader (approved):
Wow this took me longer than I wanted to get back into lol. Code itself looks good. If possible, I might suggest adding some more inline comments to explain what's going on with the code flow (I know it's not necessary in some places, like the if/elif ladders handling specific cases, but parts of this feel sufficiently in the weeds to me so as to warrant inline clarification). That's non-blocking, however, so I'll leave that to your discretion. There's also one hanging commented line left in the code that I flagged for you to delete.

nickbristow:
@bamader added some extra comments

emmastephenson:
@nickbristow to try downloading the latest version of black and running via command line.

PR #148, opened 2023-01-10
Implement double metaphone

bamader:
This PR implements the double metaphone algorithm in the PHDI library via a collection of utilities, wrappers, and FHIR operations. The number of changes looks longer than it actually is because there were changes to the `.toml` and `.lock` files in poetry, which you don't actually need to look at.


phdi/fhir/harmonization/standardization.py
```diff
@@ -1,12 +1,75 @@
...
+    dmeta = fuzzy.DMetaphone()
+    for entry in bundle.get("entry", []):
+        resource = entry.get("resource", {})
+        if resource.get("resourceType", "") == "Patient":
+            metaphone_reps = double_metaphone_patient(resource, dmeta)
+            dms[resource.get("id")] = metaphone_reps
+    return dms
+
+
+def double_metaphone_patient(patient: dict, dmeta=None) -> List[dict[str, List[str]]]:
```
m-goggins, line 32: Wondering if we would also want to consider maiden names? Or maybe how we might want to incorporate maiden names more generally?

phdi/harmonization/standardization.py
```diff
@@ -1,8 +1,31 @@
...
+def double_metaphone_string(string: str, dmeta=None) -> List[Union[str, None]]:
+    """
+    Performs the double metaphone algorithm originally developed by Lawrence
+    Phillips for the purpose of phonetic search and matching. Double Metaphone
+    is an improvement over SoundEx with broader support for languages and non-
+    English names. Returns a list holding the primary and secondary
+    phonetic encodings of the given string (including None if there is no
+    valid secondary encoding). This function experts basic text cleaning (e.g.
+    removal of numeric characters, trimming of spaces, etc.) to already
+    have been performed.
```
m-goggins, line 16: I don't think we need to include the referential stuff here. It can go in a tutorial/other documentation. Here's how I would change it: 

```
Performs the double metaphone algorithm on a given string. Returns a list holding the primary and secondary
    phonetic encodings of the given string (including None if there is no
    valid secondary encoding). This function expects basic text cleaning (e.g.
    removal of numeric characters, trimming of spaces, etc.) to already
    have been performed.
```

tests/fhir/harmonization/test_fhir_harmonization.py
```diff
@@ -1,8 +1,64 @@
...
+            / "patient_bundle.json"
+        )
+    )
+
+    # Add a couple more name uses to prove out robustness
+    patient = raw_bundle.get("entry")[1].get("resource")
+    patient.get("name").append({"family": "D", "given": ["Johnny"], "use": "usual"})
+    patient.get("name").append(
+        {"family": "Doe", "given": ["Johnathan", "Dangerson"], "use": "old"}
+    )
```
m-goggins, line 49: Can we do a non-English sounding name? Maybe one that is representative of LAC patient pop? I'm also curious to see how it handles accented names vs non-accented, e.g.,  Jose vs José.

m-goggins (changes requested):
Looks pretty good! I would just add a couple more robustness checks/names that are more specific to LAC population and demonstrate how the function handles accented names.  I also left a more general comment on if/how we want to handle maiden names.

phdi/fhir/harmonization/standardization.py
```diff
@@ -1,12 +1,75 @@
...
+      FHIR use case their resource includes.
+    """
+    dms = {}
+    dmeta = fuzzy.DMetaphone()
+    for entry in bundle.get("entry", []):
+        resource = entry.get("resource", {})
+        if resource.get("resourceType", "") == "Patient":
+            metaphone_reps = double_metaphone_patient(resource, dmeta)
+            dms[resource.get("id")] = metaphone_reps
+    return dms
```
DanPaseltiner, line 29: Should the encode strings just be returned in patient resources in a fhir bundle? Later on we won't have to constantly recompute.

phdi/fhir/harmonization/standardization.py
bamader: This actually will already consider maiden names. Maiden is one of the FHIR-recognized values for the `use` property in a `HumanName` object (https://build.fhir.org/datatypes.html#HumanName) so we'll already be covered with the DM representation of it.

phdi/harmonization/standardization.py
bamader: That's fair. The only reason I had it in there was because I think a lot of people have heard of SoundEx but most people may not have heard of double metaphone, even though it's developed by the same guy and he was improving on his work. But it's definitely not something relevant to the code functioning so I can move it elsewhere.

tests/fhir/harmonization/test_fhir_harmonization.py
bamader: Yeah that's a good idea. Good examples that come to mind (in addition to Jose) are Dominic, Alejandro, Diego, Xavier, Jorge, Eduardo, and Sofia vs Sophia. I think I'm definitely going to use the Sofia/Sophia as part of the similarity assertions, but any favorites for the more general processing part of the tests?

phdi/fhir/harmonization/standardization.py
m-goggins: Oh awesome. For some reason I thought it was stored differently, but you are correct.

tests/fhir/harmonization/test_fhir_harmonization.py
m-goggins: Those seem sufficient to me for now

phdi/fhir/harmonization/standardization.py
bamader: Just noting here that you and I spoke yesterday and then Brady and I spoke, and we've got an extension storage solution implemented. Great catch, and this is now solved.

phdi/harmonization/standardization.py
```diff
@@ -1,8 +1,28 @@
...
 import pycountry
 from typing import Literal, List, Union
 
 
+def double_metaphone_string(string: str, dmeta=None) -> List[Union[str, None]]:
+    """
+    Performs the double metaphone phonetic encoding algorithm on the given string.
+    Returns a list holding the primary and secondary phonetic representations
+    of the string (including None if there is no valid secondary encoding).
+    This function experts basic text cleaning (e.g. removal of numeric characters,
```
m-goggins, line 12: **expects

m-goggins (approved):
I found one remaining typo to fix but other than that, it looks good to go!

phdi/harmonization/standardization.py
bamader: Lol, function clearly not written by an expert 🤣

PR #150, opened 2023-01-11
Implement string comparison utility function

m-goggins:
This PR adds the `compare_strings` utility function to the harmonization library. It returns the normalized similarity between two given strings and gives the user options for which `similarity_measure` they wish to implement. I chose the normalized_similarity as the return value because it makes it easier to compare across different similarity_measures and is more intuitive (I think) than integers. 

I defaulted to Jaro-Winkler because its scores were most intuitive to understand for a handful of examples I tested and was faster as well. This will likely need more testing. 

I did not include the Hamming distance metric, as suggested in the [ticket](https://app.clickup.com/t/865bbbd4x), because it expects the input strings to be the same length, which the user would not necessarily know/expect to be true if processing a large number of patient names.


pyproject.toml
```diff
@@ -26,6 +26,7 @@ fhirpathpy = "0.1.0"
 google-auth = "^2.10.0"
 google-cloud-storage = "^2.5.0"
 jsonschema = "4.16.0"
+rapidfuzz = "^2.13.6"
```
DanPaseltiner, line 29: Seems like we are pulling in at least 2 new libraries for fuzzy matching. Is there a single one that offers all the features we need. Not worries if not, but we should be mindful of increasing our already large number of dependencies.

phdi/harmonization/utils.py
```diff
@@ -0,0 +1,41 @@
...
+     - Levenshtein: the number of edits (excluding transpositions) needed to transform
+        string1 into string2.
+     - DamerauLevenshtein: the number of edits (including transpositions) needed to
+        transform string1 into string2.
+    :return: The normalized similarity between string1 and string2, with 0 representing
+        no similarity between string1 and string2, and 1 meaning string1 and string2 are
+        dentical words.
+    """
+    if similarity_measure == "JaroWinkler":
+        return rapidfuzz.distance.JaroWinkler.normalized_similarity(string1, string2)
```
DanPaseltiner, line 35:
```suggestion
similarity_measures = {"JaroWinkler": rapidfuzz.distance.JaroWinkler.normalized_similarity, "Levenshtein":apidfuzz.distance.Levenshtein.normalized_similarity, "DamerauLevenshtein":rapidfuzz.distance.DamerauLevenshtein.normalized_similarity}

return similarity_measures[similarity_measure](string1, string2)
```

phdi/harmonization/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    elif similarity_measure == "Levenshtein":
+        return rapidfuzz.distance.Levenshtein.normalized_similarity(string1, string2)
+    elif similarity_measure == "DamerauLevenshtein":
+        return rapidfuzz.distance.DamerauLevenshtein.normalized_similarity(
+            string1, string2
```
DanPaseltiner, line 40: I'm pretty sure something like this will work. What do you think? My thought is that the additional abstraction makes adding new measures easier in the future.

```suggestion
similarity_measures = {"JaroWinkler": rapidfuzz.distance.JaroWinkler.normalized_similarity, "Levenshtein":apidfuzz.distance.Levenshtein.normalized_similarity, "DamerauLevenshtein":rapidfuzz.distance.DamerauLevenshtein.normalized_similarity}

return similarity_measures[similarity_measure](string1, string2)
```

DanPaseltiner (approved):
LGTM. Two comments for you to consider and do what you like with. Cool to learn about other options beyond straight Levenshtein.

pyproject.toml
DanPaseltiner: That makes sense. That fact that we put some thought into this satisfies me. Thanks!

bamader:
About to review now, but a couple notes:

1. absolutely agree with Jaro-Winkler as the default (I thought I put that in the ticket but I guess I didn't lol). Your observation about it being faster are correct (in the NLP work I did before this JW is roughly 3 times faster than DL because DL needs a full dynamic programming table to be calculated), which is key for the large number of calculations we're going to run. Also, I agree with its normalization being more intuitive, but importantly, most literature out there confirms that it performs more stably and with better "expected" results when used on short strings such as names (due to prefix substrings being given slightly higher weight than DL). So yeah we should always be preferring this unless the user is specifically interested in an edit-distance formulation.
2. Hamming is relatively easily modified to use intersection substrings based on the longer of the two strings to-compare, but given my own experience with JW I can't really see Hamming beating it out, so I'm fine to leave Hamming out for the time being.

phdi/harmonization/utils.py
bamader: That would work, yeah. For readability I personally prefer the if/elif ladder Marcelle has constructed, but no real strong preferences on that!

bamader (changes requested):
Code itself looks good. I think the only suggestion I have is adding some test cases for non-English names, which turned out to be a good idea with the double metaphone PR. Feel free to look at the comments I had in there about good candidate names.

phdi/harmonization/utils.py
m-goggins: Yeah, I always err on the side of readability. Especially since it seems our SDK users are less likely to have extensive Python experience.

bamader (approved):
Looks good!

PR #153, opened 2023-01-12
Require schema_name to be included in schema

DanPaseltiner:
Require schema_name to be included in schema


m-goggins (approved):
Looks good!

PR #154, opened 2023-01-17
Install gcc in Docker images for Fuzzy

nickclyde:
This will allow the `tabulation` and `ingestion` images to build properly. We can remove the installation of gcc if we find an alternative to the `Fuzzzy` package at a later date.


DanPaseltiner (approved):
Given that only `tabulation` is installing `phdi` from GitHub I think we only need this change for `tabulation. Either way, thanks for putting this together to just keep things working for now while we figure out what to do about Fuzzy.

PR #155, opened 2023-01-18
Create Nickname DB

bamader:
This PR creates the database of nicknames that PHDI will use by aggregating 6 open source academic projects into a single combined file. The originally cited files are not included, but the projects are cited appropriately in an Acknowledgments section of our README. The basic script used to aggregate the DB is also left in the examples directory.

Per conversation with Marcelle and Brady, we'll use nicknames later on when it comes to generating candidate blocking pairs, rather than incorporating additional standardization now. As such, no functions are yet implemented that use nicknames.


phdi/harmonization/standardization.py
```diff
@@ -194,3 +195,14 @@ def standardize_name(
     if isinstance(raw_name, str):
         return outputs[0]
     return outputs
+
+
+def _build_nicknames_db():
```
m-goggins, line 200: It might have been my idea to include the construction of the `nicknames_to_names` dict in harmonization 😅 but I think we should remove it and just use the resulting dictionary as an asset. If it isn't an especially large file, I wouldn't want users to get hung up on forgetting to build an asset and then not be able to use it. I guess the case could be made to hold off on removing this until we know how/when we are going to use the nickname resolution piece so that we're not hosting a bunch of data on github. I don't feel too strongly so feel free to make the decision.

examples/compile_nickname_db.py
```diff
@@ -0,0 +1,68 @@
...
+    for line in fp:
+        if line.strip() != "":
+            if i == 0:
+                i += 1
+                continue
+            toks = line.strip().split(",")
+            root_name = toks[1].strip().upper()
+            if root_name not in names_to_nicknames:
+                names_to_nicknames[root_name] = []
+            names_to_nicknames[root_name].append(toks[2].strip().upper())
```
m-goggins, line 21: This is very nitpicky so feel free to ignore, but would it be possible to standardize the vars so that this line reads the same for each of the files, i.e., something line `names_to_nicknames[root_name].append(nick_name[2].strip().upper())` so that the user knows the same thing is being added in the same way from each file.

m-goggins (approved):
Looks good! I made some nitpicky suggestions but they are not necessary if you disagree.

README.md
```diff
@@ -68,6 +68,16 @@ The PRIME Public Health Data Infrastructure prototype a sibling project to [PRIM
...
 
+## Additional Acknowledgments 
+
+The compiled database mappings of root names to various nicknames was produced by the aggregation and synthesis of open source work from a number of projects. While we do not employ the packages and wrappers used by the various projects (merely their open source data), we do wish to give credit to their various works building collections of nickname mappings. These projects are:
+
+* [Secure Enterprise Master Patient Index](https://github.com/MrCsabaToth/SOEMPI), based on OpenEMPI, conducted by Vanderbilt University
+* [Curated Nicknames](https://github.com/carltonnorthern/nicknames), scraped from genealogy webpages and run by Old Dominion University Web Science and Digital Libraries Research Group
+* [Simple Public Domain Nickname Mappings](https://github.com/onyxrev/common_nickname_csv), hand collected using various sources
+* [Lingua En Nickname](https://github.com/brianary/Lingua-EN-Nickname), collected from a series of GenWeb projects
+* [diminutives.db](https://github.com/HaJongler/diminutives.db), compiled via a nickname extract using Wikipedia and Wiktionary
```
BradySkylight, line 79: Awesome!

BradySkylight (approved):
This looks good to me and I like the discussion/thoughts that Marcelle has brought up.

phdi/harmonization/standardization.py
bamader: This is a good point. I think once we figure out what our blocking routine looks like and how we access information there, we can move this somewhere better. For now, though, because the file isn't big, I'm just gonna leave this here as is (because both of our instincts/intuition say to look in / use harmonization lol).

examples/compile_nickname_db.py
bamader: Yeah for sure, great call out. The only reason it's not all in one loop is because the files are differently formatted (e.g. some were nicks to names, some names to nick, some had multiple per line, others were singular, etc.). But standardizing variables will be a good way to keep that all smooth.

PR #156, opened 2023-01-19
Implement in house double metaphone

bamader:
This PR implements the double metaphone algorithm with specific in-code callouts for Spanish-language-origin phonetic behavior. All tests passed after implementation was complete, with one exception, and in that case, the new algorithm actually imo improved the phonetic representation (originally, the name "Price" was represented with the sound clusters "PRK", but with our new implementation, the sounds are "PRS" which is actually how the name sounds). This PR is going to get flagged as failing various build steps and checks, but that's because Sphinx and Poetry are currently fighting installation-wise. Bryan and I are looking into this issue, but I wanted to get the logic up so that folks could get eyes on and be ready to approve once the package management settles.

### DIFF
--- a/phdi/harmonization/double_metaphone.py
+++ b/phdi/harmonization/double_metaphone.py
@@ -1,7 +1,7 @@
 import unicodedata
 
 """
-The Double Metaphone phonetic encoding algorithm is an improvement to the 
+The Double Metaphone phonetic encoding algorithm is an improvement to the
 phonetic representation algorithm Soundex, commonly used to identify what an
 English name "sounds like." However, rather than produce an exact phonetic
 description of a name, Double Metaphone produces an intentionally approximate
@@ -150,7 +150,7 @@ class DoubleMetaphone(object):
             position == (start_index + 1) and buffer[start_index] == "M"
         ):
 
-            #'bellocchio' but not 'bacchus'
+            # 'bellocchio' but not 'bacchus'
             if (
                 buffer[position + 2] in ["I", "E", "H"]
                 and buffer[position + 2 : position + 4] != "HU"
--- a/phdi/harmonization/double_metaphone.py
+++ b/phdi/harmonization/double_metaphone.py
@@ -1,0 +1,815 @@
+import unicodedata
+
+"""
+The Double Metaphone phonetic encoding algorithm is an improvement to the 
+phonetic representation algorithm Soundex, commonly used to identify what an
+English name "sounds like." However, rather than produce an exact phonetic
+description of a name, Double Metaphone produces an intentionally approximate
+phonetic representation, in order to account for the way speakers vary their
+pronunciations and misspell or otherwise vary words and names. It also
+returns both a primary and a secondary encoding for a string (hence the
+name "Double" Metaphone), which accounts for some ambiguous cases as well as
+for multiple variants of surnames with common ancestry. Double Metaphone is
+capable of accounting for irregularities in the spelling and phonetics of
+English, Slavic, Germanic, Celtic, Greek, French, Italian, Spanish, Chinese,
+and more.
+"""
+
+VOWELS = ["A", "E", "I", "O", "U", "Y"]
+SILENT_STARTERS = ["GN", "KN", "PN", "WR", "PS"]
+
+
+class DoubleMetaphone(object):
+    """
+    A Double Metaphone parsing object capable of performing the double
+    coding algorithm on a given input string.
+    """
+
+    def __init__(self):
+        self.position = 0
+        self.primary_phone = ""
+        self.secondary_phone = ""
+        # next is set to a tuple of the next characters in the primary and
+        # secondary codes and to indicate how many characters to move forward
+        # in the string.  The secondary code letter is given only when it is
+        # different than the primary.
+        self.next = (None, 1)
+
+    def __call__(self, string: str):
+        return self.parse(string)
+
+    def check_word_start(self):
+        # Skip silent letters when they start a word (because they're not
+        # phonetic if they're silent)
+        if self.word.get_letters(0, 2) in SILENT_STARTERS:
+            self.position += 1
+        # Starting with 'X' frequently pronounced 'Z' e.g. 'Xavier'
+        if self.word.get_letters(0) == "X":
+            # 'Z' maps to 'S'
+            self.primary_phone = self.secondary_phone = "S"
+            self.position += 1
+
+    def process_vowel(self):
+        self.next = (None, 1)
+
+        # All starting vowels map to 'A'
+        if self.position == self.word.start_index:
+            self.next = ("A", 1)
+
+    def process_b(self):
+        # "-mb", e.g., "dumb", already accounted for, see 'M'
+        if self.word.buffer[self.position + 1] == "B":
+            self.next = ("P", 2)
+        else:
+            self.next = ("P", 1)
+
+    def process_c(self):
+        buffer = self.word.buffer
+        position = self.position
+        start_index = self.word.start_index
+
+        # Handle some germanic linguistics
+        if (
+            position > start_index + 1
+            and buffer[position - 2] not in VOWELS
+            and buffer[position - 1 : self.position + 2] == "ACH"
+            and buffer[position + 2] not in ["I"]
+            and (
+                buffer[position + 2] not in ["E"]
+                or buffer[position - 2 : position + 4] in ["BACHER", "MACHER"]
+            )
+        ):
+            self.next = ("K", 2)
+
+        # Special case 'CAESAR'
+        elif (
+            position == start_index
+            and buffer[start_index : start_index + 6] == "CAESAR"
+        ):
+            self.next = ("S", 2)
+
+        # Italian 'chianti'
+        elif buffer[position : position + 4] == "CHIA":
+            self.next = ("K", 2)
+        elif buffer[position : position + 2] == "CH":
+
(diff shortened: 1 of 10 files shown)


phdi/harmonization/double_metaphone.py
```diff
@@ -0,0 +1,815 @@
...
+                    buffer[end_index - 1 : end_index + 1] in ["AS", "OS"]
+                    or buffer[end_index] in ["A", "O"]
+                )
+                and buffer[position - 1 : position + 3] == "ALLE"
+            ):
+                self.next = ("L", "", 2)
+            else:
+                self.next = ("L", 2)
+        else:
+            self.next = ("L", 1)
```
m-goggins, line 415: If I am understanding this correctly, it doesn't look like the code accounts for the "y" sound that the double L makes in Spanish pronunciation for a word like 'tortilla'. Should it be `self.next = ("Y", 1)` ?

phdi/harmonization/double_metaphone.py
```diff
@@ -0,0 +1,815 @@
...
+        elif buffer[position + 1] in ["T", "D"]:
+            self.next = ("T", 2)
+        else:
+            self.next = ("T", 1)
+
+    def process_v(self):
+        if self.word.buffer[self.position + 1] == "V":
+            self.next = ("F", 2)
+        else:
+            self.next = ("F", 1)
```
m-goggins, line 596: Should this be a `V` instead? Maybe I just need an example word to understand. I was thinking of names like Veronica or Ava.

phdi/harmonization/double_metaphone.py
```diff
@@ -0,0 +1,815 @@
...
+        ) or buffer[position + 1] == "M":
+            self.next = ("M", 2)
+        else:
+            self.next = ("M", 1)
+
+    def process_n(self):
+        if self.word.buffer[self.position + 1] == "N":
+            self.next = ("N", 2)
+        else:
+            self.next = ("N", 1)
```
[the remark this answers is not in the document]
m-goggins: Edit: nvm I see `unicode.decode` below

phdi/harmonization/double_metaphone.py
```diff
@@ -0,0 +1,815 @@
...
+                for c in unicodedata.normalize("NFD", self.decoded)
+                if unicodedata.category(c) != "Mn"
+            )
+        )
+        self.upper = self.normalized.upper()
+        self.length = len(self.upper)
+        self.prepad = "  "
+        self.start_index = len(self.prepad)
+        self.end_index = self.start_index + self.length - 1
+        self.postpad = "      "
```
m-goggins, line 790: For my own understanding, why the different lengths for pre- and post-padding?

m-goggins (commented):
Whew, lots to review but it looks pretty good to me! I left comments that are more about clarification (handle an additional case, make sure I understand how we are handling accents) than about the code itself.

phdi/harmonization/double_metaphone.py
```diff
@@ -0,0 +1,815 @@
...
+            ):
+                # 'accident', 'accede' 'succeed'
+                if (
+                    position == (start_index + 1) and buffer[start_index] == "A"
+                ) or buffer[position - 1 : position + 4] in ["UCCEE", "UCCES"]:
+                    self.next = ("KS", 3)
+
+                # 'bacci', 'bertucci', other italian
+                else:
+                    self.next = ("X", 3)
```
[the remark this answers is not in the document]
bamader: On the surface of the substance, not really, since it looks like they're both marking the next phonetic sound as X and incrementing by 3. But in terms of the words they represent, there are potentially names out there that might end in `-CIA` vs having a double `-CC-` sound somewhere in the middle that _isn't_ the starting position. The else there just marks it as a default case if we think we're in a double C word that won't have vowel phonetics afterward (the conditional on line 144 has that `IA`, whereas if you look at the selection criteria on 154, we can't even get into this `else` clause if there's an A coming up in the buffer). Definitely confusing, but they do accomplish different things!

phdi/harmonization/double_metaphone.py
bamader: Nope, though good catch that it looks like that sound isn't being handled! If you look at the top of the code, 'Y' is included in the list of vowels as recognized by the program. Double Metaphone treats all non-starter vowels as "collapsible", meaning that they typically have lower information content than the consonants around them and are thus not represented in the final phonetic encodings. When you see the `self.next = ("L", "", 2)`, it's using the `L` on its own in the primary encoding to proxy for one type of (wrong lol) pronunciation, and then adding nothing in the secondary encoding in order to not take up one of the four available encoding slots (ostensibly so that the spot can go to a more informative consonant sound). The other `else` clauses do something similar, but just continue with the information content of the "L". Double Metaphone specifically isn't guaranteed to produce _realistic_ sound encodings if they're read on their own, but instead it creates sound encodings that are designed to _differentiate_ phonetics in words from one another. So we're actually all good here even though it seems like we're missing a sound.

phdi/harmonization/double_metaphone.py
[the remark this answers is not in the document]
bamader: Yep, the idea is to just use surrounding sounds to account for pronunciation variances rather than forcing our code to handle accents separately

phdi/harmonization/double_metaphone.py
```diff
@@ -0,0 +1,815 @@
...
+
+        # French e.g. 'Resnais', 'Artois'
+        elif position == end_index and buffer[position - 2 : position] in ["AI", "OI"]:
+            self.next = ("", "S", 1)
+        else:
+            self.next = ("S",)
+            if buffer[position + 1] in ["S", "Z"]:
+                self.next = self.next + (2,)
+            else:
+                self.next = self.next + (1,)
```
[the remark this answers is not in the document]
bamader: LOL maybe just a lil bit 🤣 . I would definitely prefer to stay away from adding an `is_french` (for a lot of reasons I think us trying to create language-origin functions will be really hard to figure out, to say nothing of how they'll impact unexpected words and names elsewhere in the algorithm). And I think _maybe_ we could experiment with more catches, but I definitely worry that there would be more common names that we're now misclassifying for more edge-case style words like Marseilles (which would actually already terminate with the encoding 'MRSL' before ever needing to decode the LLE ending). The lexicon of names is so large and we have relatively little linguistic information to go on here that it feels really risky to try to eke out a couple better encodings for what I'd guess is a pretty small subset of words. Thoughts?

phdi/harmonization/double_metaphone.py
bamader: Lol it's such a bad practice reason, but it's to avoid checking if an index is within the bounds of the buffer 8 million times 😅 . There are _so_ many checks in the code where we say `if [position : position+3] ... `, and the extra postpending space lets us avoid having to check whether position would extend past the buffer every single time (on the other hand, the checks about prepending position use 1 and 2 as values so we only need a couple spaces beforehand). It's a laziness + readability thing.

tests/fhir/harmonization/test_fhir_harmonization.py
```diff
@@ -23,16 +24,16 @@ def test_double_metaphone_bundle():
     dm_bundle = double_metaphone_bundle(raw_bundle, overwrite=False)
     dms = {
         "907844f6-7c99-eabc-f68e-d92189729a55": {
-            "familyName": [b"PRK", None],
-            "givenName": [[b"KMPR", None]],
+            "familyName": ["PRS", ""],
+            "givenName": [["KMPR", ""]],
```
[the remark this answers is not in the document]
bamader: Yes definitely!

phdi/harmonization/double_metaphone.py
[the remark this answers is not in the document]
m-goggins: Gotcha, that makes sense to me!

phdi/harmonization/double_metaphone.py
m-goggins: _Double Metaphone specifically isn't guaranteed to produce realistic sound encodings if they're read on their own, but instead it creates sound encodings that are designed to differentiate phonetics in words from one another._ 

Okay, that clears up a lot of the confusion I had been having because I was thinking of those differentiations as needing to be realistic. It's not that we want tortilla to be represented with the Y sound, it's that we would want to differentiate between torilla and tortila (one L).

phdi/harmonization/double_metaphone.py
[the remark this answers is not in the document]
m-goggins: I think you're right that we don't want to accidentally introduce more misclassifications! Once we start testing this at scale, I think we'll be able to see if we're missing any additional differentiations.

phdi/harmonization/double_metaphone.py
m-goggins: ha, okay, Feronica and Afa it is!

phdi/harmonization/double_metaphone.py
m-goggins: Yeah, that makes a lot of sense!

m-goggins (approved):
Looks good, thanks for clarifying!

phdi/harmonization/double_metaphone.py
bamader: _This exactly_! It's almost akin to FHIR is great for machine reading but not perfect for human reading; double metaphone is great for pronunciation differentiating, but not necessarily for perfect pronunciation representation

PR #157, opened 2023-01-24
Create ecr_sample_manual_result.json

nickbristow:
Result of fhir example files for other two files


BradySkylight (approved):
this will be nice to have.

PR #159, opened 2023-01-26
Create linkage match function

bamader:
This PR introduces the match function that determines whether records in a given block of a data set are matches to one another. Developed with an eye towards modularity in swapping rules and heuristics in/out, so it uses a dictionary of callables as well as a passible match criterion to make the determination.

### DIFF
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -145,7 +145,7 @@ def eval_perfect_match(feature_comparisons: List) -> bool:
     """
     return sum(feature_comparisons) == len(feature_comparisons)
 
-  
+
 def block_parquet_data(path: str, blocks: List) -> Dict:
     """
     Generates dictionary of blocked data where each key is a block and each value is a
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -25,7 +25,7 @@ def match_within_block(
     feature_funcs: dict[int, Callable],
     match_eval: Callable,
     **kwargs
-) -> None:
+) -> List[tuple]:
     """
     Performs matching on all candidate pairs of records within a given block
     of data. Actual partitioning of the data should be done outside this
@@ -48,7 +48,8 @@ def match_within_block(
     comparisons constitute a match according to X.
 
     :param block: A list of records to check for matches. Each record in
-      the list is itself a list of features.
+      the list is itself a list of features. The first feature of the
+      record must be an "id" for the record.
     :param feature_funcs: A dictionary mapping feature indices to functions
       used to evaluate those features for a match.
     :param match_eval: A function for determining whether a given set of
@@ -57,37 +58,22 @@ def match_within_block(
       in the block of data of records deemed to match.
     """
     match_pairs = []
-    checked = {}
-    for i in range(len(block)):
-        record_i = block[i]
-        id_i = record_i[0]
 
-        # Store "visited" sets so that if see j matches i, we don't later
-        # have to recompute whether i matches j
-        if id_i not in checked:
-            checked[id_i] = set()
-        for j in range(len(block)):
+    # Dynamic programming table: order doesn't matter, so only need to
+    # check each combo of i,j once
+    for i, record_i in enumerate(block):
+        for j in range(i + 1, len(block)):
             record_j = block[j]
-            id_j = record_j[0]
-            if record_i != record_j:
-                if id_j not in checked:
-                    checked[id_j] = set()
+            feature_comps = [
+                feature_funcs[x](record_i, record_j, x, **kwargs)
+                for x in range(len(record_i))
+                if x in feature_funcs
+            ]
 
-                    # Mark the record pair as "visited"
-                    # Then, compute feature-wise comparisons
-                    if id_j not in checked[id_i] or id_i not in checked[id_j]:
-                        checked[id_i].add(id_j)
-                        checked[id_j].add(id_i)
-                        feature_comps = [
-                            feature_funcs[x](record_i, record_j, x, **kwargs)
-                            for x in range(len(record_i))
-                            if x in feature_funcs
-                        ]
-
-                        # If it's a match, store the result
-                        is_match = match_eval(feature_comps)
-                        if is_match:
-                            match_pairs.append((i, j))
+            # If it's a match, store the result
+            is_match = match_eval(feature_comps)
+            if is_match:
+                match_pairs.append((i, j))
 
     return match_pairs
 
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -1,4 +1,6 @@
 import hashlib
+from phdi.harmonization.utils import compare_strings
+from typing import List, Callable
 
 
 def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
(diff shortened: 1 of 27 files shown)


phdi/linkage/link.py
```diff
@@ -16,3 +18,142 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+                            for x in range(len(record_i))
+                            if x in feature_funcs
+                        ]
+
+                        # If it's a match, store the result
+                        is_match = match_eval(feature_comps)
+                        if is_match:
+                            match_pairs.append((i, j))
+
+    return match_pairs
```
m-goggins, line 92:
```suggestion
   def match_within_block(
    block: List[List],
    feature_funcs: dict[int, Callable],
    match_eval: Callable,
    **kwargs
) -> List[list]:
    """
    Performs matching on all candidate pairs of records within a given block
    of data. Actual partitioning of the data should be done outside this
    function, as it compares all possible pairs within the provided partition.
    Uses a given construction of feature comparison rules as well as a
    match evaluation rule to determine the final verdict on whether two
    records are indeed a match.

    A feature function is of the form "feature_match_X" for some condition
    X; it must accept two records (lists of data), an index i in which the
    feature to compare is stored, and the parameter **kwargs. It must return
    a boolean indicating whether the features "match" for whatever definition
    of match the function uses (i.e. this allows modular logic to apply to
    different features in the compared records). Note that not all features
    in a record need a comparison function defined.

    A match evaluation rule is a function of the form "eval_X" for some
    condition X. It accepts as input a list of booleans, one for each feature
    that was compared with feature funcs, and determines whether the
    comparisons constitute a match according to X.

    :param block: A list of records to check for matches. Each record in
      the list is itself a list of features.
    :param feature_funcs: A dictionary mapping feature indices to functions
      used to evaluate those features for a match.
    :param match_eval: A function for determining whether a given set of
      feature comparisons constitutes a match for linkage.
    :return: A list of lists of the form [i,j], where i,j give the indices
      in the block of data of records deemed to match.
    """
    matches = {}
    for i, record in enumerate(block):
        if i in matches.keys():
            continue
        for j in range(i + 1, len(block)):
            next_record = block[j]
            feature_comps = [
                feature_funcs[x](record, next_record, x, **kwargs)
                for x in range(len(record))
                if x in feature_funcs
            ]
            # If it's a match, store the result
            if match_eval(feature_comps):
                if i not in matches.keys():
                    # intialize match set
                    matches[i] = set()
                if j not in matches.keys():
                    # intialize match set (if neccessary)
                    matches[j] = set()
                matches[j].update(
                    matches[i]
                )  # for the new match, add all previous matches
                matches[j].add(i)  # for the new match, add current record to matches
                matches[i].add(j)  # for the current record, add the new match
                # for all of the other matches, add the new match so that all each
                # record that has matches knows ALL of the other records it matches with
                for previous_match in matches[i]:
                    if previous_match != j:
                        matches[previous_match].add(j)
    collapsed_matches = collapse_matches(matches)

    return collapsed_matches


def collapse_matches(matches_dict: dict) -> List[list]:
    """
    Collapses a dictionary containing all block indices with their matches into a list
    of lists where each interior list represents the indices in a block whose records
    were deemed to be matches.

    :param matches_dict: Dictionary of block indices and matches.
    :return: List of lists containing all indices in a given block that are deemed to
    have matching records.
    """
    output = []
    indices_to_skip = []
    for idx, values in matches_dict.items():
        if idx not in indices_to_skip:
            matches = [idx, *values]
            indices_to_skip += [i for i in matches]
            output.append(matches)
    return output

```

m-goggins: I started adding individual comments but it was hard for me to keep track of the changes that needed to happen so I made this giant suggestion instead. I think your approach works if the 1st row has a single match later in the block but I couldn't produce matches for a block where record 2 and 3 were matches for some reason...still a little puzzled as to why that would be the case because your code seems like it would handle that case. I don't think your solution accounts for records having multiple matches, e.g., records 1,3,and 5 all match.

The primary reason for my suggestion, however, was performance. Looping over `i in range(len(block))` and `j in range(len(block))` seemed pretty expensive and not necessary since you know that you have already checked all the prior records, i.e., on i=2 you don't need to check j=1 because you already looked for the match at i=1 and j=2.

phdi/linkage/link.py
```diff
@@ -16,3 +18,142 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+    """
+    Determines whether a given set of feature comparisons represent a
+    'perfect' match (i.e. whether all features that were compared match
+    in whatever criteria was specified for them).
+
+    :param feature_comparisons: A list of 1s and 0s, one for each feature
+      that was compared during the match algorithm.
+    :return: The evaluation of whether the given features all match.
+    """
+    return sum(feature_comparisons) == len(feature_comparisons)
```
m-goggins, line 159: I will return to look more closely at these next week with fresh eyes (the matching loop took me SO long 😮‍💨) but these look really good, especially the flexibility they provide.

m-goggins (changes requested):
I haven't finished reviewing completely, as I want to look at the feature matching with fresh eyes, but I think this is off to a good start! I made some suggestions to `match_within_block` that will improve performance and I believe will allow us to handle cases where more than 2 records are matches. Happy to discuss further or answer any questions.

phdi/linkage/link.py
```diff
@@ -16,3 +18,142 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+    :param match_eval: A function for determining whether a given set of
+      feature comparisons constitutes a match for linkage.
+    :return: A list of 2-tuples of the form (i,j), where i,j give the indices
+      in the block of data of records deemed to match.
+    """
+    match_pairs = []
+    checked = {}
+    for i in range(len(block)):
+        record_i = block[i]
+        id_i = record_i[0]
```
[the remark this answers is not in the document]
bamader: Yeah totally fair, that's a good catch

phdi/linkage/link.py
bamader: Oh lol duh I was trying to dynamic programming but used an `or` instead of an `and`. Classic. Either way though I prefer the way you've indexed `j` to start at `i+1` outright though.

I've been digging into the weird test behavior you observed and I also can't figure out why it's not catching later blocks of matches. Like the code definitely should, but I'm going to change some stuff around and see if I can figure out why it's happening.

bamader: Welp no discovery of the original bug, but changing to _actual_ dynamic programming, eliminating the checked sets, and simplifying the candidate processing has now allowed me to catch later groups of matches, including lengths of longer than two. Moreover, with the new approach, we actually catch all appropriate pairwise matches, which imo is an improvement over what my original version had. That version caught only the matches deriving from a single "root" master patient, which made some implicit assumptions that got us into the transitivity situation we discussed with Bryan where B could match A and B could match C, but A didn't match C. Now, this behavior is actually modeled so that we can decide downstream which patients should be grouped under which masters.

phdi/linkage/link.py
```diff
@@ -16,3 +18,128 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+            feature_comps = [
+                feature_funcs[x](record_i, record_j, x, **kwargs)
+                for x in range(len(record_i))
+                if x in feature_funcs
+            ]
+
+            # If it's a match, store the result
+            is_match = match_eval(feature_comps)
+            if is_match:
+                match_pairs.append((i, j))
```
m-goggins, line 76: If I am understanding this correctly, in the case where three records match each other we would end up with an output like `[(1,2),(1,3),(2,3)]`, right? I didn't think this would be a very efficient way to identify multiple matches that ultimately need to be collapsed into a single ID. In the suggestion below, specifically, lines 133-142 and the `collapse_matches` function, we would end up with an output like` [[1,2,3]]`, indicating that records 1, 2, and 3 all match each other. I _think_ my solution would also take care of @bryan-skylight's worry about batching vs. streaming mis-matching.

phdi/linkage/link.py
m-goggins: Ugh, yeah. I spent over an hour trying to identify the source but found that writing new code was easier and solved the issue🤷‍♀️

phdi/linkage/link.py
m-goggins: I guess the form of the output also depends on how we want to go about collapsing/merging matches after they are found.

bamader: Yeah I think the key is that we don't yet know how to handle "transitivity" so to speak. In the case you outline, where all records have their own pairs, I completely agree they should ultimately be boiled down to a single master patient; but in a case where we have an excerpt like `[(1,2), (2,3)]` and `(1,3)` never appears, I don't think we've firmly decided on an approach yet. So my plan was to leave off any kind of collapsing/match grouping function from this work, create a separate ticket for it (perhaps aided by some discovery with LAC, or better yet maybe some experimentation using their test data set for what gives the best accuracy), and then probably experimentally determine the right way to handle clustering logic.

I don't feel super strongly though so if you'd like that work to be done here, I'm cool to implement that as well!

m-goggins: Hm, I think I may not have explained my code suggestion well enough. The chunk below I think would account would account for the scenario you pointed out (lines 

"""
            if match_eval(feature_comps):
	                if i not in matches.keys():
	                    # intialize match set
	                    matches[i] = set()
	                if j not in matches.keys():
	                    # intialize match set (if neccessary)
	                    matches[j] = set()
	                matches[j].update(
	                    matches[i]
	                )  # for the new match, add all previous matches
	                matches[j].add(i)  # for the new match, add current record to matches
	                matches[i].add(j)  # for the current record, add the new match
	                # for all of the other matches, add the new match so that all each
	                # record that has matches knows ALL of the other records it matches with
	                for previous_match in matches[i]:
	                    if previous_match != j:
	                        matches[previous_match].add(j)
"""

In the case where 1 matches 2, 2 matches 1 and 3, and 1 does match 3 in reality. Here's how this would work:

When i=1 and j=2, matches = { 1: [2], 2:[1]}
When i =2 and j=3, matches = {1: [2,3], 2: [1,3], 3: [1,2]}

After collapsing you end up with [[1,2,3]].

m-goggins (approved):
Approving so we can move forward with testing for false positives vs false negatives when creating clusters of matches

PR #160, opened 2023-01-30
Block function parquet

m-goggins:
This PR creates a function to block testing data into distinct groups as specified by the columns the user wishes to block on. For the sake of testing LAC's data, the input is a parquet file and the output is a dictionary containing lists of lists, where the keys are the block groups and the values are the LoL of data in each block group.

Future work on blocking will have different inputs, e.g., a SQL database, and will be more efficient, e.g., leverage database partitions rather than using pandas and `groupby()`.


tests/linkage/test_linkage.py
```diff
@@ -12,3 +14,42 @@ def test_generate_hash():
...
+    if os.path.isfile("./test.parquet"):  # pragma: no cover
+        os.remove("./test.parquet")
+    test_data_df.to_parquet(path="./test.parquet", engine="pyarrow")
+
+    blocked_test_data = block_parquet_data(path="./test.parquet", blocks=["zip"])
+
+    for key, value in blocked_test_data.items():
+        print("ZIP:", key)
+        print(value)
+        print()
```
bamader, line 39: I'm guessing we probs don't want/need these print statements merged in

bamader (approved):
Seems legit 👍 . I think we've generally been making an effort to avoid pandas (which is weird because it's in our requirements as its own dependency and I'm wondering whether we actually need it there, but that's another story), but you've already noted that we'll be able to ditch this once we move to an MPI / database solution so seems totally fine as an intermediate step.

BradySkylight (approved):
This looks good to me.  Very straightforward with some python wizardry.

tests/linkage/test_linkage.py
m-goggins: ah yep!

PR #161, opened 2023-02-02
SQLite DB to mock external MPI DB

m-goggins:
This PR addresses the [ticket](https://app.clickup.com/t/8669dzn84) to create a SQLite DB file filled with mock patient data that can be used to mock up what an external MPI DB would look like as we proceed with record linkage. The example script uses synthetic data from Synthea to create a sample DB with the following columns:

- BIRTHDATE (YYYY-MM-DD)
- FIRST
- LAST
- SUFFIX
- MAIDEN
- RACE
- ETHNICITY
- GENDER (M/F only per Synthea)
- ADDRESS
- CITY
- STATE
- COUNTY
- ZIP (5-digit)
- FIRST4 (first 4 characters of first name)
- LAST4 (first 4 characters of last name)
- ADDRESS4 (first 4 characters of address with spaces removed)


examples/MPI-sample-data/build_sample_patient_sqlite_db.py
```diff
@@ -0,0 +1,47 @@
...
+
+import pandas as pd
+import sqlite3
+
+df = pd.read_csv(
+    # locally stord file of synthetic LAC synthea data
+    "~/20230106_LAC_10000_123_456/csv/patients.csv",
+    usecols=[
+        "Id",
+        "BIRTHDATE",
```
BradySkylight, line 12: This is awesome.  Just saw on the main comment for this PR that the format was YYY-MM-DD was this just a typo or is that typical for SQLite?  I assume it's actually YYYY-MM-DD.

m-goggins: Yes, you are correct! Just updated to fix that typo. No need to guess which year someone was born in from `199` 🤣

BradySkylight (approved):
This looks great!  Nice work!

PR #162, opened 2023-02-02
Patient clustering by membership ratio

bamader:
This PR introduces a function meant for statistical testing on small validation sets. It finds clusters of matching patients within a given block of data, while allowing the user to specify the belongingness threshold a record must satisfy in order to qualify for membership in a cluster. This will be used to evaluate the impact of a variety of options--e.g. new record  must match 100% of patients currently under the considered representative patient, must match 75%, etc.


m-goggins (approved):
Looks good to me!

PR #163, opened 2023-02-03
Brady/date standardization

BradySkylight:
This PR includes all functions necessary to standardize the birthDate within a bundle or a single patient resource.  The base function accepts a string dob as well as a format string (indicating the format in which the date will be supplied).  This date format is defaulted to the FHIR standard of YYYY-MM-DD, as we anticipate this function being used in conjunction with a FHIR bundle or single patient resource.  However, this function is flexible to look for other delimiters as well as the position in which each of the date components exist.  If the format does not use the one specified by FHIR (STu3 and R4) then it places it in the YYYY-MM-DD format and stores it in the bundle/resource.  If the date is null or the format doesn't match up with the date or garbage data is supplied an exception is logged (for now, we can go back and change this later) as well as the dob being updated to "" in the bundle/resource.  This function also ensures that the date is valid and is not > today.


phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+                "day": format.lower().find("d"),
+            }
+            date_values = raw_dob.split(delim)
+
+            index = 0
+            for dictKey in dict(
+                sorted(positions.items(), key=lambda item: item[1])
+            ).keys():
+                date_dict[dictKey] = date_values[index]
+                index = index + 1
```
m-goggins, line 259: A lambda function in the wild! This code definitely works but I think we could make it slightly less dense by taking advantage of the fact that we have both the `raw_dob` and the `format` as well as the delimiter. I would create a list of the `raw_dob` and `format` and then iterate through both lists at the same time using `zip()` to create the `date_dict`. 

Something like:
```
raw_dob_values = raw_dob.split(delim)
format_values = format.replace("%","").lower().split(delim)

date_dict = {}
for format_value, dob_value in zip(format_values,raw_dob_values):
    date_dict[format_value] = dob_value
```

You could condense it further (if you still think it is understandable) to something like:
```
date_dict = {}
for format_value, dob_value in zip(format.replace("%","").lower().split(delim),raw_dob.split(delim)):
    date_dict[format_value] = dob_value
```

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+
+            if _validate_date(date_dict["year"], date_dict["month"], date_dict["day"]):
```
m-goggins, line 261: If you make the changes that I suggest above, you would need to change this to `_validate_date(date_dict["y"], date_dict["m"], date_dict["d"])1

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+                output = (
+                    date_dict["year"]
+                    + "-"
+                    + date_dict["month"]
+                    + "-"
+                    + date_dict["day"]
```
m-goggins, line 267: ditto changes here to "y","m","d"

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+                )
+            else:
+                # TODO:
+                # do we want to raise an exception here or any other action??
+                error_msg = f"Invalid birth date supplied: {raw_dob}"
```
m-goggins, line 272: Instead of an if/else statement with an error message, we should use a try/except statement that raises an error if it doesn't work. See examples of this in the `load_schema` function in phdi.tabulation.tables

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+            error_msg = (
+                f"Delimiter {delim} not found in birth date string supplied: {raw_dob}"
+            )
+    else:
+        # TODO:
+        # do we want to raise an exception here or any other action??
+        error_msg = f"Invalid birth date supplied: {raw_dob}"
+
+    if error_msg:
+        logging.exception(error_msg)
```
m-goggins, line 285: I'm not 100% but I've seen in other TODO statements elsewhere in our code that we are (or were at some point) moving away from logging and instead just using exceptions. I'm not sure if this still applies.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
     return nicknames_to_names
+
+
+def _validate_date(year: str, month: str, day: str) -> bool:
+    is_valid_date = True
+
+    try:
+        valid_date = datetime.datetime(int(year), int(month), int(day))
+        if valid_date > datetime.datetime.now():
+            is_valid_date = False
```
m-goggins, line 219: I would add a couple more checks: 

- Is the month value between 1 and 12?
- Is the days less than 32? We could add more sophisticated checks around 28, 29, 30, and 31 day checks for specific months if you think it's vital, but less than 32 should catch the egregious data entry errors for now.

m-goggins (changes requested):

phdi/harmonization/standardization.py
m-goggins: I actually think that maybe we should add the specific day checks since it's not too hard and the chances of us coming back to this any time soon are low.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+    The default being yyyy-mm-dd which is also the standard format for dates
+    that is utilized in FHIR STu3 and R4.  Then verify that the date is
+    a proper date and isn't a future date.  If the date is invalid then the
+    value for the dob will be nulled out and an error will be raised.
+
+    :param raw_dob: One birth date (dob) to standardize.
+    :param format: A python DateTime format used to parse the birthDate within
+        the Patient resource.  Default: `%Y-%m-%d` (also known as YYYY-MM-DD)
+    :return: Either a string that has the birth date in yyyy-mm-dd format
+        or a null value
```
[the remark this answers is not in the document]
BradySkylight: This is way more paired down, which I am all for.  Thank you and your suggestions make sense.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+    """
+    output = ""
+    error_msg = ""
+    if raw_dob and raw_dob.strip():
```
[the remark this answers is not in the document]
BradySkylight: Yes, I was checking to see if the raw_dob was empty or null or just whitespace.  When I was testing I was getting some errors.  I can remove these and see if the tests are still working etc...  Thanks!

phdi/harmonization/standardization.py
bamader: Marcelle I would agree with you here, but I think the `datetime.datetime` constructor actually handles that for us. I looked at the documentation for the constructor and it says that if an argument outside a valid range is given, it will throw a `ValueError` (https://docs.python.org/3/library/datetime.html#datetime-objects). So I believe the code is already checking for those things.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+    value for the dob will be nulled out and an error will be raised.
+
+    :param raw_dob: One birth date (dob) to standardize.
+    :param format: A python DateTime format used to parse the birthDate within
+        the Patient resource.  Default: `%Y-%m-%d` (also known as YYYY-MM-DD)
+    :return: Either a string that has the birth date in yyyy-mm-dd format
+        or a null value
+    """
+    output = ""
+    error_msg = ""
```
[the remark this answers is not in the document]
bamader: Ironically I think he does need them here not because he's initializing them, but because he's conditionally assigning them. `output` gets returned at the bottom of the function, but outside of this statement, the only time it gets set is line 257 which is in an `if` statement. A similar thing happens with `error_msg` only being set in some `else` clauses. Granted there's a lot of comments in this area of github so my ability to tell what's indented and what's not is basically shot lol, but I think he does need these as defaults to make tests happy (though if everything runs without them being default assigned up top, that works too).

phdi/harmonization/standardization.py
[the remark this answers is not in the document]
bamader: I definitely see how you could be getting an error if `raw_dob` was a `None` type. You can't call string manipulation functions on `None`. I'd be fine leaving the `if raw_dob` check in, but I agree with Marcelle there should be no need to check if `raw_dob.strip()` exists because most string operations shouldn't fail on an empty string, they should just do nothing. If you keep seeing errors when you take out that part of the check, I'm okay leaving it in.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +208,80 @@ def _build_nicknames_db():
...
+    if raw_dob and raw_dob.strip():
+        delim = detect(format.strip("%"))
```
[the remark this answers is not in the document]
bamader: ^^^ this is the way

phdi/harmonization/standardization.py
bamader: I'm normally a secret fan of lambda functions, but I agree with Marcelle here that this one--running some key sorting inside of a conditional for loop--is just a bit too big brain for me 🧠 . I really like the first snippet Marcelle gave where we explicitly create and zip the two objects. I'm not normally a fan of `zip` either (no fault of its own, it's an archaic dislike I have lol 😅 ), but in this case our guaranteed ordering of the date parts as well as the clarity of the code makes it a much better alternative to me.

phdi/harmonization/standardization.py
bamader: Bonus perk of try/except is that it _might_ get rid of the conditional default assignments to `output` and `error_msg` depending on how the code is structured (though not sure)

phdi/harmonization/standardization.py
bamader: We are! I was going to leave this comment like six times here, but we made an architectural decision to just raise exceptions and let the caller set up any kind of logging and reporting they wanted to. I think all you need to do within the appropriate try/except block is raise an exception (of the correct type--probably lots of `ValueError` here though this list will tell you what's most appropriate https://docs.python.org/3/library/exceptions.html#concrete-exceptions) that has meaningful information content.

poetry.lock
```diff
@@ -1057,94 +1065,1021 @@ testing = ["pytest (>=6)", "pytest-checkdocs (>=2.4)", "flake8 (<5)", "pytest-co
...
 
 [metadata.files]
-alabaster = []
-antlr4-python3-runtime = []
-attrs = []
-azure-core = []
-azure-identity = []
-azure-storage-blob = []
-babel = []
-black = []
```
bamader, line 1070: Just leaving the comment here, but _all_ of these metadata files changes are very concerning to me. The lock file shouldn't be changing with this much of an impact just by adding one package. By way of comparison, I ran `poetry add detect-delimiter`, grabbed the same version, and got only a very small change in the lockfile, with none of these metadata changes with it. I think something is up with the way your virtual environment is trying to push cached versions of these packages into the repo.

bamader (changes requested):
Agree with Marcelle on most of the pythonic comments, but of larger concern to me is the lockfile, which shouldn't be feeling changes of that magnitude for a single package addition. Something funky is definitely going on, and it's not the Funky Chicken coffee in my mug.

phdi/harmonization/standardization.py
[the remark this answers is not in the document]
BradySkylight: Yes, precisely...however, I may be able to subvert this by some of the other changes I'm making based upon all your feedback.  I'll see if these are needed.

poetry.lock
BradySkylight: This is concerning.  I only performed poetry add detect-delimiter and I don't see why it would need all those additional packages.  Any thoughts on how to untangle this?  I have a new machine and thereby a new VE, so this shouldn't be a problem going forward.  However, would you be willing to update this file in this branch with the detect_delimiter included @bamader ?

phdi/harmonization/standardization.py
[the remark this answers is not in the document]
BradySkylight: Done!

phdi/harmonization/standardization.py
m-goggins: Oh totally missed that. Whew! I was worried that we were trying to handle dates on our own.

poetry.lock
m-goggins: @BradySkylight I have had success with `poetry lock --no-update` after I added a package to the .toml file.

phdi/harmonization/standardization.py
BradySkylight: Do we want an exception to be raised that will stop processing a bundle if the dob is invalid?

phdi/harmonization/standardization.py
BradySkylight: This tremendously simplifies things.  Ugh... I was looking for something like zip() that would handle the scenario so did the best with the tools I had available.  It worked at least...but thank you so much because this is much cleaner and simple.

phdi/harmonization/standardization.py
BradySkylight: Correct.  This was the 'best practice', at least from what I could find about validating dates.  It's easier than writing our own code to validate the days and months etc...

bamader (changes requested):
The tests look fine to me, but I do still have lingering concerns around the lockfile changes. More generally though I think it might be worth talking about the scope of work entailed by this PR. I was under the impression we would be standardizing a broader inputs of dates into a uniform output as well as validating, but this PR does validation while stipulating that input dates already have a certain format, which seems to me to be less useful to us if we want to accept broader data.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +207,51 @@ def _build_nicknames_db():
                 for nickname in nicks.split(","):
                     nicknames_to_names[nickname] = name
     return nicknames_to_names
+
+
+def _validate_date(year: str, month: str, day: str) -> bool:
```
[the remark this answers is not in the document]
BradySkylight: Will do.  That makes sense to me.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +207,51 @@ def _build_nicknames_db():
...
+        valid_date = datetime.datetime(int(year), int(month), int(day))
+        if valid_date > datetime.datetime.now():
+            is_valid_date = False
+    except ValueError:
+        is_valid_date = False
+
+    return is_valid_date
+
+
+def standardize_birth_date(raw_dob: str, format: str = "%Y-%m-%d") -> str:
```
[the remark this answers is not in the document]
BradySkylight: That is the style of formatting used by the Python date, datetime functions though.  I guess we could 'add' them back in when actually applying the format?  I don't like that though...but also understand your point.  Any thoughts?

pyproject.toml
```diff
@@ -19,14 +19,15 @@ polling = "^0.3.2"
 phonenumbers = "^8.12.48"
 pycountry = "^22.3.5"
 PyYAML = "^6.0"
-pyarrow = "^8.0.0"
+pyarrow = "^11.0.0"
```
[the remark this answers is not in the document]
BradySkylight: I did this change manually to resolve the issue I was running into.

phdi/harmonization/standardization.py
```diff
@@ -205,3 +207,51 @@ def _build_nicknames_db():
...
+    if raw_dob is None or len(raw_dob) == 0:
+        raise ValueError("Date of Birth must be supplied!")
+
+    delim = detect(format.replace("%", ""))
+
+    # ensure that the delim in format exists in dob
+    if delim != detect(raw_dob):
+        raise ValueError(
+            f"Delimiter {delim} not found in birth date string supplied: {raw_dob}"
+        )
```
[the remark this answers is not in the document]
BradySkylight: I could definitely reform this to utilize regex (I love them btw...), but was just trying to get something pretty basic out.  My understanding was that this was specifically for DOB with the flexibility to be reutilized for other date formatting down the road.  Though, I must point out, that most dates are standardized already during the 'convert to fhir' BB/process prior to the standardizations.   Let me have a short convo with you and discuss this.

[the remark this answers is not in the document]
BradySkylight: The variants of the different order of date components should be handled however.

bamader (approved):
WE GOT THERE. This looks much better to me and delivers on what we talked about very well. I'm also now convinced that the lockfile changes here are real changes that are valid parts of adding `detect-delimiter` lol

m-goggins (approved):

PR #164, opened 2023-02-03
Improve error handling around smarty calls e.g. incorrect creds

gordonfarrell:
Addresses the lack of visibility into smarty errors, users should now see detailed error info for things such as bad smarty creds etc.

Wasn't 100% sure how to write a test for this, how the mock smarty server works wasn't very clear so may want to pair with someone in the next day or two. The current test that claims to test for bad creds (not blank creds) doesn't actually work, it's green no matter what you put in. Didn't want to delete it til I replaced it with something better though.


nickclyde (approved):
Looks great!

PR #165, opened 2023-02-06
Create scrambled synthetic record linkage data for testing algorithm

m-goggins:
This PR creates a couple of functions for creating a test set of synthetic data for record linkage. The new functions:

-  scramble names, dates of birth, and zip codes
- replace some first names with known associated nicknames
- introduce missing values

~~I am still figuring out some bugs with connecting to the SQLite DB but I wanted to give you all a chance to review the scrambling functions while I work on the bugs. Since we will be using the generated CSV directly, i.e., connecting to the DB is not necessary for moving forward with record linkage testing, I didn't want the test data to be a blocker for testing.~~


examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+
+    :param data: A DataFrame object.
+    :param num_copies: The number of duplicate rows to add for each existing row.
+    :return: A DataFrame object with duplicate rows.
+
+    """
+    data_with_copies = pd.DataFrame(np.repeat(data.values, num_copies, axis=0))
+    data_with_copies.columns = data.columns
+
+    return data_with_copies
```
bamader, line 141: Maybe we should make this function add a random number of copies between 1 and the number passed? That way we don't always get exactly 3 duplicates when we're trying to validate. Also, we could have the function take a float of what percentage of the data to add copies for. That might more readily mimic a population data set, since not every record will have duplicates, and not every record that is duplicated will have the same number of duplicates.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+    :return: DataFrame object that has been scrambled.
+
+    """
+
+    source_data["ZIP"] = source_data["ZIP"].astype(str).str.split(".").str[0]
+
+    # Introduce missingness
+    source_data = add_missing_values(source_data, missingness)
+
+    source_data_with_copies = add_copies(source_data, num_copies=3)
```
bamader, line 171: If we make the modifications I suggest above, I would go with a proportion like 40% for the number of records that get duplicated, and I'd do something like 7 for the maximum number of duplicates for a single record (the random mean will still be around 3 but it'll give us a bit more diversity).

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+
+    good_data = source_data_with_copies.sample(frac=0.7, random_state=seed)
+
+    # Scramble DOB in subsample
+    bad_dob = source_data_with_copies.sample(frac=0.05, random_state=seed)
+    bad_dob["BIRTHDATE"] = bad_dob["BIRTHDATE"].apply(lambda x: scramble_dob(x))
+    bad_dob["bad_dob"] = 1
```
bamader, line 178: Ahhhh here's the proportion. I'm totally fine to leave this here if you want but it could also make sense folded into the copies function directly.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
bamader: Edit: I see the actual proportion stuff is down below now

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+    # Assign nicknames in subsample
+    bad_name_nickname = source_data_with_copies.sample(frac=0.1, random_state=seed)
+    bad_name_nickname["bad_name_nickname"] = 1
+    bad_name_nickname["FIRST"] = bad_name_nickname["FIRST"].apply(
+        lambda x: swap_name_for_nickname(x)
+    )
+
+    # Scramble first names in subsample
+    bad_name_scramble_first = source_data_with_copies.sample(
+        frac=0.05, random_state=seed
```
bamader, line 194: Something that would be CLUTCH is to have a set of constants at the top of the file denoting the relative proportion of the data set to scramble for each type of scrambling. E.g.

```
PROPORTION_ZIP_SCRAMBLE = 0.10
PROPORTION_NICKNAME_SCRAMBLE = 0.24
PROPORTION_DOB_SCRAMBLE = 0.12
```

bamader (commented):
Code looks mostly good to me! Just a couple suggestions around some logic for the types of scrambling we're doing so that we get the most robust test set possible.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
bamader: Also, it occurs to me that we should probably guaranteed scramble the copies we create of a record. While it's likely that some will be randomly selected for scrambling manipulation, if the bulk of our data set's "true positives" are actually just outright duplicates, I feel like that's not enough proof that our algorithm could catch approximate/close records. I'd say we should do minimum 1 scramble (possibly/hopefully more) per duplicate.

examples/MPI-sample-data/build_sample_patient_sqlite_db.py
```diff
@@ -6,7 +6,7 @@
 
 df = pd.read_csv(
     # locally stord file of synthetic LAC synthea data
-    "~/20230106_LAC_10000_123_456/csv/patients.csv",
+    "C://Users/Marcelle/Desktop/resources/20230106_LAC_10000_123_456/csv/patients.csv",
```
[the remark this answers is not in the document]
m-goggins: Oops lol forgot to change that!

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+
+def scramble_name(name: str) -> str:
+    """
+    Scrambles a single name by randomly adding an existing letter, removing an
+    existing letter, or swapping two existing letters.
+
+    :param name: Name, as a string, to be scrambled.
+    :return: Scrambled name, as a string.
+
+    """
```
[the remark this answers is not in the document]
m-goggins: Yeah, that's a great call!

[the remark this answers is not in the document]
BradySkylight: Nice!

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
BradySkylight: I like this suggestion.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+    if method == "year":
+        scrambled_dob = dob[:2] + dob[3] + dob[2] + dob[4:]
+    # Swap the two digits of the month
+    elif method == "month":
+        scrambled_dob = dob[:5] + dob[6] + dob[5] + dob[7:]
+    # Swap the two digits of the day
+    elif method == "day":
+        scrambled_dob = dob[:-2] + dob[-1] + dob[-2]
+    # Add or subtract 1 from a DOB's year, month, or day value
+    elif method == "diff":
```
BradySkylight, line 39: I don't see "diff" in the list of methods randomly chosen above on line 27.  I do like this method though as it shifts the Date just slightly.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
```diff
@@ -0,0 +1,262 @@
...
+    existing letter, or swapping two existing letters.
+
+    :param name: Name, as a string, to be scrambled.
+    :return: Scrambled name, as a string.
+
+    """
+    # Randomly select scrambling method
+    method = random.choice(["add", "remove", "swap"])
+    # Randomly select non-first character in name
+    char = random.choice(range(1, len(name)))
```
BradySkylight, line 65: This is very similar to what my PLSQL code does.  Pretty cool to see it in Python.

BradySkylight (commented):
This is awesome!  I think Brandon had some good comments and suggestions.  I had only one major question.  This may be covered in your example already, but just wanted to make sure.  Are we storing or tagging the bad/scrambled records with some sort of ID that we can trace back to the original patient, but that won't also be used when running through our matching algorithm?  This would provide a mechanism to verify that our linking/deduplication is working....  I'm like 99.9% sure this is included but reading through the code I wasn't 100% sure.  Let me know if what I am saying is making sense or not.  I think this is more applicable to duplicates being created by this process.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
m-goggins: Yeah, definitely will add this, and I agree about increasing the number of scrambles/errors in the data that you mention in a comment above.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
m-goggins: Good call, I added it in.

m-goggins:
@BradySkylight yes, all the data (scrambled or not) has an associated ID. Check out `sample_record_linkage_data_scramble.csv` to see what the data look like at the end of the scrambling.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
m-goggins: I could have explained the copies better but basically I create 3 copies, then randomly choose a proportion to be scrambled, and then stack all the data (good, scrambled_name, scrambled_dob, etc.) so that you end up with some patients have 1 copy and others have way more copies. In the CSV right now, the number of copies looks like this:
Screenshot 2023-02-07 110735

BradySkylight:
> @BradySkylight yes, all the data (scrambled or not) has an associated ID. Check out `sample_record_linkage_data_scramble.csv` to see what the data look like at the end of the scrambling.

Thank you!  I saw your additional explanation and it makes sense. This all looks good!

BradySkylight (approved):
This is great!  Thanks for humoring my questions.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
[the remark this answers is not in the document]
bamader: Oh yeah totally keep those three! I was saying add `subset_scramble` as the 4th way a name could get mixed up (that's why I said it was 3:1 something else got picked). Randomly choose between the four types of scrambling, and then if subset is picked, randomly choose the length.

examples/Record-Linkage-sample-data/build_sample_record_linkage_data.py
bamader: WAIT LOL someone has 68 thousand duplicates of themselves? I'm still not quite sure I understand. I definitely missed the part in the code where you stack all the different facets, so my bad there, but are you saying there _are_ a random number of copies for each record to-be-duplicated?

bamader (approved):

PR #166, opened 2023-02-06
Add validation service

DanPaseltiner:
This PR adds a stubbed out initial version of our validation service.


containers/validation/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /code
+
+COPY ./requirements.txt /code/requirements.txt
+RUN pip install -r requirements.txt
+
+COPY ./app /code/app
+COPY ./description.md /code/description.md
+
+EXPOSE 8080
+CMD uvicorn app.main:app --host 0.0.0.0 --port 8080
```
BradySkylight, line 12: new container for validation... I like it!

containers/validation/app/main.py
```diff
@@ -0,0 +1,95 @@
...
+    The schema for response from the validate endpoint.
+    """
+
+    message_valid: bool = Field(
+        description="A true value indicates a valid message while false means that the "
+        "message was found to be invalid."
+    )
+    validation_results: dict = Field(
+        description="A JSON object containing detailing the validation result."
+    )
```
BradySkylight, line 48: I like this as it gives us a quick flag of yes/no did the validation succeed as well as detailed information.

containers/validation/app/main.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+# Request and and respone models
+class ValidateInput(BaseModel):
+    """
+    The schema for requests to the validate endpoint.
+    """
+
+    message_type: Literal["ecr"] = Field(
+        description="The type of message to be validated."
```
BradySkylight, line 32: I assume this is being hard coded for now as we are only working on validation for eCRs at the moment, but this will change depending upon the message type in the future.

containers/validation/app/main.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    validation_results: dict = Field(
+        description="A JSON object containing detailing the validation result."
+    )
+
+
+# Message type-specific validation
+def validate_ecr(message: str) -> ValidateResponse:
+    """
+    Validate and eCR message.
```
BradySkylight, line 54:
```suggestion
    Validate an eCR message.
```

BradySkylight (approved):
This is very awesome and I think it will work for what it is we are attempting to build.  Nice work.  I had a few comments, but nothing major and is just more nit-picky.  I could update these things as we work on the actual meat inside the BB.  Thanks for doing this Dan!

containers/validation/app/main.py
DanPaseltiner: `Literal` effectively defines an `enum`. As we add support for validating additional message types we can add to the list. Eventually I hope it looks something like `message_type: Literal["ecr", "elr", "vxu", "adt"]` once we nail down Hl7v2 validation.

PR #169, opened 2023-02-08
Implement validation of LAC model

bamader:
This PR gets us the rest of the way to replicating LAC's results. It introduces functions to run an individual linkage rules pass, a function to run LAC's suite, and some additional functions that help facilitate statistical evaluation.

### DIFF
--- a/examples/Record-Linkage-sample-data/evaluate_lac.py
+++ b/examples/Record-Linkage-sample-data/evaluate_lac.py
@@ -4,50 +4,95 @@ import pandas as pd
 
 from phdi.linkage import lac_validation_linkage, score_linkage_vs_truth
 from phdi.linkage.link import phdi_linkage_algorithm
+from typing import Union
 
 DATA_SIZE = 50000
 warnings.simplefilter(action="ignore", category=FutureWarning)
 
 
 def determine_true_matches_in_pd_dataset(data: pd.DataFrame):
-    pass
+    print("-------Identifying True Matches for Evaluation-------")
+    true_matches = {}
+    tuple_data = tuple(data.groupby("Id"))
+    for master_patient, sub_df in tuple_data:
+        sorted_idx = sorted(sub_df.index)
+        for idx in range(len(sorted_idx)):
+            r_idx = sorted_idx[idx]
+            if r_idx not in true_matches:
+                true_matches[r_idx] = set()
+            for i in range(idx + 1, len(sorted_idx)):
+                true_matches[r_idx].add(sorted_idx[i])
+    data["ID"] = data.index
+    data = data.drop(columns=["Id"])
+    return data, true_matches
+
+
+def identify_missed_matches(
+    found_matches: dict[Union[int, str], set],
+    true_matches: dict[Union[int, str], set],
+):
+    missed_matches = {}
+    for root_record in true_matches:
+        if root_record in found_matches:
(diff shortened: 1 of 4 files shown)


examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -0,0 +1,231 @@
...
+    return data
+
+
+def add_metaphone_columns_to_data(data: pd.DataFrame):
+    data["DM_FIRST"] = data["FIRST"].apply(lambda x: double_metaphone_string(x)[0])
+    data["DM_LAST"] = data["LAST"].apply(lambda x: double_metaphone_string(x)[0])
+    return data
+
+
+def identify_missed_matches(
```
m-goggins, line 152: It occurred to me while reviewing this function that maybe we shouldn't be including bad zip codes as part of this analysis. We will be geocoding the data before they are linked so the zip codes should be standardized.

m-goggins (commented):
The flow of the code looks pretty good to me. There is A LOT going on in this PR and it probably would have been easier to review in smaller chunks, e.g., identify true matches, compile matches across runs, statistical evals. To ensure that I understand what is going on, I'm going to put some time on our calendars for Wednesday.

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -0,0 +1,231 @@
...
+        "Misses on records with a missing ADDRESS:",
+        missing_address_misses / len(missed_df),
+    )
+    print(
+        "Misses on records with no scrambling:", non_scrambled_misses / len(missed_df)
+    )
+
+
+data = pd.read_csv("./sample_record_linkage_data_scrambled.csv", dtype="string")
+data = data.loc[:DATA_SIZE]
```
m-goggins, line 224: I would change this to `data = pd.read_csv("./sample_record_linkage_data_scrambled.csv", dtype="string", nrows=DATA_SIZE)` to avoid loading the full file every time.

m-goggins:
This is looking good to me after our meeting this morning! I made a small suggestion and left 1 question about the expected data type (but it could be pushed off until later).

m-goggins (approved):
.

BradySkylight (approved):
This is pretty awesome!  Lots to review, but I like how its broken out and organized.  Excited to see the data from this.  LGTM

examples/Record-Linkage-sample-data/evaluate_lac.py
bamader: We resolved this in the Wednesday discussion, documenting here: any record for which the geocoding result resolves will have the updated address and zip, but any record where the geocoder breaks will just use the existing data (and a bad zip could be one way it breaks), so we're going to leave this in at its small proportion.

phdi/linkage/link.py
```diff
@@ -79,12 +225,117 @@ def match_within_block(
...
+    matching_rule: Callable,
+    cluster_ratio: Union[float, None] = None,
+    **kwargs,
+) -> dict:
+    """
+    Performs a partial run of a linkage algorithm using a single rule.
+    Each rule in an algorithm is associated with its own pass through the
+    data.
+
+    :param data: A pandas dataframe of records to link.
```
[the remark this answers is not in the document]
bamader: Oh good call. I'll add a todo for this.

PR #170, opened 2023-02-09
Implement blocking function for data from DB

m-goggins:
This PR returns a single block of data (list of lists) for a given input. This function operates very similarly to the `block_parquet_data` function with the exception that the new function accepts a dictionary containing information from an incoming record (in theory) for the block of data that we need to retrieve for that record. In `block_parquet_data` the input required a list of columns to block on, e.g., ZIP, and returned _all_ blocks for every ZIP. The new function, `block` assumes you only want to return one block for a particular incoming record. 

As an example, if a new record comes in where ZIP is 90210, the `block` function returns only the records where ZIP is also 90210. I did this in preparation for streaming and thinking that holding all blocks in memory would be potentially risky but if we go with batching, I would be happy to re-arrange such that the function returns all blocks applicable to the incoming batch of records rather than a single record. Let me know what you think!

### DIFF
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -272,7 +272,7 @@ def block(db_name: str, table_name: str, block_data: Dict) -> List[list]:
     """
     conn = sqlite3.connect(db_name)
     cursor = conn.cursor()
-    conn.row_factory = lambda cursor, row: [i for i in row]
+    cursor.row_factory = lambda c, row: [i for i in row]
 
     query = _generate_block_query(table_name, block_data)  # Generate SQL query
     cursor.execute(query)  # Execute query
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -266,9 +266,8 @@ def block(db_name: str, table_name: str, block_data: Dict) -> List[list]:
     :param table_name: Table name.
     :param block_data: Dictionary containing key value pairs for the column name you for
       blocking and the data for the incoming record, e.g., ["ZIP"]: "90210".
-    :return: A list of records to that are within in the block, e.g., records that all
+    :return: A list of records that are within in the block, e.g., records that all
       have 90210 as their ZIP.
-
 
     """
     conn = sqlite3.connect(db_name)
@@ -297,7 +296,10 @@ def _generate_block_query(table_name: str, block_data: Dict) -> str:
     """
     query_stub = f"SELECT * FROM {table_name} WHERE "
     block_query = " AND ".join(
-        [key + f" = '{str(value)}'" for key, value in block_data.items()]
+        [
+            key + f" = '{value}'" if type(value) == str else (key + f" = {value}")
+            for key, value in block_data.items()
+        ]
     )
     query = query_stub + block_query
     return query
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -2,6 +2,7 @@ import hashlib
 import pandas as pd
 from phdi.harmonization.utils import compare_strings
 from typing import List, Callable, Dict
+import sqlite3
 
 
 def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
@@ -25,7 +26,7 @@ def match_within_block(
     block: List[List],
     feature_funcs: dict[int, Callable],
     match_eval: Callable,
-    **kwargs
+    **kwargs,
 ) -> List[tuple]:
     """
     Performs matching on all candidate pairs of records within a given block
@@ -84,7 +85,7 @@ def _match_within_block_cluster_ratio(
     cluster_ratio: float,
     feature_funcs: dict[int, Callable],
     match_eval: Callable,
-    **kwargs
+    **kwargs,
 ) -> List[set]:
     """
     A matching function for statistically testing the impact of membership
@@ -141,7 +142,7 @@ def _eval_record_in_cluster(
     cluster_ratio: float,
     feature_funcs: dict[int, Callable],
     match_eval: Callable,
-    **kwargs
+    **kwargs,
 ):
     """
     A helper function used to evaluate whether a given incoming record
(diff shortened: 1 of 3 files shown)


phdi/linkage/link.py
```diff
@@ -252,3 +253,53 @@ def block_parquet_data(path: str, blocks: List) -> Dict:
...
+    :param table_name: Table name.
+    :param block_data: Dictionary containing key value pairs for the column name you for
+      blocking and the data for the incoming record, e.g., ["ZIP"]: "90210".
+    :return: A list of records that are within in the block, e.g., records that all
+      have 90210 as their ZIP.
+
+    """
+    conn = sqlite3.connect(db_name)
+    cursor = conn.cursor()
+    cursor.row_factory = lambda c, row: [i for i in row]
```
BradySkylight, line 275: Lambda spotted!!  This all makes sense and is straight forward for db connection, cursor, and queries though.  Nice!

phdi/linkage/link.py
```diff
@@ -252,3 +253,53 @@ def block_parquet_data(path: str, blocks: List) -> Dict:
...
+
+    """
+    query_stub = f"SELECT * FROM {table_name} WHERE "
+    block_query = " AND ".join(
+        [
+            key + f" = '{value}'" if type(value) == str else (key + f" = {value}")
+            for key, value in block_data.items()
+        ]
+    )
+    query = query_stub + block_query
```
BradySkylight, line 304: A couple of things, though I know this is an initial function for blocking.  Is there some kind of criteria prior to the AND between the query-stub and block_query?  And the second thing, and we can add this later, is just a validation on table name and any column values.  I'm sure we are fine now, but may be good to have later on.

BradySkylight (approved):
Looks good.  Just had a couple of questions that were more of 'keep this in mind' type of comments as I know this is an initial pass for our SQL Lite DB that won't be our final product for the MPI.  Excellent work!

phdi/linkage/link.py
m-goggins: Hm, I'm not sure what you mean by criteria prior to the query_stub and block_query. For your second question, totally agree  about validating. We should eventually have another function to validate table name and columns.

BradySkylight: Sorry, my first question was regarding what I am seeing in two string variables and when they are put together it would seem to me that the query would look like:  "WHERE AND ...."   with the key and value pairs giving the criteria for the query.  The concern was just the "WHERE AND" without any criteria between the two key words.

phdi/linkage/link.py
```diff
@@ -252,3 +253,53 @@ def block_parquet_data(path: str, blocks: List) -> Dict:
...
+    """
+    Returns a list of lists containing records from the database that match on the
+    incoming record's block values. If blocking on 'ZIP' and the incoming record's zip
+    code is '90210', the resulting block of data would contain records that all have the
+    same zip code of 90210.
+
+    :param db_name: Database name.
+    :param table_name: Table name.
+    :param block_data: Dictionary containing key value pairs for the column name you for
+      blocking and the data for the incoming record, e.g., ["ZIP"]: "90210".
```
bamader, line 268: nit: typo at the end of line 267, I think you want to delete `you`

phdi/linkage/link.py
```diff
@@ -252,3 +253,53 @@ def block_parquet_data(path: str, blocks: List) -> Dict:
...
+    :return: A list of records that are within in the block, e.g., records that all
```
bamader, line 269: nit: typo, delete either `in` or `within`

phdi/linkage/link.py
```diff
@@ -252,3 +253,53 @@ def block_parquet_data(path: str, blocks: List) -> Dict:
...
+    block = cursor.fetchall()  # Fetch data from query
+    conn.commit()
+    conn.close()
+
+    return block
+
+
+def _generate_block_query(table_name: str, block_data: Dict) -> str:
+    """
+    Generates a query for selecting block of data from `table_name` per the block_data
```
bamader, line 288: nit: `selecting a block of data`

phdi/linkage/link.py
bamader: So this will never actually happen because she's not just concatenating with `" AND [some stuff]`, she's running a string `join` on them. That means that the string `" AND "` is inserted as a delimiter into the list of elements passed to the `.join()` function, which in this case are formatted pairs of keys and values. So under the hood the variable actually looks like this: `block_query = key_1 + f" = '{value_1}'" + " AND " + "key_2 + f" = '{value_2}'" + " AND " + ...`

bamader: The only possible issue here is if block_data has no items, the query will generate with a hanging ` WHERE ... ` and no condition, so I'd like to see a check at the top of the `block` function to verify that `block_data` has at least one key-value pair.

tests/linkage/test_linkage.py
```diff
@@ -177,3 +181,54 @@ def test_block_parquet_data():
...
+
+    assert (
+        type(list(block_data.values())[0]) != str
+        and "'" not in correct_query.split("= ")[1]
+    )  # Non-string types should not be enclosed in quotes
+
+
+def test_blocking_data():
+    db_name = (
+        pathlib.Path(__file__).parent.parent.parent
```
bamader, line 213: Once we have the check that block_data is non-empty, let's also add a test case verifying that the function correctly handles that.

bamader (changes requested):
Looks good! Aside from a few typos, the only request I have is to add a sanity check at the top of the `block` function along the lines of `if len(block_data) == 0 raise exception` and then add a test for it.

phdi/linkage/link.py
BradySkylight: Great catch bamader and thank you for the explanation!  That makes more sense looking at it now.

tests/linkage/test_linkage.py
m-goggins: Good catch! I just added that in as well as a test.

bamader (approved):
Looks good! Thanks for adding that extra check.

PR #171, opened 2023-02-10
Expand validation service

DanPaseltiner:
This PR expands the stubbed functionality in the validation service to include ELR and VXU. Currently so actually validation is actually done, but as we develop methods for validation eCR, ELR, and VXU can quickly start using them here.

### DIFF
--- a/containers/validation/app/main.py
+++ b/containers/validation/app/main.py
@@ -28,7 +28,7 @@ class ValidateInput(BaseModel):
     The schema for requests to the validate endpoint.
     """
 
-    message_type: Literal["ecr"] = Field(
+    message_type: Literal["ecr", "elr", "vxu"] = Field(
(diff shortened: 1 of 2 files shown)


containers/validation/app/main.py
```diff
@@ -51,7 +51,7 @@ class ValidateResponse(BaseModel):
 # Message type-specific validation
 def validate_ecr(message: str) -> ValidateResponse:
     """
-    Validate an eCR message.
+    Validate an XML-formatted eCR message.
```
BradySkylight, line 54: Should we include something about the fact that the eCR message is a CDA?  and/or call out eICR to make it really clear?  No big deal and totally nit-picky.

BradySkylight (approved):
Looks good.  Very nice work!

PR #172, opened 2023-02-14
Xpath ecr validator

KennethSkylight:
Description

Adding a sample validator for attributes for eCR records. It takes in a "config" file and maps them into a function to check if the record has the field then does a requirements check and a regex check.


phdi/validator/test.xml
```diff
@@ -0,0 +1,10 @@
+<foo>
```
BradySkylight, line 1: Let's keep the folder name consistent with the container.  Instead of 'validator' can we change this to 'validation'  the same for the python file validation.py instead of validate.py.

phdi/validator/validate.py
```diff
@@ -0,0 +1,171 @@
...
+        #     "textRequired": True,
+        #     "regEx": "/[0-9]{5}(?:-[0-9]{4})?/",
+        # },
+        {
+            "fieldName": "Provider ID",
+            "cdaPath": "//hl7:ClinicalDocument/hl7:componentOf/hl7:encompassingEncounter/hl7:responsibleParty/hl7:assignedEntity/hl7:id",
+            "attributes": [{"attributeName": "extension"}, {"attributeName": "root"}],
+        },
+    ]
+}
```
BradySkylight, line 101: This should be in a config file somewhere and then we should have only the portions we need for processing passed as within an input parameter probably as a dict.

phdi/validator/validate.py
```diff
@@ -0,0 +1,171 @@
...
+
+namespaces = {
+    "hl7": "urn:hl7-org:v3",
+    "xsi": "http://www.w3.org/2005/Atom",
+    "cda": "urn:hl7-org:v3",
+    "sdtc": "urn:hl7-org:sdtc",
+    "voc": "http://www.lantanagroup.com/voc",
+}
```
BradySkylight, line 109: Could we make these a global variable or perhaps as a Literal List/ENUM?

sample_config.yaml
```diff
@@ -0,0 +1,86 @@
...
+  #     - attributeName: example
+  #       regEx:
+  #     - attributeName: example
+  #       regEx:
+  #   textRequired: false
+  #   requiredIf:
+  #     otherFieldName:
+  #     otherFieldValue:
+  #     otherFieldAttribute:
+  #     otherFieldAttributeValue:
```
BradySkylight, line 86: Is this a sample or the LAC version or both?

phdi/validation/validate.py
```diff
@@ -0,0 +1,174 @@
...
+import re
+
+namespaces = {
+    "hl7": "urn:hl7-org:v3",
+    "xsi": "http://www.w3.org/2005/Atom",
+    "cda": "urn:hl7-org:v3",
+    "sdtc": "urn:hl7-org:sdtc",
+    "voc": "http://www.lantanagroup.com/voc",
+}
+
```
BradySkylight, line 11: This great!

BradySkylight (approved):
Good for a first pass.  I wish we were farther along, but we have a good list of next steps to knock out.  Thanks for the work on this!!

PR #176, opened 2023-02-16
Integrate Double Metaphone into Name Standardization

BradySkylight:
This PR essentially adds Double Metaphone encoding into the Name Standardization SDK Building block, by default, but provides the ability to opt out of performing the DM for patient names.  This will ensure that after the name standardization is complete the DM will then be added to the patient.name resource.element as an extension to ensure this data is stored with the patient name in the persistence layer.  This makes sure our record linkage process can utilize the DM data.

### DIFF
--- a/phdi/fhir/harmonization/standardization.py
+++ b/phdi/fhir/harmonization/standardization.py
@@ -126,6 +126,8 @@ def standardize_names(
         resource = _standardize_names_in_resource(
             resource, trim, case, remove_numbers, overwrite
         )
+        # Add double metaphone data to patient name after standardizing it
+        double_metaphone_patient(patient=resource, dmeta=None, overwrite=overwrite)
 
     if "entry" not in data:
         return bundle.get("entry", [{}])[0].get("resource", {})
--- a/phdi/fhir/harmonization/standardization.py
+++ b/phdi/fhir/harmonization/standardization.py
@@ -101,6 +101,8 @@ def standardize_names(
     Standardizes all names contained in a given FHIR bundle or a FHIR resource. The
     default standardization behavior is our defined non-numeric, space-trimming, full
     capitalization standardization, but other modes may be specified.
+    Names are also assigned a double metaphone encoding after standardization is
+    complete.
 
     :param data: A FHIR-formatted JSON dict.
     :param trim: Whether leading/trailing whitespace should be removed. Default: `True`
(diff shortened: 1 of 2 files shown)


DanPaseltiner (changes requested):
Thanks for putting this together @Brady. Couple comments below about minor things in the code. On a higher design level I'm torn about whether we want to bake DM into name standardization or not. On the one hand I understand that getting the DM encodings is effectively one way of standardizing name, but there are likely use cases where it is not needed and it is also computationally expensive, at least relatively speaking. Another way to go is to leave these as separate functions exposed on separate endpoints, that a user can combine if desired.

m-goggins:
This is looking good to me! Sorry it took so long for me to review; I confused this PR with the cloud implementation one. Two small things:

1. I think we should change `incl_encoding` to something more specific like `include_double_metaphone`  since that's the only encoding option we have right now.

2. I would suggest that we change the default for the DM to `False`. We don't expect to use DM in our record linkage models and it doesn't seem like it would be a standard feature for others (unless future UX research proves me wrong!). Adding the DM can be computationally expensive, especially if not necessary.

BradySkylight:
> Thanks for putting this together @brady. A couple comments below about minor things in the code. On a higher design level I'm torn about whether we want to bake DM into name standardization or not. On the one hand I understand that getting the DM encodings is effectively one way of standardizing name, but there are likely use cases where it is not needed and it is also computationally expensive, at least relatively speaking. Another way to go is to leave these as separate functions exposed on separate endpoints, that a user can combine if desired. I'm leaning towards keeping them separate because there doesn't seem to be a clear choice for the default value of `incl_encoding`. For our purposes in the starter kit we will always want to compute DM, but given the computational expense of this operation for a general use case I'd argue we should have this function default to not computing DM.

Yes, we had conversations back and forth about this.  We could do both, but, if we don't provide DM encoding, we will need to account for the fact that this data may be missing in our RL algorithm.

DanPaseltiner:
> > Thanks for putting this together @brady. A couple comments below about minor things in the code. On a higher design level I'm torn about whether we want to bake DM into name standardization or not. On the one hand I understand that getting the DM encodings is effectively one way of standardizing name, but there are likely use cases where it is not needed and it is also computationally expensive, at least relatively speaking. Another way to go is to leave these as separate functions exposed on separate endpoints, that a user can combine if desired. I'm leaning towards keeping them separate because there doesn't seem to be a clear choice for the default value of `incl_encoding`. For our purposes in the starter kit we will always want to compute DM, but given the computational expense of this operation for a general use case I'd argue we should have this function default to not computing DM.
> 
> Yes, we had conversations back and forth about this. We could do both, but, if we don't provide DM encoding, we will need to account for the fact that this data may be missing in our RL algorithm.

It sounds like we aren't currently using DM for record linkage and it could just be an additional step in the pipeline.

BradySkylight:
Based upon the fact that DM isn't used in our RL algorithm, at this time.  I'm going to leave this as as separate BB endpoint and just close this PR.

PR #177, opened 2023-02-22
Add workflow to run tests then tag a new release if tests are green

gordonfarrell:
Does what it says on the tin!

This is a workflow triggered manually from the actions tab on github. A user must input a custom string version number for it to run.

Part 1 of the full auto release process.


DanPaseltiner (approved):
LGTM!

PR #226, opened 2023-02-28
Add error type

KennethSkylight:
# Overview

This branch adds the error types, which allows for sorting of messages when validating the ecr message. The config file must now include an `errorType` definition with the new `validate_config` function in the validation python file.


tests/validation/test_validation.py
```diff
@@ -66,4 +76,18 @@ def test_validate_bad():
         error_types=["error", "warn", "info"],
     )
 
+    print(result)
```
nickbristow, line 79: we can remove these prints, but its good otherwise

nickbristow (approved):

PR #233, opened 2023-03-02
Add additional tests for Validation

nickbristow:
Adds robust testing for validation helper functions

### DIFF
--- a/phdi/validation/validation.py
+++ b/phdi/validation/validation.py
@@ -34,7 +34,7 @@ def validate_ecr(ecr_message: str, config: dict, error_types: str) -> dict:
             continue
 
         for xml_element in matched_xml_elements:
-            error_messages += _validate_attribute(field, xml_element)
+            error_messages += _validate_attribute(xml_element, field)
             error_messages += _validate_text(field, xml_element)
 
     if error_messages:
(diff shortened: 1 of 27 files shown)


phdi/validation/validation.py
```diff
@@ -188,24 +185,24 @@ def _validate_text(field, node):
...
+                + config_field.get("fieldName")
                 + " does not match regEx: "
-                + field.get("regEx")
+                + config_field.get("regEx")
             ]
         else:
             return []
     else:
-        if text is not None:
+        if text is not None and text != "":
```
KennethSkylight, line 205: I believe if you just do `if not text` that would work

Both empty string and None are falsey

KennethSkylight (commented):
Just one question about the conditional check

phdi/validation/validation.py
nickbristow: tests fail when i do that. not sure why

BradySkylight: I've had a similar issue when testing text fields as well.  This was just recently though.

BradySkylight (approved):
These tests look much better and cover all the helper functions. Excellent job Nick!  Thanks for tackling this!

Issue #217, opened 2023-02-28
Create m-probability function


bamader:

bamader:
In order to implement probabilistic correction to a deterministic model, we need to know the various "m-probabilities" of the fields in the data set. The **m-probability** is defined as the probability that two field identifiers match, conditioned on / given that the two records in question truly are a match.

**Acceptance Criteria:**
* A function `determine_m_probs` is written which accepts as input: a pandas dataframe in which rows are individual patient records and columns are fields to evaluate; a set of columns to evaluate probabilities for; a dictionary holding the true matches identified in the data set
* The created function computes *for each field in the given column set* the probability that two records match on that field, given that the candidate pair of records is one of the "true match pairs" specified in the parameter dictionary of true matches (i.e. count the number of true match pairs that agree on a column, divide it by the number of true matches there are total)
* The function returns a dictionary whose keys are column names and whose values are the m-probabilities

emmastephenson:
Fixed with #239

PR #239, opened 2023-03-03
Add m and u probabilities, log odds

bamader:
This PR adds functionality for calculating m- and u- probabilities, log odds scores, and reading/writing these structures from files (since they only need to be pre-computed once). Probabilistic linkage isn't the focus, I know, but in the interim before ticketing orchestration and cloud implementation work begins, I'm pretty sure I can get probabilistic correction done.


phdi/linkage/link.py
```diff
@@ -28,6 +32,151 @@ def block_data(data: pd.DataFrame, blocks: List) -> dict:
...
+    """
+    For a given set of patient records, calculate the per-field
+    m-probability. The m-probability for field X is defined as the
+    probability that a pair of records A and B have the same value in
+    X, given that A and B are a true matching pair. This function
+    incorporates LaPlacian Smoothing to account for unseen data and
+    to resolve future logarithms against 0.
+
+    :param data: A pandas dataframe of patient records to compute
+      probabilities for.
```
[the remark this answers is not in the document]
m-goggins: Maybe we can add a TODO item here for now.

phdi/linkage/link.py
```diff
@@ -28,6 +32,151 @@ def block_data(data: pd.DataFrame, blocks: List) -> dict:
...
+    u-probability. The u-probability for field X is defined as the
+    probability that a pair of records A and B have the same value in
+    X, given that A and B are not a true matching pair. This function
+    incorporates LaPlacian Smoothing to account for unseen data and
+    to handle future logarithms against 0.
+
+    Note: This function can be slow to compute for large data sets.
+    It is recommended to pass only a representative subsample of the
+    data to the function, even if sampling is used within (the sample
+    operation can be time-consuming).
```
m-goggins, line 131:
```suggestion
    Note: This function can be slow to compute for large data sets.
    It is recommended to pass only a representative subsample of the
    data to the function (we recommend no more than 10k records for 
    performance), even if the sample operation is used.
```

m-goggins: Not sure if 10k is appropriate but might be good to give a ballpark figure

phdi/linkage/link.py
```diff
@@ -28,6 +32,151 @@ def block_data(data: pd.DataFrame, blocks: List) -> dict:
...
+
+    u_probs = {c: 1.0 for c in cols}
+
+    # Want only the pairs of candidates that aren't true matches
+    base_pairs = list(combinations(data.index, 2))
+    neg_pairs = [
+        x
+        for x in base_pairs
+        if x[0] not in true_matches or x[1] not in true_matches[x[0]]
+    ]
```
m-goggins, line 159: Does this rely on the indices in the data and true matches being the same?

phdi/linkage/link.py
```diff
@@ -168,6 +317,34 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+      path.
+    :raises JSONDecodeError: If the file cannot be read as valid JSON.
+    """
+    try:
+        with open(path, "r") as file:
+            prob_dict = json.load(file)
+        return prob_dict
+    except FileNotFoundError:
+        raise FileNotFoundError(
+            "The specified file does not exist at the path provided."
```
m-goggins, line 340:
```suggestion
            f"The specified file does not exist at {path}"
```

m-goggins (commented):
Mostly small edits, suggestions for future iterations (namely removing pandas dependency), and a question about how we're using indices when finding the non-matches.

tests/linkage/test_linkage.py
```diff
@@ -454,3 +461,119 @@ def test_blocking_data():
...
+    )
+    assert u_probs == true_probs
+
+    loaded_probs = load_json_probs("u.json")
+    assert loaded_probs == true_probs
+
+    os.remove("./u.json")
+
+
+def test_read_write_log_odds():
```
emmastephenson, line 545: Another Python standards question - is it normal to write tests for unexpected inputs, like a `true_matches` dict that doesn't match the shape we expect? Is that something we'd ever anticipate happening?
(Again Java brain at play here - coming from a strongly typed language I find myself mistrusting basically every variable lol)

emmastephenson (commented):
Thanks for this Brandon! I'm so happy to get the chance to look at code again 😆 Let me know if it would be easier to talk through any comments offline, I know I asked a couple of broad questions :)

phdi/linkage/link.py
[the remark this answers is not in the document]
bamader: Agreed. Pandas is helpful in a lot of the prework and modeling phases but we do want to move more light-weight eventually. Will add a TODO.

phdi/linkage/link.py
bamader: 10k is a little low but I can definitely ballpark it (I used a sample of 25k out of a combinatorial list generated from 25k records)

phdi/linkage/link.py
bamader: It does, yeah. It's a tricky situation because while in the wild that's not really guaranteed (or even likely) to be the case, in literally _any_ scenario where we're pre-computing statistics (like m- and u-probs) or doing operations involving determining an _a priori_ set of true matches, we format and re-index the data such that index directly corresponds to ID. 

Example: when running the pipeline, we won't ever be running a function to `determine_true_matches`, we won't be calculating statistics that rely on knowing true matching, and we won't be estimating any sort of m- or u-probabilities, because that all relies on prior knowledge that's not available to the pipeline during daily use (nor should it be). So these kinds of dependencies won't be an issue. On the other hand, when we're doing preprocessing or probability sampling or something, because we always take subsets or representative samples of the data, we end up needing to re-scale/re-index anyway, which makes this dependency a non-issue as well.

Maybe we could add some more documentation around when these types of functions are useful? I'm also open to changing things to work based on storing record ID instead of index if that seems safer.

bamader: On second pass this might be something we're able to defer until the time when we move away from pandas. Because we'll probably wind up as a list of lists at that point (on top of the need to never know ID/indices outside of pre-computing scenarios), it might not be worth going too far down a rabbit hole of swapping out the index for a specific DF-style ID column. Thoughts?

phdi/linkage/link.py
```diff
@@ -28,6 +32,151 @@ def block_data(data: pd.DataFrame, blocks: List) -> dict:
...
+        log_odds[k] = log(m_probs[k]) - log(u_probs[k])
+    print(log_odds)
+    if file_to_write is not None:
+        with open(file_to_write, "w") as out:
+            out.write(json.dumps(log_odds))
+    return log_odds
+
+
+def calculate_m_probs(
+    data: pd.DataFrame,
```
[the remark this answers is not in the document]
bamader: You didn't know it but that's actually a somewhat thorny question 😅 . In the past, we attempted to standardize our naming conventions based on R's tidyverse (if you've never heard of it, it's basically a unified collection of the most common data analysis packages used for R, written in such a way as to emphasize one particular set of user-design principles). Bryan was really fond of the way they organized and named things, and so we tried to port those same philosophies and design perspectives into PHDI. One of their big parts was that the characteristics/relationships between "data" parameters moved from very general to very specific, e.g. the primary "data set" was always just called `data` as a function parameter and was then modified by other parameters that added context.

With all that established, we've been sort of wishy-washy on carrying that through in our repo. I think my personal preference is very much trending towards very specific parameter names in all cases, but most of the time people opened PRs with general params they'd get nit comments around naming. Since we created that "design/naming standard" all those months ago, there's been no real re-visitation other than the design philosophies we spell out in our contributing documentation. 

All that is to say, nope there's nothing pythonic you're missing here! It's just a decision we should probably make or reaffirm to either stick with general names or move over to specific names.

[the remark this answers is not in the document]
emmastephenson: This is great context, thank you! Really appreciate it. Sounds like something to put on the agenda for an office hours sync.

phdi/linkage/link.py
```diff
@@ -28,6 +32,151 @@ def block_data(data: pd.DataFrame, blocks: List) -> dict:
...
+      write the probabilities in JSON format. Default is None.
+    """
+    if cols is None:
+        cols = data.columns
+
+    u_probs = {c: 1.0 for c in cols}
+
+    # Want only the pairs of candidates that aren't true matches
+    base_pairs = list(combinations(data.index, 2))
+    neg_pairs = [
```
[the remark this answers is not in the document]
bamader: Yeah this is almost exactly it! Step 1 is correct (generate all combinations based on an index) and the idea of Steps 2 and 3 are correct, there's just a minor technical (honestly unimportant) difference between what you said and what happens: the dictionary of `true_matches` is structured in such a way that the "lower numbered" ID of any given candidate record pair is a key, and the values are sets of "higher numbered" record IDs that the key is a true match to. Example: if our data set had four elements, 0 through 3, and the first 3 were all matches to each other but the fourth didn't match anything, our dictionary would look like this:

```
{
  0: set(1,2)
  1: set(2)
  2: set()
  3: set()
}
```

This might look weird (that 2 is listed as having no matches), but it's actually an important set up for statistics checking. Introducing "ordering" based on ID means that at no point in our computations do we have to worry about double counting any candidate pair. If the value set for the key `2` above had {0,1} in it, we will have counted the pairs (0,2) and (1,2) twice, because there is no difference between saying record 0 matches record 2 and record 2 matches record 0. So this "one-sided storage" means we never have to do any kind of divide by 2 fussiness at any point. Now, that impacts what you've written because when it comes to the "second" element of the candidate pair (i.e. the "higher-numbered" ID one), we never have to check if that's a key in the dictionary, because we know if it is a match to the lower numbered ID record, it will always be captured in the lower numbered set. So, what we actually check is whether the lower numbered record appears in the dictionary at all (i.e. does anything at all match to it), and if it does, does the higher numbered record appear in the value set for that lower numbered key (which indicates a match). Like I said, very small and probably pedantic difference, but because we use this true matches dictionary for a lot of precomputing I thought it was important to walk through the structure of it. Let me know if that doesn't make sense, happy to talk more about it!

phdi/linkage/link.py
```diff
@@ -28,6 +32,151 @@ def block_data(data: pd.DataFrame, blocks: List) -> dict:
...
+            if data[c].iloc[index[0]] == data[c].iloc[index[1]]:
+                u_probs[c] += 1.0
+
+    for c in cols:
+        if n_samples is not None and n_samples < len(neg_pairs):
+            u_probs[c] = u_probs[c] / (n_samples + 1.0)
+        else:
+            u_probs[c] = u_probs[c] / (len(neg_pairs) + 1.0)
+
+    if file_to_write is not None:
```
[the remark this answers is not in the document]
bamader: Why yes it could, will add that in

tests/linkage/test_linkage.py
```diff
@@ -454,3 +461,119 @@ def test_blocking_data():
     with pytest.raises(ValueError) as e:
         block_data_from_db(db_name, table_name, block_data)
     assert "`block_data` cannot be empty." in str(e.value)
+
+
+def test_read_write_m_probs():
+    data = pd.read_csv(
+        pathlib.Path(__file__).parent.parent / "assets" / "patient_lol.csv",
```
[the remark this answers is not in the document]
bamader: Yeah sure thing, `lol` means list of lists. I wish there were something less stupid to call it 😂

tests/linkage/test_linkage.py
bamader: Yeah this is a totally fair point. In our contributing guidelines (I think), we spell out the conventions that we encourage testing against, and unexpected inputs (to a certain degree) are definitely among those things. In python you can have a little bit more trust than in java, but pedagogically speaking we should be testing on at least some capacities of weird inputs. 

In the case of a `true_matches` dict that doesn't conform to shape, that's a bit trickier of an edge case because there will never be a `true_matches` set created in production. The true matches dictionary is only ever used (and even known) when doing preprocessing and pre-computation, which means it should only ever be created by using a preprocessing function that identifies true matches in some toy data set. Given that constraint, I wound up not writing checks for a malformed true matches set because I thought a user should never be willingly creating one, but I also recognize that a user might not be _trying_ to create a malformed set and do it accidentally lol. I can add a test for input in if you'd like one?

tests/linkage/test_linkage.py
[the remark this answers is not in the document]
emmastephenson: Aha! I was thinking the more conventional "laugh out loud" but knew that couldn't be right 🤣 Thanks!

phdi/linkage/link.py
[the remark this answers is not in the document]
emmastephenson: Ooooh thank you for explaining this! Took a couple close reads but I think I've got the idea now 😆 It's a very neat way of solving the problem of double-counting, and seems highly efficient to boot! Thanks again for the detailed explanation!

emmastephenson (approved):
LGTM! Thanks again for the detailed explanations!

phdi/linkage/link.py
m-goggins: I think your reasoning for punting makes sense to me.

m-goggins (approved):
Looks good to me!

emmastephenson:
Fixes #217, #218

PR #240, opened 2023-03-03
Added fatal error type to error messages

nickbristow:


phdi/validation/validation.py
```diff
@@ -24,40 +24,39 @@ def validate_ecr(ecr_message: str, config: dict, error_types: list) -> dict:
     # TODO: utilize the error_types to filter out the different error message
     # types as well as specify the difference between the different error types
     # during the validation process
-
-    error_messages = []
-    warning_messages = []
-    messages = []
+    msgs = {"fatal": [], "error": [], "warning": [], "information": []}
```
KennethSkylight, line 27: can we use the full word messages instead of abbreviation

phdi/validation/validation.py
```diff
@@ -24,40 +24,39 @@ def validate_ecr(ecr_message: str, config: dict, error_types: list) -> dict:
...
 
     for field in config.get("fields"):
         cda_path = field.get("cdaPath")
         matched_xml_elements = _match_nodes(
             xml_elements=parsed_ecr.xpath(cda_path, namespaces=namespaces),
             config_field=field,
         )
+        message_type = (
```
KennethSkylight, line 35: Having an error type should be required by our function, so this should be added to the validate_config... But I think it's already there?

KennethSkylight (changes requested):
A few minor changes needed

phdi/validation/validation.py
BradySkylight: Correct, I am using the full word instead of abbreviations.

phdi/validation/validation.py
BradySkylight: There already is a step to validate the error types passed in at the container level.  We can move this down to a lower level if you would like, as perhaps the SDK version will be used and not just at the container level.  Let me know your thoughts.

phdi/validation/validation.py
```diff
@@ -24,40 +24,39 @@ def validate_ecr(ecr_message: str, config: dict, error_types: list) -> dict:
...
-        messages.append("Validation complete with no errors!")
+        msgs["information"].append("Validation complete with no errors!")
     response = {
         "message_valid": valid,
         "validation_results": _organize_messages(
-            errors=error_messages, warnings=warning_messages, information=messages
+            fatal=msgs["fatal"],
+            errors=msgs["error"],
+            warnings=msgs["warning"],
+            information=msgs["information"],
```
BradySkylight, line 59: I love this!  Much cleaner and easier to use as well as more organized.  Was happy to introduce this into my branch!  Nice work Nick!

phdi/validation/validation.py
BradySkylight: Also, I think we will have to always include fatal errors at this time.  No filtering should be allowed for fatal errors, which should make the validation fail anyway.  Thoughts there as well?

BradySkylight (commented):
Just a few comments and questions!  Overall, it looks good!

BradySkylight (approved):
LGTM!

KennethSkylight (approved):

PR #241, opened 2023-03-03
Add ability to filter error message types

### LINKED ISSUE #199: Filter specific error types
### Why are we doing this?
See #194.

### Action Requested
Implement the ability to filter specific error types (filter warnings out - just don't add them to the dictionary).

Validation for the different 'levels' of errors, or error types, is still performed, but the messages are just filtered from the validation_results dict in the return response.

### Acceptance Criteria
Error filtering is in place on the eCR validation building block.

BradySkylight:
This PR resolves #199 - and it handles the following:

* Updated the different error message types to be fully spelled out for readability
* Updates the organize_error_messages function to filter the lists of the different error message types if they are not indicated to be included
* Moved the config validator into the container section instead of the SDK - if the config is bad we should fail prior to trying to call the SDK BB for validation
* Other general fixes and clean up
* Added and updated tests cases 
* Updated the sample ecr config to be the same in both the base config folder as well as the tests/assets folder

### DIFF
--- a/phdi/validation/validation.py
+++ b/phdi/validation/validation.py
@@ -22,10 +22,10 @@ def validate_ecr(ecr_message: str, config: dict, include_error_types: list) -> d
     try:
         parsed_ecr = etree.fromstring(xml, parser=parser)
         parsed_ecr.xpath("//hl7:ClinicalDocument", namespaces=namespaces)
-    except AttributeError as error:
+    except AttributeError:
         return {
             "message_valid": False,
-            "validation_results": {"errors": ["eCR Message is not valid XML!" + error]},
+            "validation_results": {"errors": ["eCR Message is not valid XML!"]},
         }
 
     error_messages = []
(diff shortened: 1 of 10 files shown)


phdi/validation/validation.py
```diff
@@ -11,74 +11,90 @@
...
+    if "information" in include_error_types:
+        filtered_information = information
 
-def _organize_messages(errors: list, warnings: list, information: list) -> dict:
     organized_messages = {
-        "errors": errors,
-        "warnings": warnings,
-        "information": information,
+        "fatal": fatal,
+        "errors": filtered_errors,
```
KennethSkylight, line 95: I think we can change these lines to something like

`errors if "error" in included_error_types else []` for a single line conditional assignment

then we can get rid of lines 84-91

KennethSkylight (commented):
Had one comment to reduce number of lines

phdi/validation/validation.py
BradySkylight: Great suggestion and easy to implement.  Done!

nickbristow (approved):

KennethSkylight (approved):
lgtm

Issue #242, opened 2023-03-06
Implement a message parsing service


DanPaseltiner:
As a user of DIBBs services I would like to be able to extract specific fields from a given message.

emmastephenson:
Give it a mapping between the fields you want and our FHIR server, this message service will determine the paths needed and extract resources from the FHIR server.

This is what we'll use to extract data out of eCR instead of using the tabulation service, which is not ready.

DanPaseltiner:
fixes #242

PR #246, opened 2023-03-07
Add a message parsing service

DanPaseltiner:
This PR introduces and new containerized service for parsing desired values from messages. It uses a "parsing schema", simply a mapping between desired fields names and the FHIRpaths used to locate them in a FHIR bundle. If the message to be parse is not FHIR the service can attempt calling our FHIR converter to first convert the message to FHIR.

Notes:

1. What are folks' thoughts on the `main.py` file? I'm wondering if it is starting to be over loaded. Maybe the request and response models should be moved into a separate file to clean things up a bit?
2. ~~Updates to the `description.md` are incoming!~~


containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+            values.get("parsing_schema") != {}
+            and values.get("parsing_schema_name") != ""
+        ):
+            raise ValueError(
+                "Values for both 'parsing_schema' and 'parsing_schema_name' have been "
+                "provided. Only one of these values is permited."
+            )
+        return values
+
+    @root_validator
```
nickclyde, line 110: Nice use of validators!

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+@app.post("/parse_message", status_code=200)
+async def parse_message_endpoint(
+    input: ParseMessageInput, response: Response
+) -> ParseMessageResponse:
+    """
+    Extract the desired values values from a message. If the message is not already in
+    FHIR format convert it to FHIR first.
+
+    :param input: A JSON formated request body with schema specified by the
+        ExtractInput model.
```
nickclyde, line 147: Should this be `ParseMessageInput` instead of `ExtractInput`?

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+    :return: A JSON formated response body with schema specified by the ExtractResponse
```
nickclyde, line 148: Same, should this be `ParseMessageResponse`?

nickclyde (approved):
This looks great! I don't think `main.py` is too long, I think it's fine to leave as is!

containers/message-parser/app/main.py
DanPaseltiner: Thank you! I re-named this endpoint at some point. I'll fix this.

containers/message-parser/app/main.py
DanPaseltiner: https://github.com/CDCgov/phdi/pull/246#discussion_r1129908149

emmastephenson:
Fixes #242

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+
+# Instantiate FastAPI and set metadata.
+description = Path("description.md").read_text(encoding="utf-8")
+app = FastAPI(
+    title="PHDI Message Parser",
+    version="0.0.1",
+    contact={
+        "name": "CDC Public Health Data Infrastructure",
+        "url": "https://cdcgov.github.io/phdi-site/",
+        "email": "dmibuildingblocks@cdc.gov",
```
BradySkylight, line 27: This is more of a curiosity question... is this a real email address?  Do we have access to this?

containers/message-parser/app/utils.py
```diff
@@ -0,0 +1,175 @@
...
+from phdi.transport.http import http_request_with_retry
+from phdi.cloud.azure import AzureCredentialManager
+from phdi.cloud.core import BaseCredentialManager
+from phdi.cloud.gcp import GcpCredentialManager
+
+
+@cache
+def load_parsing_schema(schema_name: str) -> dict:
+    """
+    Given a path load and extraction schema.
```
BradySkylight, line 19: should this be 'Given a path load an extraction schema'?

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    # 1. Load schema.
+    if input.parsing_schema != {}:
+        parsing_schema = input.parsing_schema
+    else:
+        try:
+            parsing_schema = load_parsing_schema(input.parsing_schema_name)
+        except FileNotFoundError as error:
+            response.status_code = status.HTTP_400_BAD_REQUEST
+            return {"message": error.__str__(), "parsed_values": {}}
```
BradySkylight, line 159: Have you considered possibly validating the schema prior to utilizing it?  Similar to what is being done in tabulation?

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+            credential_manager=input.credential_manager,
+        )
+        if fhir_converter_response.status_code == 200:
+            input.message = fhir_converter_response.json()["FhirResource"]
+        else:
+            response.status_code = status.HTTP_400_BAD_REQUEST
+            return {
+                "message": f"Failed to convert to FHIR: {fhir_converter_response.text}",
+                "parsed_values": {},
+            }
```
BradySkylight, line 187: Awesome!

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # 3. Generate parsers for FHIRpaths specified in schema.
+    parsers = get_parsers(frozendict(parsing_schema))
```
BradySkylight, line 190: What does frozendict do?

DanPaseltiner: `frozendict` gives us an immutable dictionary. We are using it here so that it can be hashed for caching. Computing the FHIRpath parsers is expensive so caching the results of `get_parsers` significantly improves performance, at least based on our initial experience working with `fhirpathpy` on tabulation.

containers/message-parser/app/main.py
DanPaseltiner: Yes, I have thought about it. I was thinking that we could have additional tickets for:

1. Parsing schema validation
2. Adding support for users to upload additional schemas for the service to store internally. Basically POSTing a schemas to `/schemas`. Ideally we would validate these before accepting them.

containers/message-parser/app/main.py
DanPaseltiner: It is a real email. Not clue if anyone is monitoring it though. I pulled it from the PHDI website.

containers/message-parser/app/utils.py
DanPaseltiner: yes! Thank you!

BradySkylight (approved):
This is great!  It took me a bit to get through.  Other than the small comments/questions I think this looks good.  Nice use of FHIRPath and excellent modularization!

containers/message-parser/app/main.py
BradySkylight: I love both those ideas and they can come later.

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+        description="A message describing the result of a request to "
+        "the /parse_message endpoint."
+    )
+    parsed_values: dict = Field(
+        description="A set of key:value pairs containing the values extracted from the "
+        "message."
+    )
+
+
+@app.post("/parse_message", status_code=200)
```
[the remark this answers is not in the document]
emmastephenson: Reading further through the code and I'm getting the sense that 1 is halfway correct - if an exception is thrown, the response status can change, but it's on the developer to manually change the status?

containers/message-parser/app/utils.py
```diff
@@ -0,0 +1,177 @@
...
+            "including these values or add them as environment variables to this "
+            f"service. missing values: {', '.join(missing_values)}."
+        )
+
+    return message
+
+
+def convert_to_fhir(
+    message: str,
+    message_type: Literal["elr", "vxu", "ecr"],
```
emmastephenson, line 101: This isn't the first time this `Literal` is defined - is there a utility or configuration file available somewhere where we could define this once and then just reuse it throughout the code?

containers/message-parser/app/utils.py
```diff
@@ -0,0 +1,177 @@
...
+    conversion service.
+
+    :param message: The serialized contents of the message to be converted to FHIR.
+    :param message_type: The type of the message.
+    :param fhir_converter_url: The URL of an instance of the FHIR conversion service.
+    :return:
+
+    """
+    conversion_settings = {
+        "elr": {"input_type": "hl7v2", "root_template": "ORU_R01"},
```
emmastephenson, line 117: Same question goes for these definitions - they feel more universal than just the message parsing service.

containers/message-parser/description.md
```diff
@@ -0,0 +1,53 @@
+## Getting Started with the DIBBs message parser
```
emmastephenson, line 1: These docs look great!

containers/message-parser/tests/test_parse_message_endpoint.py
```diff
@@ -0,0 +1,181 @@
...
+with open(fhir_bundle_path, "r") as file:
+    fhir_bundle = json.load(file)
+
+expected_successful_response = {
+    "message": "Parsing succeeded!",
+    "parsed_values": {"first_name": "John ", "last_name": "doe"},
+}
+
+
+def test_parse_message_success_internal_schema():
```
emmastephenson, line 25: These tests are great! Super through 🙌

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+        " the ones come by default."
+    )
+
+
+@app.get("/schemas", status_code=200)
+async def list_schemas() -> ListSchemasResponse:
+    """
+    Get a list of all the parsing schemas currently available. Default schemas are ones
+    that are packaged by default with this service. Custom schemas are any additional
+    schema that users have chosen to upload to this service.
```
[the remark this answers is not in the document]
emmastephenson: Once again I've read further and answered some of my own question :) Seems like these schemas are stored directly in memory along a given directory path, which I assume is fine in this case because everything is self-hosted by the end user.

emmastephenson (commented):
Thank you for sending this my way, and my apologies that it took so long to review! It looks great but I have some general questions - if any of those are easier to chat through over Zoom just let me know; happy to go that route.

containers/message-parser/app/utils.py
DanPaseltiner: This is a good point. I think we could turn this into a function in the Python package.

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+from app.config import get_settings
+
+# Read settings immediately to fail fast in case there are invalid values.
+get_settings()
+
+# Instantiate FastAPI and set metadata.
+description = Path("description.md").read_text(encoding="utf-8")
+app = FastAPI(
+    title="PHDI Message Parser",
+    version="0.0.1",
```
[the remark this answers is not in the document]
DanPaseltiner: Part of implementing our formal release process will be to have a single consistent version across the SDK and all containerized services. This version number is really just a temporary placeholder for now.

containers/message-parser/app/main.py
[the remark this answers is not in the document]
DanPaseltiner: [Here](https://fastapi.tiangolo.com/advanced/response-change-status-code/#use-a-response-parameter) are the doc on dynamically changing a status code. We have just been iterating on this example. I suppose on endpoints that will only ever have one status code we could drop the `status_code=200` in the decorator. Most of our endpoints end up with the potential to return some kind of 4XX because of additional validation we have implemented.

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+    Check service status. If an HTTP 200 status code is returned along with
+    '{"status": "OK"}' then the extraction service is available and running properly.
+    """
+    return {"status": "OK"}
+
+
+# /parse_message endpoint #
+
+
+# Request and respone models
```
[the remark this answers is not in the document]
DanPaseltiner: Thanks

containers/message-parser/app/main.py
```diff
@@ -0,0 +1,264 @@
...
+    license_info={
+        "name": "Creative Commons Zero v1.0 Universal",
+        "url": "https://creativecommons.org/publicdomain/zero/1.0/",
+    },
+    description=description,
+)
+
+
+# /health_check endpoint #
+@app.get("/")
```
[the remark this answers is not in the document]
DanPaseltiner: We haven't discussed this, but I think this would work and would really reduce some of what has just become "boilerplate" every time we add a new service. The question though is where would the definition of this class live? I don't think it fits into the SDK. Maybe at the root level in `containers/`?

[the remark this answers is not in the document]
emmastephenson: The root level of `containers/` feels like an appropriate place to me! And agreed, what prompted this question was a suspicion that we have to do a lot of the same boilerplate for every new service that's added.

PR #250, opened 2023-03-08
Probabilistic enhancements MVP

bamader:
This PR adds the MVP of probabilistic enhancement to our algorithm as an optional boolean parameter. It uses precomputed log odds and some new decision functions to make this work, as well as introduces a graphical profiler for determining the separation boundary between the two bimodal distributions of matches and non-matches.


examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -11,7 +11,14 @@
     perform_linkage_pass,
     compile_match_lists,
     feature_match_fuzzy_string,
+    # calculate_m_probs,
+    # calculate_u_probs,
+    # calculate_log_odds
```
m-goggins, line 16:
```suggestion

```

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -11,7 +11,14 @@
...
     load_json_probs,
+    feature_match_log_odds_exact,
+    feature_match_log_odds_fuzzy_compare,
+    # profile_log_odds,
```
m-goggins, line 20:
```suggestion

```

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -268,11 +309,14 @@ def display_missed_matches_by_type(matches: dict, true_matches: dict):
...
 data = add_split_birth_fields(data)
 data = derive_mrn4(data)
 data = set_record_id(data)
 
+cols = list(data.columns)
+idx_to_col = dict(zip(range(len(cols)), cols))
+
 # start = time.time()
 # m_probs = calculate_m_probs(data, true_matches, file_to_write="m_probs_synthetic.json")  # noqa
 # end = time.time()
```
m-goggins, line 322:
```suggestion

```

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -292,9 +336,25 @@ def display_missed_matches_by_type(matches: dict, true_matches: dict):
 
 start = time.time()
 # matches = lac_validation_linkage(data, None)
```
m-goggins, line 338:
```suggestion
```

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -292,9 +336,25 @@ def display_missed_matches_by_type(matches: dict, true_matches: dict):
...
+
+# profile_log_odds(
+#     data,
+#     true_matches,
+#     log_odds,
+#     [],
+#     ["FIRST", "LAST"],
+#     idx_to_col,
+#     neg_samples=25000,
+# )
```
m-goggins, line 360:
```suggestion
```

phdi/linkage/link.py
```diff
@@ -228,6 +229,22 @@ def eval_perfect_match(feature_comparisons: List) -> bool:
...
+    to be the result of a true patient link instead of just random chance.
+    This is represented using previously computed log-odds ratios.
+
+    :param feature_comparisons: A list of floats representing the log-odds
+      score of each field computed on.
+    :return: Whether the feature comparisons score well enough to be
+      considered a match.
+    """
+    if "true_match_threshold" not in kwargs:
+        raise KeyError("Cutoff threshold for true matches must be passed.")
```
m-goggins, line 244: Do we want to add this as an explicit input?

phdi/linkage/link.py
```diff
@@ -456,6 +527,106 @@ def perform_linkage_pass(
...
+            )
+            non_match_scores.append(score)
+
+    min_length = min(len(true_match_scores), len(non_match_scores))
+    true_match_scores = true_match_scores[:min_length]
+    non_match_scores = non_match_scores[:min_length]
+
+    _, bins, _ = plt.hist(true_match_scores, bins=75, range=[0, 25])
+    _ = plt.hist(non_match_scores, bins=bins, alpha=0.5)
+    plt.show()
```
m-goggins, line 627: This is a blocking function, which could prevent users from proceeding if they do not manually close the window. We could instead save off the graph, use a fig and add `plt.close(fig)`, or use `plt.ion()`. I haven't used the last option before but it [looks like](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.ion.html) it might make it easier to use `.close()` and `.show()`

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -106,27 +113,52 @@ def lac_validation_linkage(
...
-        3: feature_match_fuzzy_string,
-        8: feature_match_exact,
-    }
-    # print("-------Matching on Rule 1: Metaphone first/last, exact DOB-------")
+    if use_log_odds_enhancement:
+        funcs = {
+            0: feature_match_log_odds_fuzzy_compare,
+            2: feature_match_log_odds_fuzzy_compare,
+            3: feature_match_log_odds_fuzzy_compare,
+            8: feature_match_log_odds_exact,
```
m-goggins, line 126: This one is `sex` right? I think we should remove in light of the conversation with LAC last week. It might also be nice to add comments about which columns the `funcs` apply to.

examples/Record-Linkage-sample-data/evaluate_lac.py
```diff
@@ -106,27 +113,52 @@ def lac_validation_linkage(
...
+        }
+        eval_rule = eval_perfect_match
     matches_1 = perform_linkage_pass(
-        data, ["MRN4", "ADDRESS4"], funcs, eval_perfect_match, cluster_ratio, **kwargs
+        data,
+        ["MRN4", "ADDRESS4"],
+        funcs,
+        eval_rule,
+        cluster_ratio,
+        true_match_threshold=16.5,
```
m-goggins, line 143: The `true_match_threshold` is coming from the log_odds ratio JSON, right?

m-goggins (changes requested):
I added a few clarifying questions and removed some commented out sections.  I noticed that the code coverage decreased a bit so you may want to check the misses.

examples/Record-Linkage-sample-data/evaluate_lac.py
bamader: For the commented code on the `evaluate_lac.py` script, I left it in in case someone wanted to just run some of the functions (i.e. calculate new probabilities, etc) without needing to worry about imports. They could just comment/uncomment the parts of the code they were using. I have no problem taking it out if you'd rather do that though?

phdi/linkage/link.py
bamader: The reason I didn't is because our other `eval_` function doesn't use it as a parameter at all, or even accept it. If we made it an explicit param here, then we'd have to make it an explicit param in the `eval_perfect_match` function which didn't make sense to me (and the reason we'd have to mirror the two signatures is because the linkage pass function just does a handoff call to whatever eval function the user supplies, so doing all of the params except the vector of comparisons via `kwargs` allows us to preserve that modularity).

phdi/linkage/link.py
bamader: I guess my thinking was it actually _should_ be a blocking function. The only reason to create the histogram is so that you can visually inspect where the two bimodal distributions separate to figure out the best cutoff score (which I think should also leverage matplotlib's built in graph zooming functionality). I didn't envision a use case where a user would call this function and then leave the histogram up while trying to do other stuff.

That being said, just because I didn't envision it doesn't mean the user won't do it lol. It looks like the link you pasted goes to docs around parameters for interactive mode--wouldn't that be the same kind of blocking function you want to avoid?

examples/Record-Linkage-sample-data/evaluate_lac.py
bamader: Yeah it is, and yeah I can definitely do both of those things

examples/Record-Linkage-sample-data/evaluate_lac.py
bamader: No, that threshold needs to be selected by the user by looking at the overlap between the bimodal distributions of true match scores and non match scores. I guess we could store it in the JSON file after it's been experimentally determined, which would make some sense, but I was envisioning it to be easier for users to have a constant at the top of their script that they would just write the value into after they figured it out. No strong feelings either way so I'm happy to shove it into the JSON if desired!

examples/Record-Linkage-sample-data/evaluate_lac.py
m-goggins: Your reasoning makes sense to me. I'm good to keep it in.

phdi/linkage/link.py
m-goggins: Gotcha, that makes sense!

examples/Record-Linkage-sample-data/evaluate_lac.py
m-goggins: At the top is fine by me, I think we just need to make that clearer/implement in our script.

phdi/linkage/link.py
m-goggins: hmm I'm not sure what I meant to post or what I thought you might get out of those docs ha. I think it's okay to leave for now since I don't anticipate a rush of users but we may want to revisit if this kind profiling is ever rolled into a pipeline.

bamader: Yeah good call. I'll leave a comment with a note to that effect.

m-goggins (approved):
Looks good to me!

PR #253, opened 2023-03-08
Cluster fix

bamader:
This PR addresses a bug in the scoring of the cluster mode of the record linkage algorithm which massively underreported the algorithm's success rate due to pairwise enumeration.


m-goggins (approved):
Looks good to me!

PR #258, opened 2023-03-09
Standardized error messages

nickbristow:
Adds value and standard info of fields


BradySkylight (approved):
This looks good.  We may get feedback from LAC that they want different information, but we can modify this at that point, if necessary.  Nice work Nick!

PR #266, opened 2023-03-10
Fix container build issues

DanPaseltiner:
This PR adds steps to install Git in our dockerfiles where we are currently installing `phdi` from GitHub. It also fixes a couple minor typos in the `message-parser` service docs.


BradySkylight (approved):
Thanks for helping get this figured out and resolved.  Looks like a quick fix!

PR #267, opened 2023-03-10
Auto update GH pages with latest generated pdoc

gordonfarrell:
PR for new system of storing and tracking docs generated by the release workflow. New docs are saved to the docs branch, but the latest docs are also saved into a "Latest" directory that the GH pages site is pointed at. Visiting the pages site will always display the most recent set of docs, but older docs can be linked to using a URL.

Two changes that aren't in this PR but are also part of how this works:

- Update PHDI repo settings so GH pages tracks the docs folder on the docs branch
- Index html update on the docs branch to pull in the docs from the latest directory

Cheers!


DanPaseltiner (approved):
LGTM!

PR #303, opened 2023-03-13
Base MPI Connector Client

m-goggins:
I put together the `BaseMPIConnectorClient`, largely basing it off our [geocoding base client](https://github.com/CDCgov/phdi/blob/main/phdi/geospatial/core.py) as I don't have a ton of experience building out this kind of thing. 

Initially, I thought about including a `connect` method but I think we would be better served with connect to a db within `__init__` so that the specifics of connecting to a Postgres DB vs. SQLite could be handled there. I'm open to any and all suggestions, though.

### DIFF
--- a/phdi/linkage/core.py
+++ b/phdi/linkage/core.py
@@ -1,0 +1,38 @@
+from typing import List, Dict
+from abc import ABC, abstractmethod
+
+
+class BaseMPIConnectorClient(ABC):
+    """
+    Represents a vendor-agnostic Master Patient Index (MPI) connector client. Requires
+    implementing classes to define methods to retrive blocks of data from the MPI.
+    Callers should use the provided interface functions (e.g., geocode_from_str)
+    to interact with the underlying vendor-specific client property.
+    """
+
+    @abstractmethod
+    def connect(**params) -> None:
+        """
+        Connects to a database server.
+
+        :param params: Vendor-specific database connection params.
+
+        """
+        pass  # pragma: no cover
+
+    @abstractmethod
+    def block_data(db_name: str, table_name: str, block_data: Dict) -> List[list]:
+        """
+        Returns a list of lists containing records from the database that match on the
+        incoming record's block values. If blocking on 'ZIP' and the incoming record's
+        zip code is '90210', the resulting block of data would contain records that all
+        have the same zip code of 90210.
+
+        :param db_name: Database name.
+        :param table_name: Table name.
+        :param block_data: Dictionary containing key value pairs for the column name for
+        blocking and the data for the incoming record, e.g., ["ZIP"]: "90210".
+        :return: A list of records that are within the block, e.g., records that all
+        have 90210 as their ZIP.
+        """
+        pass  # pragma: no cover
--- a/phdi/linkage/core.py
+++ b/phdi/linkage/core.py
@@ -29,32 +29,15 @@ class BaseMPIConnectorClient(ABC):
 
     @abstractmethod
     def upsert_match_patient(
-        patient_record: Dict, patient_table: str, person_table: str, person_id=None
+        patient_record: Dict, table_name: str, person_id=None
     ) -> None:
         """
         If a matching ID has been found in the MPI, inserts a new patient record into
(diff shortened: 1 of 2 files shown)


BradySkylight:
> 

I think that the DAL will handle the connection to the DB and we can either use a config for the connection details.  This will abstract the database being utilized and allows us to use our default PGSQL DB, but will also allow us to access other MPIs.  At least that is how I have been thinking about it.  Open for discussion (currently working on Base MPI DAL).

phdi/linkage/core.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Represents a vendor-agnostic Master Patient Index (MPI) connector client. Requires
+    implementing classes to define methods to retrive blocks of data from the MPI.
+    Callers should use the provided interface functions (e.g., geocode_from_str)
+    to interact with the underlying vendor-specific client property.
+    """
+
+    @abstractmethod
+    def block_data(db_name: str, table_name: str, block_data: Dict) -> List[list]:
+        """
```
BradySkylight, line 15: We may want to have this more abstract than this, as the DAL should handle the information regarding the db_name and table names...however, the block_data makes sense.

m-goggins: @BradySkylight does the further abstraction of this still make sense without the DAL?

m-goggins:
@BradySkylight I added two upsert functions. I decided to split out the upsert by patient and person table but let me know if you think they should go together. I suspect that we'll know which way these functions should be created once we start implementing the postgres version.

phdi/linkage/core.py
```diff
@@ -0,0 +1,44 @@
...
+        pass  # pragma: no cover
+
+    @abstractmethod
+    def upsert_match_patient(
+        patient_record: Dict, table_name: str, person_id=None
+    ) -> None:
+        """
+        If a matching ID has been found in the MPI, inserts a new patient record into
+        a table with a matching personID; else a new patient record is
+        inserted with a new personID.
```
BradySkylight, line 37: I think this is close.  "If a matching personID has been found in the MPI, then insert a new patient and update the person to link to the new patient; else a new patient is inserted and a new person is inserted with a new personID (the new patient will be linked to the newly created personID)."  You can wordsmith it better, but that is essentially the logic underneath the hood.

m-goggins: Nice, your addition makes it clearer! I updated to reflect that change and spelled out `patient_table` and `person_table` inputs to further clarify.

phdi/linkage/core.py
```diff
@@ -0,0 +1,49 @@
...
+    ) -> None:
+        """
+        If a matching person ID has been found in the MPI, inserts a new patient into
+        the patient table and updates the person table to link to the new patient; else
+        inserts a new patient into the patient table and inserts a new person into the
+        person table with a new personID, linking the new personID to the new patient.
+
+        :param patient_record: A FHIR patient resource.
+        :param patient_table_name: Name of patient table to update or insert into.
+        :param person_table_name: Name of person table to update or insert into.
```
BradySkylight, line 45: This may be nit-picky, but my only concern with adding in 'table' is if there is a case where we are interacting with an API MPI rather than a DB MPI, if that makes sense.  We can always change it later as well.  Just wanted to mention it and see what you thought.

BradySkylight (approved):
This looks great and thanks for the quick changes, as well as the good discussion!  Other than my comments about using 'table' in the definition, I think we are good to go.  Let me know if you have any questions/input to my comments.   Otherwise it all looks good, but I don't want to hold up our movement forward.

phdi/linkage/core.py
m-goggins: I'm inclined to leave as is for now since we know that's how we'll be setting up the MVP version. But you're absolutely right that we can change when we encounter an API MPI.

bamader: Yep I like leaving as is for now, but good call out

bamader: And probably something that a specific "vendor's" implementation will take account of that may not even need to be in the generic class

bamader (approved):
LGTM!

PR #310, opened 2023-03-15
Add ability to check siblings, parents, or children

nickbristow:
Allows you to configure it to check an elements siblings, parent, grand parent, or children


phdi/validation/validation.py
```diff
@@ -93,25 +93,34 @@ def _get_field_details_string(xml_element, config_field) -> str:
...
-        else []
-    )
-    config_parent_attributes = config_field.get("parent_attributes")
-    parent_attributes = (
-        ["Parent attributes:"]
-        + _get_attributes_list(config_parent_attributes, xml_element)
-        if config_parent_attributes
-        else []
-    )
+    relative_string = []
```
KennethSkylight, line 96: I think as this function gets more complex, we should have a description at the top of the function on what it is trying to do

phdi/validation/validation.py
```diff
@@ -120,11 +129,14 @@ def _get_attributes_list(attributes, xml_element) -> list:
         attribute_name = attribute.get("attributeName")
         reg_ex = attribute.get("regEx")
         reg_ex_string = f" RegEx: '{reg_ex}'" if reg_ex else ""
-        attribute_value = (
-            f" value: '{xml_element.get(attribute_name)}'"
-            if xml_element is not None and xml_element.get(attribute_name)
-            else ""
-        )
+        if (
+            xml_element is not None
```
KennethSkylight, line 133: I think this can be simplified to `if xml_element and xml_element.get(attribute_name)`

None is already falsey

phdi/validation/validation.py
```diff
@@ -170,33 +182,70 @@ def _match_nodes(xml_elements, config_field) -> list:
...
+        if not _check_relatives(xml_element, config_field):
+            continue
         found = _check_field_matches(xml_element, config_field)
 
         if found:
             matching_elements.append(xml_element)
     return matching_elements
 
 
+def _find_related_element(xml_element, cda_path, config_field):
```
KennethSkylight, line 194: We should a description of what it's doing

phdi/validation/validation.py
```diff
@@ -170,33 +182,70 @@ def _match_nodes(xml_elements, config_field) -> list:
...
+    if xml_element is None:
+        return False
+    relative = config_field
+    relative_cda_path = relative.get("cdaPath")
+    relative_tag_name = re.search(r"(?!\:)[a-zA-z]+\w$", relative.get("cdaPath")).group(
+        0
+    )
+    diff = len(cda_path.split("/")) - len(relative_cda_path.split("/"))
```
KennethSkylight, line 202: 202 to like 220 is hard to understand, maybe it could be its own function or add comments to explain what it is?

KennethSkylight (commented):
Just a few suggestions

phdi/validation/validation.py
emmastephenson: +1, even though this is internal I think it'd be helpful to a doc comment.

phdi/validation/validation.py
```diff
@@ -93,25 +93,34 @@ def _get_field_details_string(xml_element, config_field) -> str:
...
-    )
-    config_parent_attributes = config_field.get("parent_attributes")
-    parent_attributes = (
-        ["Parent attributes:"]
-        + _get_attributes_list(config_parent_attributes, xml_element)
-        if config_parent_attributes
-        else []
-    )
+    relative_string = []
+    if config_field.get("relatives"):
```
emmastephenson, line 97: Could we put repeating strings (like "relatives", "name") into a variable? (Could definitely be done in a followup PR)

emmastephenson (commented):
Looks good to me!

phdi/validation/validation.py
BradySkylight: I agree.  I think we should have a doc comment for all of these functions to make it clear what they are doing.  If we can split out some of the functionality into other utility functions or internal functions, that would be great too.  Any additional comments on blocks to help make clear what is occurring would be nice.

phdi/validation/validation.py
BradySkylight: I agree.  I think we should have a doc comment for all of these functions to make it clear what they are doing.  If we can split out some of the functionality into other utility functions or internal functions, that would be great too.  Any additional comments on blocks to help make clear what is occurring would be nice.

phdi/validation/validation.py
BradySkylight: Correct, because the function in lxml that returns the 'block' of elements or the actual xml_element returns None if not found, which is different than just "" or False.  I ran into something similar.

KennethSkylight (approved):

phdi/validation/validation.py
```diff
@@ -86,32 +86,50 @@ def validate_ecr(ecr_message: str, config: dict, include_error_types: list) -> d
 
 
 def _get_field_details_string(xml_element, config_field) -> str:
+    """
+    Gets the name, value, of the field referenced as well as the details of the
+    relatives and formats it into a string for use in messages
+
+    :param xml_elements: A list of xml elements
```
BradySkylight, line 93: Is this singular or multiple xml elements?

phdi/validation/validation.py
```diff
@@ -170,33 +188,99 @@ def _match_nodes(xml_elements, config_field) -> list:
...
             matching_elements.append(xml_element)
     return matching_elements
 
 
+def _find_related_element(xml_element, cda_path, config_field):
+    """
+    Returns an xml element that is related to another xml element based on
+    the cda_path of the element you are searching for
+
+    :param xml_elements: A list of xml elements
```
BradySkylight, line 205: Singular or plural?

phdi/validation/validation.py
```diff
@@ -170,33 +188,99 @@ def _match_nodes(xml_elements, config_field) -> list:
...
+    :param cda_path: A string representing the location of the item being looked for.
+    :param config_field: A dictionary of the requirements of the field.
```
BradySkylight, line 207: You said you are going to change this to relative_cda_path - both comments and the parameter name.  Nice!

BradySkylight (approved):
Looks good!  I'll be making some changes in names and maybe moving some to a utils.py as we have discussed in my next ticket.  Nice work!

PR #312, opened 2023-03-16
Create basic SQL DDL file

bamader:
This PR creates a really basic ddl file for our MPI tables within the linkage folder. The database used at the top should be swapped out whenever we actually stand a DB up, and I left everything transactional so that we get rollback if the creation fails. Let me know if the field lengths don't seem long enough or we want other info here.

### DIFF
--- a/phdi/linkage/tables.ddl
+++ b/phdi/linkage/tables.ddl
@@ -1,0 +1,18 @@
+USE DibbsMpiDB;
+BEGIN;
+CREATE TABLE IF NOT EXISTS patient (
+    patient_id  VARCHAR(32) PRIMARY KEY
+    person_id   VARCHAR(32) FOREIGN KEY
+    name    JSONB
+    sex ENUM ('male', 'female', 'other', 'unknown')
+    birthdate   DATE
+    telecom JSONB
+    identifier  JSONB
+    address JSONB
+);
+
+CREATE TABLE IF NOT EXISTS person (
+    person_id   VARCHAR(32) PRIMARY KEY
+    external_person_id  VARCHAR(64)
+);
+COMMIT;


phdi/linkage/tables.ddl
```diff
@@ -0,0 +1,18 @@
+USE DibbsMpiDB;
+BEGIN;
+CREATE TABLE IF NOT EXISTS patient (
+    patient_id  VARCHAR(32) PRIMARY KEY
+    person_id   VARCHAR(32) FOREIGN KEY
```
BradySkylight, line 5: I just was looking to confirm that UUIDs, at least in SQL are max of 32 characters.  Looks correct!  There are other lengths in other 'systems', which I find weird, but this looks correct.  We may also want to add the auto-creation of the UUIDs for these identifier columns, however, that syntax may differ in the various sql flavors.  This should work just fine for now.

DanPaseltiner: Postgres has a [UUID type](https://www.postgresql.org/docs/current/datatype-uuid.html)

DanPaseltiner (changes requested):
I thought we were going to simply going to have a single jsonb column containing the entire patient resource contents. [See this conversation](https://skylight-hq.slack.com/archives/C04T9HFFETX/p1678465578607039)

phdi/linkage/tables.ddl
BradySkylight: Right.  If this is specific to PgSQL then I would not only add UUID as the type, but I would add the following to auto-generate the IDs as data is added:  "patient_id uuid DEFAULT uuid_generate_v4 (),"  then at the end of the list of columns add:  "PRIMARY KEY (patient_id)"

BradySkylight: Above I was thinking of a more generic SQL solution.

phdi/linkage/tables.ddl
```diff
@@ -0,0 +1,18 @@
...
+    name    JSONB
+    sex ENUM ('male', 'female', 'other', 'unknown')
+    birthdate   DATE
+    telecom JSONB
+    identifier  JSONB
+    address JSONB
+);
+
+CREATE TABLE IF NOT EXISTS person (
+    person_id   VARCHAR(32) PRIMARY KEY
```
BradySkylight, line 15: if PgSQL specific then I would do the same as recommended above for the UUID type and auto-generate the value.

phdi/linkage/tables.ddl
DanPaseltiner: I think we want a PG specific solution here. If we decide to support additional DB flavors in the future we'll write additional DDL. Let's get the most we can out of the DB we are using. At least that's how I see it.

bamader:
@DanPaseltiner you're absolutely right, that's my bad. I forgot we'd had that slack convo and went back to the original draw.io diagram. Will fix.

bamader:
Had a conversation with Dan this morning around the structure of the Person table, and we agreed to go with a two-column barebones approach: one for a DIBBS generated UUID, and one that stores an external Person ID (e.g. from IRIS). We're opting not to maintain a list of either the patients who are linked to the person, or a list of the UUIDs of those patients. Because we can always recover Person IDs from the Patient-side, aggregating Patients to their respective Person groups is as simple as a JOIN between Patient and Person on Person_ID, followed by a GROUPBY. We think this is simplest and most straightforward for now and will help keep our update process focused on only one table.

phdi/linkage/tables.ddl
```diff
@@ -0,0 +1,15 @@
+USE DibbsMpiDB;
+BEGIN;
+CREATE TABLE IF NOT EXISTS patient (
+    patient_id  UUID DEFAULT uuid_generate_v4 ()
+    person_id   VARCHAR(32) FOREIGN KEY
```
BradySkylight, line 5: Personid in this table should probably also be a UUID...then we don't have to worry about size either. ;)

DanPaseltiner (approved):
Thanks for the conversation this morning @bamader. I'm going to approve so that I'm no longer blocking. Thank you again for putting this together!

BradySkylight (approved):
I'll approve with the assumption that you'll update the patient.person_id to be a UUID type.  Thanks for working through this and your SQL is on-point!

emmastephenson (approved):
Looks great!

bamader:
Documenting here:

Dan, I checked on the uuid generation and yes it looks like postgres will only need to generate a default UUID if we don't supply one, so this should still check out just fine with our FHIR server UUID.

PR #316, opened 2023-03-17
Fix failing Container Validation Failing Tests

BradySkylight:


containers/validation/tests/test_validation.py
```diff
@@ -133,6 +126,7 @@ def test_validate_ecr_invalid():
     actual_result3 = validate_ecr_msg(
         message=sample_file_bad, include_error_types=test_error_types
     )
+    print(actual_result3)
```
nickbristow, line 129: just need to remove

nickbristow (approved):

PR #322, opened 2023-03-20
eCR Validation Refactor and Clean-up

### LINKED ISSUE #298: eCR Validator clean-up Refactor
Why are we doing this:

1- Separate out the XML validation from just validation in general - where possible.  This way, when we include HL7V2 validation etc... we can possibly still leverage some of the functions
2- If there are xml specific validation functions, they should be labeled as such so that developers can understand what is just for XMl and what is not.  This will prevent the incorrect use of functions and just make it clearer in the code.
3- Add Doc comments to all functions to make it clear what each function is doing.
4- Split out functionality that is reused into separate utility/helper/internal function if necessary and possible... (but don't go too overboard with this).
5- We may be able to remove the validated_message portion of the dict response

BradySkylight:
The following PR fixes #298 and includes the following key updates:

1- Separate out the XML validation, into it's own util file, from just validation in general - where possible. This way, when we include HL7V2 validation etc... we can possibly still leverage some of the functions
2- If there are xml specific validation functions, they should be labeled as such so that developers can understand what is just for XMl and what is not. This will prevent the incorrect use of functions and just make it clearer in the code. (1- kind of handles this, but I wanted to call out the naming of the functions as well)
3- Add Doc comments to all functions to make it clear what each function is doing.
4- Split out functionality that is reused into separate utility/helper/internal function if necessary and possible... (but don't go too overboard with this).
5- We will remove the validated_message portion of the dict response

1-4 Are complete and 5 is complete except not on the container side - a separate PR will handle the container update to prevent failures.


nickbristow (approved):

DanPaseltiner (approved):
LGTM. Thanks @BradySkylight

PR #323, opened 2023-03-20
Postgres implementation of MPI connector

m-goggins:
This PR creates the Postgres implementation of an MPI connector. The postgres implementation has two main functions and 1 helper function:

- `block_data`: returns blocks of data from the patient table as specified by the user-provided `block_data` dictionary
- `upsert_patient_match`: updates patient and person tables depending on whether a match was found or not
- `_generate_block_query`: generates a Postgres/JSONB specific query for retrieving blocks of data. This one will likely need to be modified for the specific key, values we allow users to query with as the current query generator only goes 1 level deep and could be further specified for the `patient_resource` such that it knows how to generate a query for zip code, for example. 

I tested all of the functions locally and Nick C helped me with the GH actions for testing. The SQL could definitely be improved, especially with regard to the UUIDs and JSONB formatting, but I wanted to get this out for review sooner rather than later.


phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+        self.database = database
+        self.user = user
+        self.password = password
+        self.host = host
+        self.port = port
+        self.patient_table = patient_table
+        self.person_table = person_table
+
+        try:
+            self.connection = psycopg2.connect(
```
emmastephenson, line 34: We may want to open and close this connection at the start of a given function call, rather than leaving it permanently open. I'm a little concerned that by using a single connection for every function call, we'd run into issues around throttling or locking rows (or hitting our max connection limits if we have multiple instances running).

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+        :param person_id: The personID matching the patient record if a match has been
+        found in the MPI, defaults to None.
+        """
+        # Match has been found
+        if person_id is not None:
+            # Insert into patient table
+            insert_patient_table = (
+                f"INSERT INTO {self.patient_table} "
+                + "(patient_id, person_id, patient_resource) "
+                + f"""VALUES ('{patient_resource.get("id")}', '{person_id}', """
```
emmastephenson, line 102: Is the patient resource ID guaranteed to be a UUID?

tests/linkage/test_postgres_mpi_connector.py
```diff
@@ -0,0 +1,290 @@
+from phdi.linkage.postgres import PostgresConnectorClient
+import pathlib
+import pytest
+import json
+
+
+def test_postgres_connection():
```
emmastephenson, line 7: Will these tests pass locally without a postgres database stood up? Or are they intended to only run within the Github action?

emmastephenson (commented):
Overall this is looking great, thanks Marcelle! Just a couple of questions.

bamader (changes requested):
The dictionaries of commands and SQL curly bracketed stuff is making my eyes bleed 😭 🤮 . Major props to you for putting this altogether. I had a couple questions about turning the blocked data into LoLs and a big discussion point around external person ID that I think Dan and Brady should weigh in on, but otherwise this is so promising to see 💪

phdi/linkage/postgres.py
BradySkylight: This is a good point, but not sure if the connection is the issue or if the cursor matters.  We need to make sure to close the cursors after each operation is completed.  I haven't worked on DB connections in python so much, but typically we would have a db connection factory in Java which managed the connections.  I wonder if there is some way to do that in python.  I'll do some digging..

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+            # Insert into patient table
+            insert_patient_table = (
+                f"INSERT INTO {self.patient_table} "
+                + "(patient_id, person_id, patient_resource) "
+                + f"VALUES ('{patient_resource.get('id')}','{person_id[0][0]}', "
+                + f"'{json.dumps(patient_resource)}');"
+            )
+            self.cursor.execute(insert_patient_table)
+            self.connection.commit()
+
```
BradySkylight, line 147: I'm not sure, based upon the very little of what I've read about psycopg2, if we need to close the cursors manually or not....but it may just be a good practice to close them once the query has completed what it needs to do.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+        for key in list(extracted_data[0][-1].keys()):
+            blocked_data[0].append(key)
+
+        # Unnest patient_resource data
+        for row in extracted_data:
+            row_data = [row[0], row[1]]
+            for value in list(row[-1].values()):
+                row_data.append(value)
+            blocked_data.append(row_data)
+
```
BradySkylight, line 78: I'm not sure, based upon the very little of what I've read about psycopg2, if we need to close the cursors manually or not....but it may just be a good practice to close them once the query has completed what it needs to do.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+        return blocked_data
+
+    def upsert_match_patient(
+        self,
+        patient_resource: Dict,
+        person_id=None,
+    ) -> None:
+        """
+        If a matching person ID has been found in the MPI, inserts a new patient into
+        the patient table and updates the person table to link to the new patient; else
```
BradySkylight, line 88: Based upon our model, this comment should change.  I know I was the one who helped come up with this note, so I take full responsibility here.  

"If a matching person ID has been found in the MPI, inserts a new patient into the patient table with the provided person ID; else inserts a new person into the person table and inserts a new patient with the new personID, from the newly created person record, linking the new person with the new patient."

BradySkylight: Probably should update the comment in the Base class as well.

phdi/linkage/postgres.py
BradySkylight: Yes, it is.  Should we have two columns for patientid, like a patient_id that is the UUID in postgres, and another that contains the FHIR Patient Identifier (resource id) (maybe name the column 'patient_resource_id')?

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+            self.cursor.execute(insert_person_table)
+            self.connection.commit()
+
+            # Retrieve newly generated person_id
+            select_statement = (
+                f"SELECT person_id from {self.person_table} "
+                + f"WHERE external_person_id = '{patient_resource.get('id')}'"
+            )
+            self.cursor.execute(select_statement)
+            self.connection.commit()
```
BradySkylight, line 135: Are commit's required for select statements?  We may just need to fetch the results, as you are doing below.  Again, new to the postgres python package, so just checking.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+            person_id = self.cursor.fetchall()
```
BradySkylight, line 136: Theoretically, there should only be one result returned.  Should we have a check around this or a looping mechanism to ensure there is only one record?  If not, and we let the DB handle the errors by preventing duplicates, that would be fine too, but we will just need to make sure there is a constraint on the table to ensure this.  We could just use cursor.fetchone() as well, if we are going to allow the DB to manage handling duplicates.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        # TODO: Update queries for nested values, e.g., zip within address
+        query_stub = f"SELECT * FROM {table_name} WHERE "
+        block_query = " AND ".join(
+            [
+                f"patient_resource->>'{key}' = '{value}'"
+                if type(value) == str
+                else (f"'{key}' = {value}")
+                for key, value in block_data.items()
+            ]
```
BradySkylight, line 167: Nice!  I see the TODO above, but at least this is a good start!

tests/linkage/test_postgres_mpi_connector.py
```diff
@@ -0,0 +1,290 @@
...
+    # Clean up
+    postgres_client.cursor.execute(
+        f"DROP TABLE IF EXISTS {postgres_client.patient_table}"
+    )
+    postgres_client.connection.commit()
+    postgres_client.cursor.execute(
+        f"DROP TABLE IF EXISTS {postgres_client.person_table}"
+    )
+    postgres_client.connection.commit()
+    postgres_client.connection.close()
```
BradySkylight, line 290: I like the table creation functions and the clean-up here.  Sweet!

BradySkylight (changes requested):
Looks good overall! A few questions which may result in some changes.  I'm open/willing to discuss in more detail if needed/desired.

phdi/linkage/postgres.py
m-goggins: Yep, I based if off the [tables.ddl](https://github.com/CDCgov/phdi/blob/main/phdi/linkage/tables.ddl)

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+import json
+
+
+class PostgresConnectorClient(BaseMPIConnectorClient):
+    """
+    Represents a Postgres-specific Master Patient Index (MPI) connector client. Requires
+    implementing classes to define methods to retrive blocks of data from the MPI.
+    Callers should use the provided interface functions (e.g., geocode_from_str)
+    to interact with the underlying vendor-specific client property.
+    """
```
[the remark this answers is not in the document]
m-goggins: Oops, good call.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def __init__(
+        self,
+        database: str,
+        user: str,
+        password: str,
+        host: str,
+        port: str,
+        patient_table: str,
+        person_table: str,
```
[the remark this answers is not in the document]
m-goggins: Correct! Initially, I didn't think it should be required for the class, but if it's not specified here, the user has to specify the tables in the underlying methods (e.g., specify patient table name for `block_data`). I can see it both ways if you think we shouldn't make it required for the class so I'm okay to change if others want to.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+        query = self._generate_block_query(self.patient_table, block_data)
+
+        # Execute query
+        self.cursor.execute(query)
+        extracted_data = self.cursor.fetchall()
+
+        # Set up blocked data
+        blocked_data = [["patient_id", "person_id"]]
+        for key in list(extracted_data[0][-1].keys()):
+            blocked_data[0].append(key)
```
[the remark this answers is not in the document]
m-goggins: It works exactly like the last sentence you wrote for #1! 

I do think you're right about point 2. We should implement some kind of check to make sure the data are returned in the same order. Sorting the keys alphabetically would be a low lift to implement.

phdi/linkage/postgres.py
emmastephenson: Yep, this is exactly what I'm thinking Brady! That's the pattern I'm used to as well - there's a factory to handle all these transactions for you, so you can just say `execute query` and the factory will handle setting up the connection, closing it at the end of the transaction, handling retries, etc.

m-goggins: Oh yeah, that's a really good point that I forgot to return to and I manually close the connection in the tests so it didn't come back as an issue. 

In this case, it would move the connection out of the class and into a separate factory method, right?

phdi/linkage/postgres.py
m-goggins: I'm pretty sure when you close the connection, the cursor is also closed but I'm also new to psycopg2 so I could be wrong.

phdi/linkage/postgres.py
m-goggins: ditto above

phdi/linkage/postgres.py
emmastephenson: +1 to this Brady - I think we can [close them manually](https://www.psycopg.org/docs/cursor.html#cursor.close), so we probably should.

phdi/linkage/postgres.py
m-goggins: I think you're correct; we don't need it for select statements since we're not changing anything in the table (so wouldn't need to roll back).

phdi/linkage/postgres.py
m-goggins: Good call. I do think we want to check to make sure that only 1 result is returned. I hesitate to use fetchone() though. If there is more than 1 result, it doesn't seem like[ fetchone()](https://www.psycopg.org/docs/cursor.html?highlight=fetchone#cursor.fetchone) orders the results in a particular way. And if there is more than one result, I think we should alert the user to an error. What do you think about raising an exception here?

tests/linkage/test_postgres_mpi_connector.py
```diff
@@ -0,0 +1,290 @@
...
+def test_postgres_connection():
+    postgres_client = PostgresConnectorClient(
+        database="testdb",
+        user="postgres",
+        password="pw",
+        host="localhost",
+        port="5432",
+        patient_table="test_patient_mpi",
+        person_table="test_person_mpi",
+    )
```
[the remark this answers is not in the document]
m-goggins: @emmastephenson asked about this same thing so I can put together some documentation. It's pretty simple but would be good to have written down somewhere. Where would be an appropriate place for the write up?

phdi/linkage/postgres.py
emmastephenson: That's what I'm thinking Marcelle. If we don't want to bother with a factory yet we could also manage the opening and closing manually within this class; just moving the connection setup into the methods instead of at class instantiation and closing both connections and cursors before the method returns.
The psycopg library also seems to support [context managers](https://www.psycopg.org/docs/connection.html?highlight=connection#connection:~:text=Connections%20can,for%20several%20contexts.), if we wanted to go that route instead.

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+            self.cursor.execute(insert_person_table)
+            self.connection.commit()
+
+        # Match has not been found
+        else:
+            # Insert a new record into person table to generate new person_id
+            insert_person_table = (
+                f"INSERT INTO {self.person_table} "
+                + "(external_person_id) "
+                + f"VALUES ('{patient_resource.get('id')}');"
```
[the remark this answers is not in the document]
m-goggins: Ah I think I misunderstood the ddl file then. I thought that the `person.external_person_id` should correspond to the `patient.patient_id` in the patient table and to the patient's FHIR ID. 

I only inserted so that I can retrieve the newly created `person_id` as you point out in another comment. 

Definitely want to get on the same page about this.

tests/linkage/test_postgres_mpi_connector.py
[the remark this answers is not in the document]
emmastephenson: @m-goggins in the new DIBBs wiki would be great! 😄 https://github.com/CDCgov/phdi/wiki/Local-Development

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+            self.connection.commit()
+
+            # Insert into person table
+            insert_person_table = (
+                f"INSERT INTO {self.person_table} "
+                + "(person_id, external_person_id) "
+                + f"VALUES ('{person_id}', '{patient_resource.get('id')}');"
+            )
+            self.cursor.execute(insert_person_table)
+            self.connection.commit()
```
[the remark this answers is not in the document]
m-goggins: Will the FHIR's patient ID be the same every time for the same patient?

phdi/linkage/postgres.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+            self.cursor.execute(insert_patient_table)
+            self.connection.commit()
+
+            # Insert into person table
+            insert_person_table = (
+                f"INSERT INTO {self.person_table} "
+                + "(person_id, external_person_id) "
+                + f"VALUES ('{person_id}', '{patient_resource.get('id')}');"
+            )
```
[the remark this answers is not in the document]
m-goggins: Will the FHIR patient resource id be the same every time for the same patient? I think this goes to Brandon's comment in the internal channel as to whether we should use the resource id as the `patient_id` and whether that will be consistent over time.

tests/linkage/test_postgres_mpi_connector.py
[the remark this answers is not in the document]
m-goggins: Updated with directions

phdi/linkage/postgres.py
bamader: I would say we don't need a separate patient_resource_id column, personally. That information is very easily accessed just using a JSON query in the patient resource itself, and we deliberately structured our tables to lean and small to get this started. Let's not start adding more columns quite yet 😅

tests/linkage/test_postgres_mpi_connector.py
bamader: Yeah Emma's comment gets at what I was talking about with some small documentation additions around testing this locally. Without setting up a container with that postgres image, you'll never get a successfully running `pytest` from the root test directory, which is imo an important thing to be able to see for large scale refactors or code work. I don't mind that the tests require a DB set up locally, but I do think a little paragraph (maybe in the readme somewhere under testing?) about configuring the docker container and its DB image would be good.

phdi/linkage/postgres.py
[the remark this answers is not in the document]
bamader: No strong opinions on my part! If nobody else comments on this I say we leave as is.

phdi/linkage/postgres.py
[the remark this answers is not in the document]
bamader: I don't think that's guaranteed, no. If two patients come in that are super similar to each other but have a typo'ed letter off, the FHIR server won't recognize them as the same on its own so they'll get different IDs. And if two patient records really are the same, because we standardize and link before we store back to the FHIR server, I don't know if the FHIR IDs will always match up pre-linkage anyway.

phdi/linkage/postgres.py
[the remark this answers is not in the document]
bamader: For sure. In the ideal world, I would say we want to insert null or none into the external ID, because that's the truest reflection of what that column in the table is supposed to hold--for a patient that's new and has just come in, it _won't_ yet be in another database, so it shouldn't have an external ID. Based on my quick google search, it looks like we can do that while still getting the ID of the thing we just created using something called a Returning Clause, https://www.postgresql.org/docs/8.3/sql-insert.html (ctrl+F returning). https://stackoverflow.com/questions/2143379/returning-a-value-with-psycopg2

phdi/linkage/postgres.py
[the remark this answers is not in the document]
bamader: Looks like psycopg2 supports returning clauses so we can use those

bamader: I have the same hesitation you do Marcelle so I like your suggestion of using fetchall but raising an exception if there's more than 1.

phdi/linkage/postgres.py
DanPaseltiner: @emmastephenson for context the FHIR converter creates UUIDs for all of the FHIR resources it generates. We are going to use this ID from the patient resource as the ID in the MPI.

phdi/linkage/postgres.py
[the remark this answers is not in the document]
BradySkylight: Into the Person table though?  A patient table insert every time, yes, but not a person insert unless the patient doesn't find a match.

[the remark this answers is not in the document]
DanPaseltiner: yes!

BradySkylight (approved):
Approving, based upon our conversation we had yesterday.  Pending final updates and tests.  Excellent work on this!

BradySkylight (approved):
Nice updates!  Great work on this!  LGTM

emmastephenson (approved):
This looks great! Thanks for all your hard work here, Marcelle!

PR #324, opened 2023-03-21
Read and write algo as JSON

bamader:
This PR takes the first step towards implementing record linkage in a containerized service. In order to allow users to either customize the linkage algorithm as they wish, or to simply access the DIBBS Default algorithm, we need easy ways to read and write configuration files for the details of the algorithm. This PR adds those ways.

Once communication with the MPI is done, and functions to return blocks of data queried from the MPI are implemented, these functions can be used to write a `run_linkage_from_config` function that actually does linkage. 

Tagging Dan on this as an additional reviewer since this config idea came out of our strategy session.


DanPaseltiner (commented):
ugh typo

tests/linkage/test_linkage.py
```diff
@@ -728,3 +730,64 @@ def test_feature_match_log_odds_fuzzy():
...
+            "matching_rule": eval_perfect_match,
+        },
+        {
+            "funcs": {10: feature_match_fuzzy_string, 16: feature_match_fuzzy_string},
+            "blocks": ["FIRST4", "LAST4"],
+            "matching_rule": eval_perfect_match,
+            "cluster_ratio": 0.9,
+            "kwargs": {"similarity_measure": "JaroWinkler", "threshold": 0.7},
+        },
+    ]
```
DanPaseltiner, line 753: If this is our actual default algorithm I think we probably want this defined in a .json file somewhere else so it can be easily used in practice, and simply loaded from that file for this unit test. We could test reading before writing. I might be misunderstanding something here though.

tests/linkage/test_linkage.py
```diff
@@ -728,3 +730,64 @@ def test_feature_match_log_odds_fuzzy():
...
+def test_read_algo_errors():
+    with pytest.raises(FileNotFoundError) as e:
+        read_linkage_config("invalid.json")
+    assert "No file exists at path invalid.json." in str(e.value)
+    with open("not_valid_json_test.json", "w") as fp:
+        fp.write("this is a random string that is not in json format\n")
+    with pytest.raises(JSONDecodeError) as e:
+        read_linkage_config("not_valid_json_test.json")
+    assert "The specified file is not valid JSON" in str(e.value)
+    os.remove("not_valid_json_test.json")
```
DanPaseltiner, line 793: Nice!

tests/linkage/test_linkage.py
```diff
@@ -728,3 +730,64 @@ def test_feature_match_log_odds_fuzzy():
...
     )
+
+
+def test_algo_read_write():
+    dibbs_deterministic_algo = [
+        {
+            "funcs": {
+                0: feature_match_fuzzy_string,
+                2: feature_match_fuzzy_string,
+                3: feature_match_fuzzy_string,
```
DanPaseltiner, line 741: I don't think we should sink time into changing this now, but in the future I think it would be worth considering doing this mapping between functions and specific columns using column name. I realize column requires inclusion of a header and may be marginally less performant than selecting a column by index, however selection by index can be very brittle. Using column name is robust against the order columns in a table.

DanPaseltiner (changes requested):
Thanks for this @bamader! A couple pretty minor things. The only thing I feel particularly strongly about is that we should define the dibbs default algorithm in its own .json file instead of hard coding into the unit test. This will clean up the unit tests a bit, but more importantly it puts that config in a format where it is ready to be packaged into the record linkage container and shipped as part of the image.

m-goggins (approved):
This looks really good to me! The docstrings are clear and everything was well documented so i was able to follow. As part of closing out record linkage, we should make sure we add additional documentation (like we did for tabulation), which I think will make this part of the code much more understandable to first-time readers/users.

BradySkylight (approved):
No other comments/requests to add other than, this is great to have!  LGTM!

tests/linkage/test_linkage.py
bamader: I've now added two pre-defined algorithms, dibbs basic and dibbs enhanced, as json configs in the linkage directory. I've also re-written/re-ordered the tests so that we test not only loading our own config files, but a more robust set of parameters in the read/write (e.g. making sure we can appropriately load all the different feature comparison functions we offer).

DanPaseltiner (approved):
Thanks for the changes @bamader. LGTM!

phdi/linkage/link.py
```diff
@@ -838,6 +867,49 @@ def score_linkage_vs_truth(
...
+    name of a function in the DIBBS linkage module (such as
+    `feature_match_fuzzy_string`)--note that these are the actual
+    functions, not string names of the functions
+    - `"blocks"` should map to a list of columns to block on (e.g.
+    ["MRN4", "ADDRESS4"])
+    - `"matching_rule"` should map to one of the evaluation rule functions
+    in the DIBBS linkage module (i.e. `eval_perfect_match`)
+    - `"cluster_ratio"` should map to a float, if provided
+    - `"kwargs"` should map to a dictionary of keyword arguments and their
+    associated values, if provided
```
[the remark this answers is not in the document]
bamader: For sure!

bamader:
The container failure is a known issue that Brady has a PR up for, so I'm treating that as non-blocking for purposes of merging this PR.

PR #336, opened 2023-03-24
Change ccda to ecr in input_types for fhir converter

gordonfarrell:


DanPaseltiner (commented):
LGTM. Let's just get the tests passing.

PR #346, opened 2023-03-28
Change to psycopg2-binary.

DanPaseltiner:
This PR changes our psycopg2 dependency from `psycopg2` to `psycopg2-binary`. There are [known](https://stackoverflow.com/questions/71631259/can-not-install-psycopg2-on-mac-12-3-i-keep-getting-this-error) pip installation issues with `pyscopg2` and building directly from source via the `psycopg2-binary` is a simple solution recommended by pip.

>     If you prefer to avoid building psycopg2 from source, please install the PyPI 'psycopg2-binary' package instead.


gordonfarrell (approved):
LGTM

PR #358, opened 2023-03-28
Add licenses as a parameter for geocoding endpoint

### LINKED ISSUE #329: Add license as a parameter for geocoding endpoint
### Why are we doing this?

The smarty building block [currently defaults](https://github.com/CDCgov/phdi/blob/main/phdi/geospatial/smarty.py#L17) to the `us-standard-cloud` license type for Smarty, but our Smarty account now uses a different type of license. Because the geocoding endpoint in the ingestion container doesn't accept a parameter for a different license type, it defaults to `us-standard-cloud` and currently fails in dev. 

### Action Requested

Add license as a parameter for geocoding endpoint

### Acceptance Criteria

GIVEN a user wants to use a different smarty license type
WHEN they run the geocoding endpoint or deploy a pipeline
THEN they want to be able to specify which license type to use

### Additional Context

_No response_

nickclyde:
Resolves #329


bamader (approved):
LGTM! Thanks Nick!

DanPaseltiner (approved):
Send it!

PR #361, opened 2023-03-29
remove 'licenses' from geocode_bundle function call

emmastephenson:
# PULL REQUEST

## Summary
The recent changes to the Smarty license introduced a bug when hitting the geocode endpoint, with the following error:

Screenshot 2023-03-29 at 3 17 40 PM

This change removes the `licenses` param before sending a request to the `geocode_bundle` function, as the `geocode_bundle` function doesn't recognize `licenses`.

## Related Issue
n/a - discovered during demo prep

## Additional Information
Long-term this seems like a good opportunity to add an integration test for the ingestion service, but first things first let's get it working again :) 


nickclyde (approved):
Aha, good catch. I was running into this bug today and was getting confused. Thanks for fixing this!

PR #368, opened 2023-03-30
Change licenses parameter to single license type

nickclyde:
# PULL REQUEST

## Summary
Azure is having trouble parsing environment variables as a list. Since we are unlikely to need to support use of more than once license type at a time for Smarty, this changes our geocode endpoint to only accept a single license type.

### DIFF
--- a/containers/ingestion/app/routers/fhir_geospatial.py
+++ b/containers/ingestion/app/routers/fhir_geospatial.py
@@ -17,13 +17,7 @@ router = APIRouter(
 )
 
 
-class LicenseType(str, Enum):
-    usStandardCloud = "us-standard-cloud"
-    usCoreCloud = "us-core-cloud"
-    usRooftopGeocodingCloud = "us-rooftop-geocoding-cloud"
-    usRooftopGeocodingEnterpriseCloud = "us-rooftop-geocoding-enterprise-cloud"
-    usAutocompleteProCloud = "us-autocomplete-pro-cloud"
-    internationalGlobalPlusCloud = "international-global-plus-cloud"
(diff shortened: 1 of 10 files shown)


containers/ingestion/app/routers/fhir_geospatial.py
```diff
@@ -16,6 +17,15 @@
 )
 
 
+class LicenseType(str, Enum):
+    usStandardCloud = "us-standard-cloud"
+    usCoreCloud = "us-core-cloud"
+    usRooftopGeocodingCloud = "us-rooftop-geocoding-cloud"
+    usRooftopGeocodingEnterpriseCloud = "us-rooftop-geocoding-enterprise-cloud"
+    usAutocompleteProCloud = "us-autocomplete-pro-cloud"
+    internationalGlobalPlusCloud = "international-global-plus-cloud"
```
DanPaseltiner, line 26:
```suggestion
license_types = Literal["us-standard-cloud", "us-core-cloud", "us-rooftop-geocoding-cloud", "us-rooftop-geocoding-enterprise-cloud", "us-autocomplete-pro-cloud", "international-global-plus-cloud"]
```

containers/ingestion/app/routers/fhir_geospatial.py
```diff
@@ -33,9 +43,9 @@ class GeocodeAddressInBundleInput(BaseModel):
         "'geocode_method' is 'smarty'.",
         default="",
     )
-    licenses: Optional[list[str]] = Field(
-        description="A list of license types for the geocoding service.",
-        default=None,
+    license_type: Optional[LicenseType] = Field(
+        description="The license type for the geocoding service.",
+        default="",
```
DanPaseltiner, line 48:
```suggestion
    license_type: Optional[license_types] = Field(
        description="The license type for the geocoding service.",
        default="",

```

DanPaseltiner (approved):
I think we should construct the list of licenses in the container and not make a change to the SDK (as discussed). Also, I think we can simplify our enum/Literal or license types.

DanPaseltiner (approved):

PR #371, opened 2023-03-31
Update mpi queries jsonpaths

m-goggins:
# Update MPI blocking queries to work on nested jsonb data

## Summary
This PR updates the Postgres MPI client to query within the `patient_resource` jsonb object of the `patient_table`. The updated query formatting allows users to specify which columns they want to block on (from a pre-determined list) and the transformations they want performed (e.g., for the DIBBs algo we block on last4 of MRN in the first pass). 

Importantly, the queries check against all values included in patient recources. For example, if a patient resource has multiple MRNs, the updated query generation will check against all values and return the resource as part of a block if one of the MRNs matches the incoming record's MRN. The same is true for given names.

## Related Issue
Fixes # 340

## Additional Information
Two important things to note:

1.   Blocked data are now returned as LoLoL to account for multiple values being stored in a single column. For example if a patient was named `John Danger Doe`, the old blocking function returned data like so:

  ```
  [ 
    ['first_name','last_name'],
    ['John','Doe']
  ]
  ```
  The new blocking function returns data like so, with more complete information:
  ```
  [ 
    ['first_name','last_name'],
    [['John','Danger'],['Doe']]
  ]
  ```
  
  I believe this will require us to change the way that the matching function operates, but we will be able to match on more complete name and address information.
  
  2. Removing whitespace when blocking on first 4 characters of address is not yet implemented.


phdi/linkage/postgres.py
```diff
@@ -153,26 +160,83 @@ def upsert_match_patient(
...
-        # TODO: Update queries for nested values, e.g., zip within address
-        query_stub = f"SELECT * FROM {table_name} WHERE "
-        block_query = " AND ".join(
-            [
-                f"patient_resource->>'{key}' = '{value}'"
-                if type(value) == str
-                else (f"'{key}' = {value}")
-                for key, value in block_data.items()
-            ]
+        fields_to_jsonpaths = {
```
emmastephenson, line 180: It looks like this dictionary is the same as the one above in `block_data` - can you extract it out so there's only one source of truth for both methods?

emmastephenson (commented):
This looks great Marcelle! Just one comment about DRYing it up, but this is excellent work!

phdi/linkage/postgres.py
m-goggins: Oh good call!

phdi/linkage/postgres.py
```diff
@@ -7,7 +7,7 @@
 class PostgresConnectorClient(BaseMPIConnectorClient):
```
DanPaseltiner, line 7: Could we change this name to reflect that this is a connector for the DIBBs MPI schema implemented on a Postgres DB? We would take our schema and use of flavors of SQL, additionally we could potentially need to work with a Postgres MPI that has a different schema in the future. I realize this is a little nit picky and not relevant currently, but if a simple name change now saves us a major potential head ache in the future I think it is worth doing.

phdi/linkage/postgres.py
```diff
@@ -28,6 +28,17 @@ def __init__(
...
+            "address": """$.address[*] ?(@.use=="home").line""",
+            "birthdate": "$.birthDate",
+            "city": """$.address[*] ?(@.use=="home").city""",
+            "first_name": """$.name[*] ?(@.use=="official").given""",
+            "last_name": """$.name[*] ?(@.use=="official").family""",
+            "mrn": """$.identifier ?(@.type.coding[0].code=="MR").value""",
+            "sex": "$.gender",
+            "state": """$.address[*] ?(@.use=="home").state""",
+            "zip": """$.address[*] ?(@.use=="home").postalCode""",
+        }
```
DanPaseltiner, line 41: Hell yeah! :metal:

bamader, line 41: 💪 🔥 🤮

phdi/linkage/postgres.py
```diff
@@ -58,31 +70,25 @@ def block_data(self, block_vals: Dict) -> List[list]:
...
+            raise ValueError("`block_vals` cannot be empty.")
 
         # Generate raw SQL query
-        query = self._generate_block_query(self.patient_table, block_vals)
+        query = self._generate_block_query(block_vals)
 
         # Execute query
         self.cursor.execute(query)
-        extracted_data = self.cursor.fetchall()
+        blocked_data = [list(row) for row in self.cursor.fetchall()]
```
bamader, line 80: Can you walk me through the rationale for making a list compression of list calls in this, vs the original fetchall()? Is it something with the type that the cursor returns?

phdi/linkage/postgres.py
```diff
@@ -153,26 +159,71 @@ def upsert_match_patient(
...
+        for col_name, param in block_vals.items():
+            # Add appropriate transformations
+            if "transformation" in param.keys():
+                # first4 transformations
+                if block_vals[col_name]["transformation"] == "first4":
+                    query = f"""
+                        CAST(jsonb_path_query_array(patient_resource,
+                        '{self.fields_to_jsonpaths[col_name]} starts with
+                        "{block_vals[col_name]["value"]}"') as VARCHAR)
+                        = '[true]'"""
```
bamader, line 209: 🤢 🤮

bamader (approved):
Seems good to me! Just a couple Q's but nothing blocking, just explanation.

DanPaseltiner (approved):
LGTM! One suggested (repeated) change for simplifying some our dict key accessing.

phdi/linkage/postgres.py
```diff
@@ -58,31 +70,25 @@ def block_data(self, block_vals: Dict) -> List[list]:
             raise ValueError(f"{error}")
 
         if len(block_vals) == 0:
-            raise ValueError("`block_data` cannot be empty.")
+            raise ValueError("`block_vals` cannot be empty.")
 
         # Generate raw SQL query
-        query = self._generate_block_query(self.patient_table, block_vals)
+        query = self._generate_block_query(block_vals)
```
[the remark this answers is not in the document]
m-goggins: You're right! I think it was a remnant of old code. Fixed it to just be the string

phdi/linkage/postgres.py
m-goggins: Something like `DIBBSMPIConnectorClient`? I think it's fine to change now given that I probably need to update the base class  documentation anyway to reflect some of the changes we've made

m-goggins: I called it `DIBBsConnectorClient` and also updated `upsert_match_patient` to `insert_match_patient`

PR #377, opened 2023-04-04
Make linkage work with MPI

### LINKED ISSUE #283: Create function to run record linkage algorithm on a single record
### Why are we doing this?
Until now, the record linkage algorithm has been run on groups of records instead of a single record, and matches identified between the group instead of between a single record and a database of patients. To run in production, the algorithm needs to be updated to run on a single record, since the pipeline will look at a single bundle at a time.

### Action Requested
Create a function to run record linkage on a single record.

### Acceptance Criteria
It's possible to pass a single record to the function and determine if it's a match with existing records in the MPI.

### Additional Information

- needs to accept all the usual algorithm specifiers, including blocking criteria, fields to match on, and overall decision rule (for log-odds scoring, the appropriate dictionaries as well)
- needs to work using the cluster membership part of the algorithm (i.e. given a block of data, put them into clusters based on their master Person IDs, and then check membership in each cluster formed)
- since the algorithm will consist of multiple passes/blocking rules, we can stop computation if we find a match in an early block (only the batch mode needs to keep going to make sure missing fields are still demographically recovered)
- we don’t know that one block will always correspond to a single Master Person resource, so we’ll also need to track which Person’s Patient records an incoming resource best matches to (example: we have an incoming record John Shepard born 11/7/1980, and the MPI has records John Sheplar born 11/7/1981 and John Shepston born 6/7/1990, each of whom belongs to a different Master Person; our algorithm’s second pass blocks on first 4 chars of first and last name, so both John Sheplar and John Shepston will be returned as potential candidates; our new John Shepard will partially match John Shepston (it’ll be a weak score, but the first name will match and both birthdate and last name will score at least a few points, but probably not many), but will have a better match with John Sheplar (closer birthdate and a last name that basically looks like a typo)
- return value is probably a boolean indicating whether the given record should belong in the cluster returned by the block

bamader:
# Linking against an MPI

## Summary
This PR introduces the functionality needed to link an incoming record against an existing MPI database. A number of helper functions are introduced to smooth this process, and the naming and documentation within the code should explain what each does (i.e. I'm actually going to try to provide less information here to make sure that the code and comments within sufficiently explain what's happening). I left a number of TODOs in the code as well that are known future iterations beyond the MVP.

## Related Issue
Fixes #283


phdi/linkage/link.py
```diff
@@ -327,36 +341,41 @@ def extract_blocking_values_from_record(
...
-        "zip": "Patient.address.postalCode",
-        "city": "Patient.address.city",
-        "state": "Patient.address.state",
-        "sex": "Patient.gender",
-    }
+    :param blocking_fields: A List of dictionaries giving the blocking
+      fields and any transformations that should be applied to them. Each
+      dictionary in the list should include a "value" key with one of the
+      supported blocking fields above, and may also optionally contain a
+      "transformation" key whose value is one of our supported tranforms.
```
emmastephenson, line 348:
```suggestion
      "transformation" key whose value is one of our supported transforms.
```

phdi/linkage/link.py
```diff
@@ -971,6 +1115,72 @@ def _eval_record_in_cluster(
...
+def _find_strongest_link(linkage_scores: dict) -> str:
+    """
+    Helper method that determines the highest belongingness level that an
+    incoming record achieved against a set of clusers based on existing
+    patient records in the MPI. The cluster with the highest belongingness
+    ratio is chosen as the Person to link the new record to.
+    """
+    max_membership = 0.0
+    best_person = None
+    for person_id in linkage_scores:
```
emmastephenson, line 1149: I think this function could be simplified down to:
```
best_person = max(linkage_scores, key=linkage_scores.get)
return best_person
```

tests/linkage/test_linkage.py
```diff
@@ -864,3 +883,169 @@ def test_algo_write():
         },
     ]
     os.remove("./" + test_file_path)
+
+
+def test_link_record_against_mpi():
```
emmastephenson, line 888: This test looks great, but it feels a bit more like an e2e or integration test than a unit test for the linkage test with the PostgresMPI implementation. Any chance there's a specific sub-folder or different file for those kinds of tests?

emmastephenson (commented):
This looks awesome Brandon! Just a few minor comments, but great work!

phdi/linkage/link.py
```diff
@@ -971,6 +1115,72 @@ def _eval_record_in_cluster(
...
+    """
+    flattened_record = [
+        extract_value_with_resource_path(resource, LINKING_FIELDS_TO_FHIRPATHS[f])
+        for f in sorted(LINKING_FIELDS_TO_FHIRPATHS.keys())
+    ]
+    flattened_record = [resource["id"], None] + flattened_record
+    return flattened_record
+
+
+def _group_patient_block_by_person(data_block: List[list]):
```
DanPaseltiner, line 1170: nit: missing return type hint.

DanPaseltiner (changes requested):
This is pretty righteous. I do think that we should pass in an MPI client instead of building it inside `link_record_against_mpi`. I don't think this function should be responsible for instantiating MPI clients given DB connection details, or selecting the appropriate client to use if we eventually support multiple MPIs in the future.

phdi/linkage/link.py
```diff
@@ -523,6 +542,116 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+    :param patient_table: The name of the Patient table in the MPI.
+    :param person_table: The name of the Person table in the MPI.
+    :returns: A tuple consisting of a boolean indicating whether a match
+      was found for the new record in the MPI, followed by the ID of the
+      Person entity now associated with the incoming patient (either a
+      new Person ID or the ID of an existing matched Person).
+    """
+    db_client = DIBBsConnectorClient(
+        database, user, password, host, port, patient_table, person_table
+    )
```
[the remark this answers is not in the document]
bamader: Yeah definitely agree here. I like the suggestion of not making it the responsibility of the linking function to instantiate a client, so I'll make it just accept one as a parameter.

phdi/linkage/link.py
```diff
@@ -523,6 +542,116 @@ def generate_hash_str(linking_identifier: str, salt_str: str) -> str:
...
+        field_blocks = extract_blocking_values_from_record(record, blocking_fields)
+        data_block = db_client.block_data(field_blocks)
+
+        # TODO: First row of returned block is column headers for eventual
+        # mapping of indices to cols. Right now, just ignore the first element.
+        data_block = data_block[1:]
+
+        # TODO: Currently, some fields may be LoL (name, address) since they
+        # can contain historical values. For the MVP, take only the first
+        # element as the one that should be used with the record.
```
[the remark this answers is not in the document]
bamader: Yep will do!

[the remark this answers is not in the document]
bamader: Lol I wound up thinking of a way to efficiently solve this problem too, so I noted that in the ticket. Turns out when you structure your code efficiently/modularized, you only have to change a small helper to get this result.

tests/linkage/test_linkage.py
emmastephenson: Agreed, I'm fine to punt on it for now. Would you mind adding a TODO in the code to move the test to an integration test suite later?

phdi/linkage/link.py
bamader: Woah didn't know max could take a key param!

DanPaseltiner (approved):
Looks like we are on the same page about a few minor tweaks we should make before merging. I'll approve now so I'm not blocking.

PR #383, opened 2023-04-04
Remove "use" filter in jsonb queries

m-goggins:
# Remove "use" filter in jsonb queries

## Summary
Because "use" is not mandatory in address and name fields for patient resources, we have to remove the filtering on "use".


DanPaseltiner (approved):
Send it! Thanks for the quick turn around on this @m-goggins

PR #390, opened 2023-04-05
Add context manager to postgres MPI implementation

### LINKED ISSUE #338: Add context manager to Postgres MPI
Currently, the Postgres MPI manually opens and closes a connection to the Postgres database within each method. It would be useful to [instead employ a context manager](https://github.com/CDCgov/phdi/blob/main/phdi/linkage/postgres.py#L44).

emmastephenson:
# PULL REQUEST

## Summary
* Adds a context manager to the Postgres MPI implementation
* DRY up some of the connection/cursor creation in unit tests

## Related Issue
Fixes #338 

## Additional Information
This PR is mostly moving code blocks around, which the Github diff-er doesn't show too well. Essentially in `postgres.py` I wrapped the connection and cursor in a try/with/finally structure, like this:

```
try:
  with psycopg2.connect(...) as db_connection:
      with db_connection.cursor() as db_cursor:
        # queries (unchanged)

except:
  # same exception/error handling

finally:
  # close connection and cursor

```


tests/linkage/test_postgres_mpi_connector.py
```diff
@@ -514,3 +408,25 @@ def test_insert_match_patient():
...
+
+
+def open_mpi_connection(postgres_client: DIBBsConnectorClient):
+    return psycopg2.connect(
+        database=postgres_client.database,
+        user=postgres_client.user,
+        password=postgres_client.password,
+        host=postgres_client.host,
+        port=postgres_client.port,
+    )
```
m-goggins, line 432: Not sure if you meant to leave this in.

m-goggins (approved):
Looks good (and much, much cleaner) to me! I only had one question about some functions at the end of the test file.

tests/linkage/test_postgres_mpi_connector.py
emmastephenson: Yep! These are meant to be helper functions for the tests, so that instead of each test passing in all the params to create a connector client or connection, they can just call these functions. But if there's a more Pythonic way to implement a helper function in tests like this I'm happy to update it!

DanPaseltiner (approved):
LGTM! The whole `with ... as` construct is definitely more pythonic.

PR #400, opened 2023-04-06
Update Dockerfiles to slim-bullseye

### LINKED ISSUE #364: Update slim 10 bullseye and check to see if security issue is resolved
### Why are we doing this?

Our Snyk vulnerability scanner highlighted an issue with libtasn1-6, a dependency imported from python3.10-slim-buster. We need to upgrade from the current version to address this vulnerability.

[See Snyk scan](https://app.snyk.io/org/prime-phdi/project/7d0199f0-4ed5-4535-8e8f-170900ffd312/).

### Action Requested

Attempt to upgrade from python:3.10-slim-buster to python:3.10-slim-bullseye, and see if that addresses the vulnerability.

If not, we may need to cherrypick the libtasn library to pin a version without the vulnerability.

### Acceptance Criteria

The vulnerability is no longer present in our Snyk reports.

emmastephenson:
# PULL REQUEST

## Summary
Update dockerfiles to slim-bullseye, as there's a vulnerability associated with slim-buster.

## Related Issue
Fixes #364 

## Additional Information
n/a

### DIFF
--- a/containers/alerts/Dockerfile
+++ b/containers/alerts/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim-bullseye
+FROM python:3.10-slim
 
 WORKDIR /code
 
--- a/containers/alerts/Dockerfile
+++ b/containers/alerts/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim-buster
+FROM python:3.10-slim-bullseye
 
 WORKDIR /code
 
--- a/containers/ingestion/Dockerfile
+++ b/containers/ingestion/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-slim-bullseye
+FROM python:3.10-slim
 
 WORKDIR /code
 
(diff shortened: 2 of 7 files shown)


nickclyde (commented):
We could also use `python:3.10-slim`, which would just default to the latest debian (bullseye) - what do you think?

emmastephenson:
> We could also use `python:3.10-slim`, which would just default to the latest debian (bullseye) - what do you think?

@nickclyde sure! The only question I have with that approach is if we've ever hit issues that needed a manual fix during a debian update before. This one didn't need anything other than changing the version, but if manual tweaks have been required in the past I'd prefer to upgrade on our own schedule to be on the safe side. But if that's not a concern, then I'm happy to use `slim` directly!

nickclyde:
Debian releases every 2 years so we haven't had a new release yet. I wouldn't expect new versions to have much of an impact though. Especially since these are slim images without many packages in them by default. Any breaking changes would probably be very minor and easy to fix!

emmastephenson:
> Any breaking changes would probably be very minor and easy to fix!

Famous last words if I ever heard them 🤣 But agreed, given all that I'm happy to switch to `python:3.10-slim`.

nickclyde (approved):

nickbristow (approved):

PR #405, opened 2023-04-07
Remove first element limitation from linkage match

### LINKED ISSUE #386: Remove first element limitation from linkage match

m-goggins:
# Remove first element limitation from linkage match

## Summary
Previously, when matching we could only compare the 1st element of a list. This is not ideal for data that may have multiple valid values, e.g., address and name. The new functions allow the user to compare multiple values from an incoming record's column to multiple values stored in an MPI.

## Related Issue
Fixes #386 

## Additional Information
There are 2 helper functions that have not yet been integrated into the algorithm/`link_record_against_mpi`. I held off on integrating as it may require more work around replacing column indices with names, e.g., do regular matching functions if not address or name column, etc. 

We need two helper functions because matching of names is order dependent and other data, like address, are not. That is, you want to make sure you match name on "First Name, Middle Name" and not "Middle Name, First Name" while for address you only care that an incoming record's addresses matches _any_ of the addresses in the MPI patient data.


tests/linkage/test_linkage.py
```diff
@@ -1076,3 +1078,115 @@ def test_add_person_resource():
...
+    different_address = _compare_address_elements(
+        record=record, mpi_patient=mpi_patient2, feature_func=feature_funcs, x=x
+    )
+    assert different_address is False
+
+
+def test_compare_name_elements():
+    feature_funcs = {0: feature_match_fuzzy_string}
+    x = 0
+    record = [
```
emmastephenson, line 1136: Out of scope for this PR, but I'm curious if the idea of a test data generator has been considered for this project? I know unit testing in Python is thorny so implementing it may be a pain, but variable duplication like this kills me a little 😅

phdi/linkage/link.py
```diff
@@ -1122,6 +1122,50 @@ def _compare_records(
...
+    **kwargs,
+) -> bool:
+    """
+    Helper method that compares all elements from the flattened form of an incoming
+    new patient record to all elements of the flattened patient record pulled from
+    the MPI.
+    """
+
+    for r in record[2:][x]:
+        for m in mpi_patient[2:][x]:
```
emmastephenson, line 1139: Also out of scope for this PR, but it's things like indexing via `[x]` that make me want an ORM 😄

emmastephenson (approved):
This looks great, thanks Marcelle!

I continue to be a grumpy recovering Java dev but please don't take it personally 😆

tests/linkage/test_linkage.py
m-goggins: Yeah, I would be in favor of a test data generator. The assets folder is also difficult to navigate. Dan and I briefly discussed working on test data after LAC because some of our test data, e.g., patient resources, are malformed.

PR #421, opened 2023-04-12
Reference postgres vars properly

bamader:
# Postgres vars fast fix

## Summary
If the connection to the MPI is refused or otherwise errors out, the python code proceeds to the finally block without having correctly initialized the `db_cursor` or the `db_connection`, leading to a `variable referenced before assignment` error (which I saw in my linkage endpoint PR). This PR corrects that by defaulting each of the variables to `None` before its invoked so that the finally block can proceed correctly

## Related Issue
Fixes not a github issue, but a bug I found

### DIFF
--- a/phdi/linkage/postgres.py
+++ b/phdi/linkage/postgres.py
@@ -59,6 +59,8 @@ class DIBBsConnectorClient(BaseMPIConnectorClient):
             raise ValueError("`block_vals` cannot be empty.")
 
         # Connect to MPI
+        db_cursor = None
+        db_connection = None
         try:
             with psycopg2.connect(
                 database=self.database,
@@ -80,8 +82,10 @@ class DIBBsConnectorClient(BaseMPIConnectorClient):
 
         finally:
             # Close cursor and connection
-            db_cursor.close()
-            db_connection.close()
+            if db_cursor is not None:
+                db_cursor.close()
+            if db_connection is not None:
+                db_connection.close()
 
         # Set up blocked data by adding column headers as 1st row of LoL
         # TODO: Replace indices with column names for reability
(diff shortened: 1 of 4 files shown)


DanPaseltiner (changes requested):
Definitely see the problem we are trying to address here. Based on our conversation I think we should tackle this by refactoring instantiation of our connector objects into a method with error handling. This will help:

1. DRY out the code
2. Allow us to test our MPI connection more easily
3. Ensure we fail fast in the event that a connection to the MPI cannot be established

phdi/linkage/postgres.py
```diff
@@ -1,15 +1,22 @@
...
+    Represents a Postgres-specific Master Patient Index (MPI) connector
+    client for the DIBBs implementation of the record linkage building
+    block. Callers should use the provided interface functions (e.g.,
+    block_vals) to interact with the underlying vendor-specific client
+    property.
+
+    When instantiating a DIBBSConnectorClient, the connection to the
+    database is automatically tested using the provided parameters.
+    A refused, timed-out, or otherwise error-ed connection will raise
+    an immediate exception.
```
DanPaseltiner, line 19: 🚀

DanPaseltiner (commented):
This looks awesome, nice touch to add _close_connections() !

Since we are doing a better job handling the connections do we still need to set `db_cursor` and `db_conn` to `None`? Maybe I'm missing something, but I think we have handled that reference before assignment issue you mentioned earlier.



[1:44](https://skylight-hq.slack.com/archives/D034L91UL1Y/p1681321445748769)
I’m still not grokking why we need
        db_cursor = None
        db_conn = None





[1:44](https://skylight-hq.slack.com/archives/D034L91UL1Y/p1681321445748769)
I’m still not grokking why we need
        db_cursor = None
        db_conn = None

DanPaseltiner (approved):
I understand now. We are still defining the connection and cursor within the try/except block. Thanks for your work on this!

PR #425, opened 2023-04-12
Convert parquet to fhir

### LINKED ISSUE #415: Create FHIR resources out of MPI seed data
### Why are we doing this?

The MPI seed data that LAC will provide each week takes the form of a Parquet data dump. To add this information as records in the MPI, we'll need to convert the information within to Person/Patient FHIR records.

### Action Requested

Create a conversion script that takes the data within the Parquet file and creates FHIR Patient records.

### Acceptance Criteria

The MPI seed data can be transformed into separate FHIR bundles, which can in turn be inserted into the MPI.

### Additional Context

We do not yet know if this data will be Person or Patient records (i.e., has it already be deduplicated?). Once that information has been gathered from LAC, this ticket will be updated.

Related work: #414 

Also related: #411. Unlike 411 however, this ticket is specifically about the data conversion, not setting up a Synapse job.

m-goggins:
# Convert patient data from parquet to FHIR patient resources

## Summary
This PR creates SDK functions and a script that eventually will be used in Synapse to read in a parquet file of patient data, convert each row of patient into a patient resource, and return a tuple of `(iris_id, patient_resource)` that can hit the RL endpoint as part of seeing the MPI.

The script will likely move to the `phdi-azure` repot. When Nick C has Synapse up and running, he and I will coordinate.

## Related Issue
Fixes #415 


phdi/linkage/seed.py
```diff
@@ -0,0 +1,77 @@
...
+    generated id as well as a the `iris_id`.
+
+    :param data: Dictionary of patient data that optionionally includes the following
+      fields: mrn, ssn, first_name, middle_name, last_name, home_phone, cell-phone, sex,
+      birthdate, address, city, state, zip.
+    :return: Tuple of the `iris_id` and FHIR patient resource
+    """
+
+    # Iterate through each patient and convert patient data to FHIR resource
+    patient_resource = {
```
emmastephenson, line 18: For creation of new FHIR resources, could we use a library like [`fhir.resources`](https://pypi.org/project/fhir.resources/#:~:text=Example%3A%203%3A%20This%20example%20creates%20a%20new%20Patient%20resource%20from%20json%20object(py%20dict)%3A)? I'm hesitant to hardcode the resource like this.

emmastephenson (changes requested):
This is a great start!

phdi/linkage/seed.py
m-goggins: I didn't know that this exists! I did some testing with it. Some things to note:

 - This allows us to have a fully formed Patient resource even if we only feed it partial information. Most elements will just have `None` values.
- We have to do a bit of massaging to get the row data in the correct format so that we can pass it into the Patient resource. the resource wouldn't be hard coded though. Basically, we have to put the into a dictionary like the one created in the original script, e.g., 
```
{
    "resourceType": "Patient",
    "id": str(uuid.uuid4()),
    "identifier": [
        {
            "type": {
                "coding": [
                    {
                        "system": """http://terminology.hl7.org
                                /CodeSystem/v2-0203""",
                        "code": "MR",
                    }
                ]
            },
            "value": f"{data.get('mrn',None)}",
        },
        {
            "type": {
                "coding": [
                    {
                        "system": """http://terminology.hl7.org
                                /CodeSystem/v2-0203""",
                        "code": "SS",
                    }
                ]
            },
            "value": f"{data.get('ssn',None)}",
        },
    ],
    "name": [{"family": f"{data.get('last_name',None)}"}],
}
```

Overall, I think this is a good improvement and could be useful for refactoring other testing assets after the pilot 😄 I will implement it here as well.

examples/MPI-sample-data/build_synthetic_mpi_parquet.py
```diff
@@ -0,0 +1,132 @@
...
+        "home_phone",
+        "cell_phone",
+    ]
+]
+
+# Save
+df.to_parquet(
+    "./examples/MPI-sample-data/synthetic_patient_mpi_seed_data.gzip",
+    compression="gzip",
+)
```
BradySkylight, line 132: This is very awesome to have!!  I love how much you have been working with and integrating synthetic data for all our testing!  Nice work Marcelle!

BradySkylight (approved):
With the changes you will make based upon Emma's feedback, I think this LGTM.  Excellent work!

phdi/linkage/seed.py
m-goggins: Hmm upon further review, I am not sure this will actually suit our needs, but maybe you can help me think this through. This resource, which has a special type `<class 'fhir.resources.patient.Patient'>`, gets added to a bundle which is then passed to the RL endpoint. The RL endpoint is expecting a `dict`. When I try to convert the `<class 'fhir.resources.patient.Patient'>` to `dict`, it doesn't appear to correctly convert to a dict. 

For example, the `name` section should look like:
```
"name" : {
      "family" : "du Marché",
      "_family" : {
        "extension" : [{
          "url" : "http://hl7.org/fhir/StructureDefinition/humanname-own-prefix",
          "valueString" : "VV"
        }]
      },
      "given" : ["Bénédicte"]
    }
```

but using `fhir.resources` results in the `name` section looking like:
```
HumanName(resource_type='HumanName', fhir_comments=None, extension=None, id=None, family='Kilback', family__ext=None, given=['Delmar', 'Middle'], given__ext=None, period=None, prefix=None, prefix__ext=None, suffix=None, suffix__ext=None, text=None, text__ext=None, use=None, use__ext=None)
```

emmastephenson: Hmmm I was worried about this - seems like unwrapping back to `dict` isn't fully unwrapping all the nested fields. Let's chat offline?

m-goggins: Based on offline conversation, we're going to hold off on implementing `fhir.resources`. While it does have some functionality that converts the resource to JSON format, doing so does not add anything different from what is currently implemented. We decided to create a tech debt ticket to investigate how this might be used in the future, especially as it relates to assets and testing.

BradySkylight: Yeah, it seems that Python is lacking in their FHIR packages.  This was a simple implementation in Java from my previous experience.  It seems they are treating each data type as a resource.  SMH

m-goggins:
One thing to note: If LAC's data does not conform to the FHIR standard, this code won't correct for it. For example, if the `gender/sex` column is in a format other than `male` or `female`. It will be easy to modify the data on the fly before putting it into the bundle but just something to flag.

emmastephenson (approved):
LGTM, thanks Marcelle!

PR #448, opened 2023-04-18
Update FHIR Converter to use ecr directory and eicr template for ecr messages

BradySkylight:
# PULL REQUEST

## Summary
Updated the Observation template to use extensions within extension for specimen information.  Also updated the FHIR Converter to use the ECR directory and the EICR template for ECR messages.  Updated tests accordingly. 

## Related Issue
N/A


DanPaseltiner (approved):
LGTM

PR #453, opened 2023-04-18
Match on multiple elements

### LINKED ISSUE #406: Integrate multiple element matching
### Why are we doing this?
Now that we have helper functions for matching on [multiple elements](https://app.zenhub.com/workspaces/dibbs-63f7aa3e1ecdbb0011edb299/issues/gh/cdcgov/phdi/386), we need to integrate them into the `link_record_against_mpi` function, specifically in the `_compare_records` helper function.

### Action Requested
Add multiple element comparison functions into the record linkage pipeline.

### Acceptance Criteria
Only columns that have multiple elements for comparison, e.g., address and given name, are compared against multiple values and not just the 1st element. This may require us to switch to indexing on column names, e.g., "address" instead of 0.

bamader:
# Multiple Element Matching

## Summary
This PR introduces functionality necessary to link records using multiple pieces of information in the name and address fields of a FHIR bundle. We know this is important for linkage, since people can move or otherwise change addresses, as well as marry or otherwise change their name (or provide different names in different cases). FHIR bundles store this information as additional `line` entries within the associated objects of the `name` and `address` fields (these are the only fields which contain lists, rather than raw strings).

Previously, our algorithm looked at only the first piece of data in each of these fields, which could lead to large numbers of missed matches once deployed to production. Now, though, with the helper functions we have written, this code enables us to orchestrate linkage using all of the variable information in a patient. We accomplish this by selectively concatenating the information in the `line` properties of the two fields, preserving separation between distinct objects; for name, we consider the fields a full match if the full concatenated strings of both an incoming record and an MPI patient agree, and for address, we consider the fields a full match if the line concatenation of a patient fuzzy matches _at least one_ concatenated string from the MPI records.

## Related Issue
Fixes #406 

## Additional Information
While it might seem counterintuitive to treat name and address differently, in this case, we believe there are valid logical reasons to do so. FHIR data, and our MPI in particular, doesn't give us any temporal ordering on the information it stores. I.e. we have no way to say which of a patient's two addresses is their current address and which is their newest. We've set up our JSONB querying so that this shouldn't be a problem, though, because we currently compare extracted blocking values to _each_ object in the relevant field, to see if the block would be a match to any previously seen data. This allows us to process records without needing to receive them in the "correct" chronological order (i.e. we could have a patient in our MPI who's moved a couple times, get a FHIR bundle containing info from one of the patient's old addresses, and still link them correctly).


phdi/linkage/link.py
```diff
@@ -1115,17 +1125,37 @@ def _compare_records(
...
 
+def _compare_records_field_helper(
+    record: List,
+    mpi_patient: List,
+    idx: int,
+    idx_to_col: dict,
+    feature_funcs: dict,
+    **kwargs,
+) -> bool:
+    if idx_to_col[idx] == "first_name":
```
emmastephenson, line 1145: Can I ask why we're not also comparing multiple elements for last name? To me that seems the more likely name to change 🤔

BradySkylight: Actually this  is a good question.  So, in FHIR a Patient.name is a datatype of HumanName.  A given name can be the first name, but the middle name is also a 'given' name so both are included in that field as a list.  The family name is just a single string.  However, a Patient resource can have 0..* (none or multiple) 'names'/HumanNames.  This is where, in the case of a maiden and current, name for family name could come into play.  The Patient may have two family names.  The Patient.name.period.start and Patient.name.period.end are not required and Patient.name.period is not required either.  This is what that field is for, however, to denote 'when' the name was/is used, but this isn't always supplied and most likely we can't rely on having this data available.  We should gather all names and all addresses so they can be compared for matches.

BradySkylight: Hopefully, this makes sense.  I can go into more detail with you all tomorrow if you'd like.

bamader: Yeah originally this was there to just flatten out all the different "given" names that could exist within one instance of a HumanName in a patient (in particular, we were thinking of the convention many Hispanic and Latin names follow of multiple middle names, which we wanted to make sure we captured). I can see what you're saying about wanting to extend that to the last name, too, but I think there are some tricky things involved with making sure we don't match in a way that gives false positives.

E.g., with first/given names, we rely on fuzzy matching on a giant concatenation of all given names a patient has had, because our testing shows that this can account for things like including more/fewer names in some messages but also excluding people that might match on a small subset of middle names but not first names. For addresses, we say it's a match if the incoming address fuzzy matches _any_ of the addresses an MPI patient might be storing, but that works because when you move address changes completely. The key thing about both of these fields, though, is that they can be lists in the FHIR bundle. For HumanName and Address object, both `given name` and `line` are lists that can hold multiple values. So when we concatenate this information to match on a larger scale, we have a lot more data to fuzzy match against for veracity. My worry with using either of these approaches with last name is that because each `family` name in a HumanName object is a single string, the "concatenated" version of all family names a patient might have had will be much smaller, which will make any deviations much harsher in the fuzzy match scoring. I don't know if that would qualify too many people for false positives or disqualify too many people and cause false negatives, but it's definitely something whose handling is less clear than first name or address. @m-goggins curious to hear what you think about this breakdown?

phdi/linkage/link.py
```diff
@@ -581,21 +581,29 @@ def link_record_against_mpi(
...
+        # contains extracted values, so minimally block on the first line
+        # if applicable
         field_blocks = extract_blocking_values_from_record(record, blocking_fields)
         data_block = db_client.block_data(field_blocks)
 
-        # TODO: First row of returned block is column headers for eventual
-        # mapping of indices to cols. Right now, just ignore the first element.
+        # First row of returned block is column headers
+        # Map idx to column name, not including patient/person IDs
+        idx_to_col = {k: v for k, v in enumerate(data_block[0][2:])}
```
m-goggins, line 593: Nice!

phdi/linkage/link.py
m-goggins: Yeah, I agree with Brandon on treating family name as a string and only using 1 value. If I recall correctly, Jaro-Winkler also slightly favors strings that match from the beginning (e.g., Johnson Smith and Johnsen Smith) because it assumes order matters. With the case that Brandon presents above of the last names being somewhat randomly concatenated together, we lose the importance of order.

Bigger picture, I think we're okay with the implementation as is. Given the context of this record linkage MVP, i.e., COVID-only, I think it is more likely that we will only see a single `HumanName` and a single `address` as opposed to a full history. Or at least, that's what Dan suspected when I talked to him about the address field a few weeks ago.

m-goggins: Another thing we could do, post-LAC, is test the filtering on name use == 'official'. Again, I don't think we should do anymore tinkering with the name functions right now but it could be worth exploring if linking on prod data doesn't look good or we have time to tinker with the underlying algorithm (more stats 😅!)

tests/assets/patient_bundle_to_link_with_mpi.json
```diff
@@ -130,6 +134,22 @@
...
+                        ],
+                        "use": "official"
+                    },
+                    {
+                        "family": "Nar Raya",
+                        "given": [
+                            "Tali",
+                            "Zora"
+                        ],
+                        "use": "official"
```
m-goggins, line 152: If I am understanding the current implementation, we would only be using the "Vas Neema" family name for blocking/matching? I may have missed it, but do we have a test to show what happens when blocking on this scenario?

m-goggins (approved):
Looks good to me! I especially like the simple fix for mapping idx to column name; much cleaner. 

I added some comments about the given name matching and had a question about how we might test the scenario that Emma's comment brought up (multiple `HumanNames`).

tests/assets/patient_bundle_to_link_with_mpi.json
bamader: This is a great catch. You didn't miss it, so I've added a test that now shows that we can successfully retrieve patients from the MPI when blocking on either/any of the family names a patient might hold in their array of HumanNames (this also supports what we talked about on slack around how the JSONB querying would return stuff, so it's nice to get that confirmation 😅 ).

BradySkylight (approved):
Based upon our slack conversation, this looks good to me.  Cool Python going on in this PR!  I think my only question would be is around the use of 'random' in our selection criteria (do we still need that?).

PR #477, opened 2023-04-20
Add additional checks and bolstering for collectTime

BradySkylight:
# PULL REQUEST

## Summary
There are times when the collectTime is not part of the Specimen portion of a lab observation within and ECR message.  There were a few messages that were throwing errors when trying to convert them to FHIR related to this.  I have added additional logic to store 'None' for the collectTime variable when it's not present in the ECR message and a check to not include that portion of the FHIR message if collectTime is None.

## Related Issue
Fixes #


emmastephenson (approved):
Nice catch!

PR #501, opened 2023-04-24
Removing datetime that is added at the end of the filename

KennethSkylight:
Removes the datetime string that appends to the file name in the write to storage function. It didn't serve a meaningful purpose, so we are replacing it wtih runid in a different branch


DanPaseltiner (approved):
LGTM

PR #513, opened 2023-04-26
Generate new UUID for patient resources.

DanPaseltiner:
# PULL REQUEST

## Summary
This PR ensures that the FHIR converter gives every patient resource unique UUID.

## Additional Information
Although we previously assumed that every "UUID" generate by the FHIR converter was unique, recent testing has proved that this is not the case. This is an issue because the current assumption of our record linkage approach is that every patient resource will have a unique ID which is causing end to end testing failures.

### DIFF
--- a/containers/fhir-converter/app/main.py
+++ b/containers/fhir-converter/app/main.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 import subprocess
 import json
-
+import uuid
 from enum import Enum
 from fastapi import FastAPI, Response, status
 from pydantic import BaseModel
@@ -173,6 +173,13 @@ def convert_to_fhir(
     # Process the response from FHIR Converter.
     if converter_response.returncode == 0:
         result = json.load(open(output_data_file_path))
+        # Generate a new UUID for the patient resource.
+        for entry in result["FhirResource"]["entry"]:
+            if entry["resource"]["resourceType"] == "Patient":
+                new_id = str(uuid.uuid4())
+                entry["fullURL"] = f"urn:uuid:{new_id}"
+                entry["resource"]["id"] = new_id
+                break
     else:
         result = vars(converter_response)
         # Include original input data in the result.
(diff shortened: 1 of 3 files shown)


containers/fhir-converter/app/main.py
```diff
@@ -173,6 +173,13 @@ def convert_to_fhir(
     # Process the response from FHIR Converter.
     if converter_response.returncode == 0:
         result = json.load(open(output_data_file_path))
+        # Generate a new UUID for the patient resource.
+        for entry in result["FhirResource"]["entry"]:
+            if entry["resource"]["resourceType"] == "Patient":
+                new_id = str(uuid.uuid4())
+                entry["fullURL"] = f"urn:uuid:{new_id}"
+                entry["resource"]["id"] = new_id
+                break
```
BradySkylight, line 182: Is there anywhere else in the bundle where we may need to change this?  Maybe also here in the section of request that occurs AFTER the resource:  

`},
      "request": {
        "method": "PUT",
        "url": "Observation/1863c018-d142-29c1-2c97-1ef39472a14c"
      }`

BradySkylight: Full example of Observation

{
      "fullUrl": "urn:uuid:1863c018-d142-29c1-2c97-1ef39472a14c",
      "resource": {
        "resourceType": "Observation",
        "id": "1863c018-d142-29c1-2c97-1ef39472a14c",
        
      },
      "request": {
        "method": "PUT",
        "url": "Observation/1863c018-d142-29c1-2c97-1ef39472a14c"
      }
    },

BradySkylight (changes requested):
Make sure all requests and references of the patient are updated... ;)

containers/fhir-converter/app/main.py
DanPaseltiner: oh yep. Good catch. Thank you!

PR #516, opened 2023-04-28
Move migration function call to its own endpoint for use in deploymen…

gordonfarrell:
…t workflow

# PULL REQUEST

## Summary
Removing migration function call from health check endpoint and record-linkage endpoint and moving it instead to its own endpoint which we can hit during the deployment process to engage the migrationification process.

## Related Issue
Half of the fix for #432 , once this is merged I'll open a PR with the deployment workflow changes in phdi-azure
https://app.zenhub.com/workspaces/dibbs-63f7aa3e1ecdbb0011edb299/issues/gh/cdcgov/phdi/432

## Additional Information
I feel a little weird about having a run migrations endpoint, if anyone else has great ideas on how to run these once on deployment feel free to chime in!

### DIFF
--- a/containers/record-linkage/app/main.py
+++ b/containers/record-linkage/app/main.py
@@ -152,7 +152,7 @@ async def health_check() -> HealthCheckResponse:
 
 
 @app.get("/run-mpi-migrations")
-async def run_mpi_migrations() -> HealthCheckResponse:
+async def run_mpi_migrations():
 
     try:
         run_migrations()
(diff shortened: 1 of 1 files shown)


DanPaseltiner:
I'm concerned about this approach too. I'd love to get this function to run anytime the service spins up. Requiring users to hit a special endpoint to ensure the MPI is available and configured appropriately seems awkward. I realize I'm not providing and alternative solution here, but I think it is worth thinking about this a little more.

gordonfarrell:
Not sure why its not linking the PR for this automatically, its still stuck on the first one which was using a different approach we decided not to pursue.

This issue is fixed in https://github.com/CDCgov/phdi/pull/518

PR #517, opened 2023-04-28
New PHDI Release Workflow

### LINKED ISSUE #434: Add versioning to container images
### Why are we doing this?
Currently our containers all read from the `main` tag. This is a very brittle process, as an API update could unwittingly break live containers. Instead, container images should be versioned.

### Action Requested

- Create a tagged `release` branch with the container image versions that LAC should use.
- Create an automated workflow in Github that we can manually run when a new release is ready that will do the following:
-- Run all tests to verify release is ready with no failing tests
-- Create a new Tagged Release
-- Rebuild containers with the new tagged release

### Acceptance Criteria

- Containers no longer point at `main`, but instead to a specific release version.
- A stable release is created
- A Github workflow is ready to test, tag, build

The containers deployed to LAC should point to a stable version.

### Additional Information
n/a

BradySkylight:
# PULL REQUEST

## Summary
We need to be able to version our releases.  This version needs to be part of a tag, release in our GitHub repo, as well as tagging the Packaged Containers, with the same tag/release, in our repo.  This version number also needs to be carried over into the project 'PHDI's version number (init.py), but also should be in the .toml file as well as the version number specified in the FastAPI version within our containers.  This workflow accomplishes all that and more. This workflow will be kicked off manually when we are ready to 'cut' a release.

I have also included a release.yaml that should auto generate Release Documentation within our Release in the REPO when we cut a new release.

I have also updated the alerts container to use the base class to ensure that we can get the version number into the internal containers that way.  Robert is currently working on updating the rest of the containers to do the same.  Once they are all changed like this, then all containers will have the proper internal version number associated.

Work flow steps:

1. Request version number (ie. v1.0.2) which is manually entered.
2. Run all tests.
3. If tests pass then create a tag, based upon the manually entered version number, update the .toml file for the project, which also automatically updates all other internal version numbers for PHDI and the containers.  This step also creates a new release for said tag.
4. Rebuild all containers and store them as packages in our repo with the 'latest' tag as well as the version specified tag.
5. Updates the docs branch with the latest PHDI SDK Documentation

## Related Issue
Fixes #434 

## Additional Information
There are other efforts in progress, that can be added to this workflow, to update the container documentation and store that in the docs branch.  Also, we would like to release the latest version of PHDI SDK to Pypi.  Both of these efforts are almost done and have been a collab. between myself, Nick C., and Kenneth.


nickclyde (approved):
Amazing work Brady! Glad we were able to get this all sorted out together today 🚀

tests/not_valid_json_test.json
```diff
@@ -0,0 +1 @@
+this is a random string that is not in json format
```
emmastephenson, line 1: This doesn't seem to be used in this PR?

containers/alerts/requirements.txt
```diff
@@ -2,6 +2,7 @@ azure-communication-identity
 azure-communication-phonenumbers
 azure-communication-sms
 azure-identity
+phdi @ git+https://github.com/CDCgov/phdi.git@main
```
emmastephenson, line 5: Should we tag a specific release here, instead of `main`?

emmastephenson (approved):
Looks great to me! Excited to get this working!

containers/alerts/requirements.txt
DanPaseltiner: Yeah, we should probably have the release workflow automagically target the version that we are making a release for.

DanPaseltiner (changes requested):
This is so exciting to see this coming together! Thank you @BradySkylight! I think we should address @emmastephenson's comment about targeting phdi versions in the service requirement.text files. Also, we aren't currently building docs for the container APIs, just the SDK. I'm open to adding support for these docs in a follow up PR, but we should probably have the functionality in place before making our first release with this workflow.

.github/workflows/createNewRelease.yaml
```diff
@@ -0,0 +1,85 @@
...
+          pip install poetry
+          poetry install
+
+      - name: Generate docs and move to docs branch
+        run: |
+          poetry run pdoc ./phdi -o ./docs/${{ github.event.inputs.custom_tag }}
+          git checkout docs --
+          rm -rf ./docs/latest
+          cp -r ./docs/${{ github.event.inputs.custom_tag }} ./docs/latest
+
```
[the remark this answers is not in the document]
nickclyde: I have a branch going for this!
https://github.com/CDCgov/phdi/tree/nick/release-container-docs

containers/alerts/requirements.txt
nickclyde: I added this into my branch:
https://github.com/CDCgov/phdi/commit/9f4a1a5fa7c782019799df941c310bdb29713cbe

KennethSkylight (approved):
Excited for this one!

PR #520, opened 2023-04-29
Replace FastAPI class with PHDI BaseService for container apps

robertmitchellv:
# PULL REQUEST

## Summary
For most of the container apps' `main.py` there is boilerplate metadata being passed to `FastAPI`. In an effort to be more DRY, `phdi.containers.base_service` offers a `BaseService` class to handle the redundancies.

## Related Issue
Fixes ? (not sure if there is already an issue for this)


robertmitchellv:
Still working on `tabulation` and `validation`

robertmitchellv:
wrapped up everything (except FHIR server; have a question on that one for Dan when he's back)

emmastephenson (approved):
This looks great, Robert! Thanks!

DanPaseltiner (changes requested):
@robertmitchellv this is looking really great! Thanks for your work here. Before merging I think we can simplify some of our unit testing. The services that use the `/health-check` endpoint from `base_service` no longer need to have that endpoint covered by the service specific unit test since we have unit test coverage for this endpoint in `phdi`. Could these, unit tests that are now redundant be removed as part of this PR?

robertmitchellv:
Thanks Dan; yes, absolutely, I can work on that this morning.

DanPaseltiner (approved):
Thanks for removing those tests!

robertmitchellv:
Tests are passing on local--I built the docker container for the building block locally to see what was going on--looks like a shell command is trying to remove something that isn't there:

```bash
Step 5/20 : RUN rm src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid
 ---> Running in 09851ded28aa
rm: cannot remove 'src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid': No such file or directory
The command '/bin/sh -c rm src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid' returned a non-zero code: 1
```

Will work on fixing this error.

BradySkylight (approved):
This looks great!  I'm so glad we are doing this with all the containers.  Excellent work Robert!

BradySkylight (approved):
This looks great!  I'm so glad we are doing this with all the containers.  Excellent work Robert!

PR #522, opened 2023-05-02
Add filter to MRN values

### LINKED ISSUE #509: Add filter for MRN
### Why are we doing this?
In the DIBBs record linkage algorithm, we block on the last 4 digits of MRN. Currently, when we query the MPI and the incoming record for the MRN, we are looking at the identifier fields, but not selecting for MRN values in particular. As such, an SSN (for example) could be pulled for comparison instead of the MRN.

### Action Requested
On the MPI querying side, re-add the coding filter that was removed accidentally. See line 38 [here](https://github.com/CDCgov/phdi/commit/a9db865a98b4badd15f4055d9c0faf1ad0dc0f0c#diff-2c559bcf0d8add0a7a0353bd1fac9fe82c559db19d0cc73476b0ce10811375d5L38) for how the code should be changed. 

On the incoming record side, the MRN in `LINKING_FIELDS_TO_FHIRPATHS` within `link.py` needs to be changed to something like `identifier.where(coding= 'MR').value` so that it returns only MRNs.

### Acceptance Criteria
Only MRN values are returned when querying the identifier fields on both the MPI and incoming record.
### Additional Information

bamader:
# Correctly filter on identifier

## Summary
This PR corrects a bug in our JSON path querying and FHIR Path accessing where we were grabbing _any_ identifier the patient had, rather than MRN specifically. Now, we correctly filter on both the MPI query side as well as the incoming patient / flattening side for exclusively MRN.

## Related Issue
Fixes #509

### DIFF
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -24,7 +24,7 @@ LINKING_FIELDS_TO_FHIRPATHS = {
     "city": "Patient.address.city",
     "state": "Patient.address.state",
     "sex": "Patient.gender",
-    "mrn": "Patient.identifier.value",
+    "mrn": "Patient.identifier.where(type.coding.code='MR').value",
 }
 
 
--- a/phdi/linkage/postgres.py
+++ b/phdi/linkage/postgres.py
@@ -42,7 +42,7 @@ class DIBBsConnectorClient(BaseMPIConnectorClient):
             "city": """$.address[*].city""",
             "first_name": """$.name[*].given""",
             "last_name": """$.name[*].family""",
-            "mrn": """$.identifier.value""",
+            "mrn": """$.identifier ?(@.type.coding[0].code=="MR").value""",
             "sex": "$.gender",
             "state": """$.address[*].state""",
             "zip": """$.address[*].postalCode""",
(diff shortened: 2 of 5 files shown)


phdi/linkage/link.py
```diff
@@ -24,7 +24,7 @@
     "city": "Patient.address.city",
     "state": "Patient.address.state",
     "sex": "Patient.gender",
-    "mrn": "Patient.identifier.value",
+    "mrn": "Patient.identifier.where(type.coding.code='MR').value",
```
DanPaseltiner, line 27: 🚀

phdi/linkage/postgres.py
```diff
@@ -43,7 +43,7 @@ def __init__(
             "city": """$.address[*].city""",
             "first_name": """$.name[*].given""",
             "last_name": """$.name[*].family""",
-            "mrn": """$.identifier.value""",
+            "mrn": """$.identifier ?(@.type.coding[0].code=="MR").value""",
```
DanPaseltiner, line 46: 🚀

DanPaseltiner (approved):
Send it! Thanks @bamader

m-goggins (approved):
Looks good!

PR #526, opened 2023-05-03
Add additional param to the BaseService class to allow for passing other licenses to FastAPI

robertmitchellv:
# PULL REQUEST

## Summary
Added a `license_info` param that is a `dict` with a default value that matches all DIBBs defaults for the container apps in `phdi/containers`. Given the `fhir-converter` is pulling a Microsoft base image in its Dockerfile and using [Micosoft's FHIR Converter](https://github.com/microsoft/FHIR-Converter) we should make a best effort to pass its MIT license and add it to the `description.md` for the `fhir-converter`, which is already done in #520 

## Related Issue
Fixes #520


tests/containers/test_base_service.py
```diff
@@ -3,25 +3,48 @@
...
+default_app_version = metadata.version("phdi")
+default_app_contact = {
+    "name": "CDC Public Health Data Infrastructure",
+    "url": "https://cdcgov.github.io/phdi-site/",
+    "email": "dmibuildingblocks@cdc.gov",
+}
+default_app_license = {
+    "name": "Creative Commons Zero v1.0 Universal",
+    "url": "https://creativecommons.org/publicdomain/zero/1.0/",
+}
```
DanPaseltiner, line 15: Looks like this dictionary is also defined in `base_service.py`. Could we DRY this out a bit?

DanPaseltiner (approved):
LGTM. I think it would be nice if we could address that `default_app_license ` is defined twice. This isn't a huge deal though so I'm not going to block merging.

PR #535, opened 2023-05-04
Trigger release workflow on special keyword

nickclyde:
The current release workflow can't run because it attempts to commit to `main`, which is a protected branch. Instead, we should make the required commit to `pyproject.toml` ourselves in a PR with the text "[RELEASE]" in the PR title. Then we can trigger the release workflow to only run on commits that contain this text!


DanPaseltiner (approved):
LGTM. Party time 🎊

PR #551, opened 2023-05-08
Install phdi from PyPI in all containers

nickclyde:
Now that PyPI releases are working properly, we can just install `phdi` from PyPI in all of our containers.


BradySkylight (approved):
Nice work!  Thanks for taking this even further Nick!  I like that we are dependent on the tag release as well as the release to pypi prior to building the containers.

Issue #544, opened 2023-05-08
Get the Kafka to Delta service running in Azure


emmastephenson:
@DanPaseltiner and @nickbristow to fill out details here

PR #568, opened 2023-05-11
Add blob_exists() method to the AzureCloudContainerConnection class

DanPaseltiner:
# PULL REQUEST

## Summary
This PR adds a blob_exists() method to the AzureCloudContainerConnection class. Given a container name and file/blob name this method returns `True` if it exists and `False` if it does not.

## Related Issue
Related to #544


nickbristow (approved):

phdi/cloud/azure.py
```diff
@@ -219,3 +219,20 @@ def list_objects(self, container_name: str, prefix: str = "") -> List[str]:
...
 
         return blob_name_list
+
+    def blob_exists(self, container_name: str, filename: str) -> bool:
+        """
+        Check if a blob exists within a container given its name and the name of the
+        container.
+
+        :param container_name: The name of the container to look for the blob in.
+        :param filename: The name of the blob to check to check the existence of.
```
robertmitchellv, line 229:
```suggestion
        :param filename: The name of the blob to check the existence of.
```

robertmitchellv (approved):
looks great! nice addition to the class 👏🏻

phdi/cloud/azure.py
DanPaseltiner: Nice catch. Thanks!

PR #578, opened 2023-05-16
Resolve errors related to converting/uploading eCR FHIR messages

BradySkylight:
# PULL REQUEST

## Summary
There are 3 separate errors reported here:

1. Missing MedicationStatement.status - the MS FHIR Converter templates were missing the value set codes for MedicationStatement.status. I found the list on HL7's FHIR IG website and added them to the ValueSet file for the ECR templates within the MS FHIR Converter App.
2. Missing DiagnositicReport.code - There was a bug in the default MS FHIR Converter template that was not evaluating the proper location for the codes for the DiagnosticReports. This has been updated to look in the proper location.
3. Error: "Conditional operation on Practitioner was not attempted because search criteria was not selective enough." - according to the documentation on the HL7 FHIR IG webpage it appears this error occurs when: "Multiple matches: The server returns a 412 Precondition Failed error indicating the client's criteria were not selective enough preferably with an OperationOutcome" - I'm not sure if we can resolve this without seeing some of the data, however, this may be due to the issues listed above. I am unable, at this point to recreate this issue. We can retest once the fixes listed above are included to see if the issue is resolved.

## Related Issue
Fixes [Issue 133](https://github.com/CDCgov/phdi-azure/issues/133)

## Additional Information
The 3rd error could be related to update, create etc...however, I didn't think we were adding a section called if-none-exists section into our 'Request' section for each of the resources.  Here is some more documentation about the different operations and WHY this error would result (see links below).  I still think this may be resolved once we have proper resources being created with the FHIR Converter for the message(s) in question.

https://build.fhir.org/http.html#cond-update

https://build.fhir.org/http.html#ccreate

https://build.fhir.org/http.html#cond-patch


emmastephenson (approved):
Looks great, thanks for addressing these Brady! Fingers crossed it also resolves that third issue.

PR #580, opened 2023-05-16
Handle empty lists when flattening blocks returned by the MPI during record linkage.

DanPaseltiner:
# PULL REQUEST

## Summary
This PR allows us to handle flattening records returned in blocks during the blocking phase of record linkage that contain empty lists. Recent user testing in LAC demonstrated that we had not considered this edge case.

## Related Issue
Fixes CDCgov/phdi-azure#131

## Additional Information
During testing with LAC we encountered many `500 internal server errors` from the record linkage service. After investigation it was determined that these errors were caused by empty lists returned by the MPI in blocks for records that were missing an MRN. More details in the ZenHub ticket.


bamader (approved):
Great catch, Dan and Brady!

PR #581, opened 2023-05-17
Log a warning for duplicate patients instead of failing

DanPaseltiner:
# PULL REQUEST

## Summary
This PR adds graceful handling for when record linkage encounters a patient ID that already exists in the MPI. Previously the service would have returned a 400 causing a failure of the pipeline in ADF. We previously updated the FHIR converter to ensure that every patient resource is given a unique ID, therefore in our current usage we only encounter these duplicate patient IDs when a failed execution of the pipeline is re-run. This PR changes the behavior in these situations so that we log a warning instead of failing. Ultimately more sophisticated handling of potential duplicate IDs is needed, but these changes will allow our debugging and testing in Azure to continue unencumbered for now.

*Note the unique key constraint is still maintained and we do not modify the patient table when a duplicate ID is presented.

Example of new behavior instead of simply raising an exception.
```
WARNING:root:Patient with ID f6a16ff7-4a31-11eb-be7b-8344edc8f36b already exists in the MPI. The patient table was not updated.
```


phdi/linkage/link.py
```diff
@@ -603,6 +603,10 @@ def link_record_against_mpi(
                     continue
                 if idx_to_col[j - 2] != "first_name" and idx_to_col[j - 2] != "address":
                     while type(blocked_record[j]) == list:
+                        # Handle empty list edge case.
+                        if len(blocked_record[j]) == 0:
+                            blocked_record[j] = ""
+                            break
```
bamader, line 609: I don't think it's anything to worry about, but I'm guessing this branch was developed off the branch where we had the flatten fix created so this code is duplicated. Just something to be aware of when merging.

bamader (approved):
LGTM!

phdi/linkage/link.py
DanPaseltiner: Yeah good catch. That PR already landed so this isn't a big deal.

phdi/linkage/postgres.py
```diff
@@ -160,7 +161,13 @@ def insert_match_patient(
...
                         json.dumps(patient_resource),
                     ]
-                    db_cursor.execute(insert_new_patient, data)
+                    try:
+                        db_cursor.execute(insert_new_patient, data)
+                    except psycopg2.errors.UniqueViolation:
+                        logging.warning(
+                            f"Patient with ID {patient_resource.get('id')} already "
+                            "exists in the MPI. The patient table was not updated."
+                        )
```
BradySkylight, line 170: awesome!

PR #585, opened 2023-05-17
Remove the Kafka to Delta table service

DanPaseltiner:
# PULL REQUEST

## Summary
Since we are no longer using this service we are removing it before the next release.


nickclyde (approved):
😭

PR #588, opened 2023-05-17
Point containers to pypi

DanPaseltiner:
# PULL REQUEST

## Summary
Now that the release process is in place we can point all of our containers back to PyPi. This also means that we no longer need to install git in the containers.

### DIFF
--- a/containers/tabulation/Dockerfile
+++ b/containers/tabulation/Dockerfile
@@ -2,12 +2,6 @@ FROM python:3.10-slim
 
 WORKDIR /code
 
-# Install git.
-# TODO: remove this step when a PyPi release of PHDI is available that includes Tabulation.
-RUN apt-get -y update && \
-    apt-get upgrade -y && \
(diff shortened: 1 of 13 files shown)


containers/tabulation/Dockerfile
```diff
@@ -2,12 +2,6 @@ FROM python:3.10-slim
 
 WORKDIR /code
 
-# Install git.
-# TODO: remove this step when a PyPi release of PHDI is available that includes Tabulation.
-RUN apt-get -y update && \
-    apt-get upgrade -y && \
-    apt-get -y install git gcc
```
nickclyde, line 9: Does this still need gcc? You'll also need to remove the run command that uninstalls git.

DanPaseltiner: oops yeah. I think we keep gcc and loose git.

nickclyde (approved):
:shipit:

PR #599, opened 2023-05-19
Update Record Linkage Container to utilize and make the external person id accessible

BradySkylight:
# PULL REQUEST

## Summary
I was able to get the external id added to the sdk and to the container (for the most part) and got tests updated and passing.  Once this has been merged into main then I can do another PR to get the external id in the container so that it matches the existing SDK in main and then the tests will pass.  I am working on an Azure PR now to update the MPI seeding script. 

This is the update to make the container use the functionality in the updated SDK
## Related Issue
Fixes #?

### DIFF
--- a/containers/record-linkage/Dockerfile
+++ b/containers/record-linkage/Dockerfile
@@ -2,6 +2,10 @@ FROM python:3.10-slim
 
 WORKDIR /code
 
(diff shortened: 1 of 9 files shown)


nickclyde (approved):
Looks good, thanks for hammering on this Brady!

PR #600, opened 2023-05-19
Update _DiagnosticReport.liquid

BradySkylight:
## Summary
During the demo we had a FHIR Upload failure because the DiagnosticReport.code was missing for all DiagnosticReports within the bundle.  This was caused by the FHIR Converter where in the ECR under the Results section for each of the "Observations/Results" there were more than one code.translation tags (usually the code can be found either in a single code.translation tag or within the code tag itself).  This message had multiple translations, which is why it failed (SMH).  I updated the template to get the code for DiagnosticReport from the FIRST iteration of code.translations if present or otherwise just from the code tag itself.


emmastephenson (approved):
LGTM, thanks for resolving this so quickly!

PR #613, opened 2023-05-24
Split up FHIR bundles (by FHIR Resources) that are too big for FHIR server to handle

BradySkylight:
# PULL REQUEST

## Summary
This PR contains changes to ensure that FHIR Bundles that have more than a specified maximum number, default to 500, of FHIR Resources will be split up into separate bundles prior to uploading them to the FHIR server.  The following changes can be found:

1- New parameter max_bundle_size, which defaults to 500 (per the issues we saw with the Azure FHIR Server) in both the call to upload bundles to FHIR server as well as the new function _split_bundle_resources.  Note: we can add this parameter to the container at a later time and as needed for our clients.
2- New function split_bundle_resources that returns a list of bundles to be processed.
3- Changes to the upload_bundle_to_fhir_server so that it calls the new split function and loops through the results.  Update the Response.status_code if there is a failure in any of the uploads.  Note: We may need to find a way to append the Response.json() to the final return Response in this function.  I have placed a TODO in the code for this work.  Note that if there are ever any errors they will be logged at the very least.
4- A few documentation updates including updating the PHDI init.py to contain the latest version.  We may need to have this automatically set by our processes in the future when cutting a new release.

## Related Issue
Fixes #https://app.zenhub.com/workspaces/dibbs-63f7aa3e1ecdbb0011edb299/issues/gh/cdcgov/phdi-azure/130


phdi/fhir/transport/http.py
```diff
@@ -76,46 +77,58 @@ def upload_bundle_to_fhir_server(
...
+                    )
+            if previous_response_status != response.status_code:
+                response.status_code = previous_response_status
+        else:
+            _log_fhir_server_error(response.status_code)
+            previous_response_status = response.status_code
+    #TODO: Find a way to return ALL of the Response.json()
+    #  For each of the split bundles instead of just the last 
+    #  One, however, we are able to pass the proper status_code 
+    #  witht he logic above.
```
DanPaseltiner: I think we probably should do this now, otherwise we are effectively taking away functionality from our users. If there is a failure on resource in the first partial bundle then they won't be able to see the actual response to debug. Now that I'm writing this out this might even impact our ability to debug FHIR upload issues.

DanPaseltiner (changes requested):
Thanks for putting this together! Looking good. I think we need to address the TODO as a part of this PR (see comment below).

phdi/fhir/transport/http.py
BradySkylight: Yeah, I was worried about this.  I'll continue plugging away at getting this to work.

phdi/fhir/transport/http.py
```diff
@@ -187,3 +200,35 @@ def _log_fhir_server_error(status_code: int, batch_entry_index: int = None) -> N
...
+        #  We may need to tweak this if there are additional fields 
+        #  the client wants at the Bundle level in their FHIR messages
+        new_bundle = copy.deepcopy(bundle_prefix)
+        # grab all the resources and place them in the entry list within 
+        # the new bundle dictionary up to the maximum number specified
+        new_bundle["entry"] = resources[entry_index:entry_index+max_bundle_size]
+        # add the new split bundle to the list to be returned
+        split_bundles.append(new_bundle)
+        entry_index = entry_index+max_bundle_size
+    return split_bundles
```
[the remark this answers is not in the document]
BradySkylight: Yes, if you do that then the last replace of entry in bundle_prefix will be present in both 'bundles' in the split_bundles list.  That was the exact issue I was running into.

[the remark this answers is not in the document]
DanPaseltiner: Now that we are no longer using `.append()` I don't think that will be an issue. Maybe I need to run this code myself to understand, but I really think each iteration would do a clean replace.

phdi/fhir/transport/http.py
```diff
@@ -76,47 +79,54 @@ def upload_bundle_to_fhir_server(
...
+
+            entries = response_json.get("entry", [])
+            for entry_index, entry in enumerate(entries):
+                entry_response = entry.get("response", {})
+
+                # FHIR bundle.entry.response.status is string type - integer status code
+                # plus may inlude a message
+                if entry_response and not entry_response.get("status", "").startswith(
+                    "200"
+                ):
```
DanPaseltiner, line 120: I think this will make us consistent with what we do in the container an handle 201s. We have seen differences across FHIR servers and differences when a resource already exists v. when it is created for the first time.

```suggestion
                if entry_response and entry_response.get("status", "") not in ["200 OK", "201 Created", "200", "201"]:
```

containers/ingestion/app/routers/fhir_transport_http.py
```diff
@@ -59,40 +59,48 @@ def upload_bundle_to_fhir_server_endpoint(
...
-        if failed_resources != []:
-            fhir_server_response.status_code = 400
-
-    status_code = "200"
-    if fhir_server_response.status_code != 200:
-        response.status_code = status.HTTP_400_BAD_REQUEST
-        status_code = "400"
+    fhir_server_responses = upload_bundle_to_fhir_server(**input)
+    full_fhir_server_response_body = {
+        "entry": [],
```
DanPaseltiner, line 64: Major nit: Doesn't `"entry"` usually get listed after `"resourceType"` and `"tyep"`?

DanPaseltiner (approved):
This looks awesome. One minor nit you can feel free to ignore. I also left a suggestion for changing when we log failures so that we can handle both `200` and `201` status codes.

containers/ingestion/app/routers/fhir_transport_http.py
BradySkylight: You are correct.  I'll fix that up real quick!  Nice catch.

PR #640, opened 2023-06-05
Resolved issues with ECR FHIR Conversion - ensure diagnosticReport has code and PractitionerId is not null

### LINKED ISSUE #628: Resolved FHIR Upload Error (Can't resolved practitioner)
Error during 3rd round of QA testing at LAC: 

{
	"status_code": "400",
	"message": {
		"fhir_server_response": {
			"fhir_server_status_code": 400,
			"fhir_server_response_body": {
				"entry": [
					{
						"response": {
							"status": "400",
							"outcome": {
								"resourceType": "OperationOutcome",
								"id": "5f44308033a5389fd6b97b64e98c93e2",
								"meta": {
									"lastUpdated": "2023-05-30T22:07:03.078029Z"
								},
								"issue": [
									{
										"severity": "error",
										"code": "invalid",
										"diagnostics": "Conditional operation on Practitioner was not attempted because search criteria was not selective enough.",
										"location": [
											"ConditionalParameters // OperationOutcome.Location is deprecated, please use OperationOutcome.Expression"
										],
										"expression": [
											"ConditionalParameters"
										]
									}
								]
							}
						}
					}
				],
				"resourceType": "Bundle",
				"type": "transaction-response"
			}
		}
	},
	"bundle": null,
	"ADFWebActivityResponseHeaders": {
		"strict-transport-security": "max-age=31536000; includeSubDomains",
		"Date": "Tue, 30 May 2023 22:06:49 GMT",
		"Server": "uvicorn",
		"Content-Length": "699",
		"Content-Type": "application/json"
	},
	"effectiveIntegrationRuntime": "AutoResolveIntegrationRuntime (West US)",
	"executionDuration": 14,
	"durationInQueue": {
		"integrationRuntimeQueue": 0
	},
	"billingReference": {
		"activityType": "ExternalActivity",
		"billableDuration": [
			{
				"meterType": "AzureIR",
				"duration": 0.016666666666666666,
				"unit": "Hours"
			}
		]
	}
}

BradySkylight:
Added logic to not generate a diagnosticReport resource if the code was missing from the ecr message - also updated all references to practitioner in all the other resources to be unique - updated logic when creating an ID for practitioners to handle additional use cases

## Summary
The following changes should address the two errors related to this [support issue](https://github.com/CDCgov/phdi-azure/issues/133), specifically the 3rd issue I discuss in the notes there as well as an additional issue found with Diagnostic Reports not having a code:

1. Perform a check to see if a code.code or code.translation.code is present for a diagnostic report in the ECR prior to actually 'including' it as a resource in the Bundle.
2. Updated the logic in the utils to generate ids to ensure that there isn't a loop if there aren't at least 1 id tag present.  If there isn't then just use the entire object to generate an id (this is one use case where it could be a null id being returned by this function).  While looping, if the conditions in the loop aren't met at least create an ID using the entire object passed in (another use case that could potentially return a null id).  Also updated the logic to be more comprehensive in regards to the potential conditions we were already trying to handle.
3. Updated all cases where practitionerId is used as a REFERENCE in other resources to ensure that the variable name is different.  Otherwise, due to liquids functionality, if 'practitionerId' and 'fullPractitionerId' are set upstream, then that value will be used for all other subsequent uses of the same variable, especially if the functionality to re-populate the variable fails...it doesn't NULL out the value in the variable but instead uses what it was already set to.

## Related Issue
Fixes #628 

## Additional Information
If you have questions or need a quick overview on Liquid, please let me know.


containers/fhir-converter/Templates/eCR/metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  "type": "ccda"
-}
+	"type": "ccda"
```
KennethSkylight, line 2: I think this was an accidental tab adding?

KennethSkylight (approved):
Just one spacing comment. but most of this looks like name changes so it looks good with me.

containers/fhir-converter/Templates/eCR/metadata.json
BradySkylight: This was actually what appears to be a correction to the spacing.  Doesn't impact anything.

emmastephenson (approved):
Looks good to me!

PR #648, opened 2023-06-07
Link based on column names not indices

### LINKED ISSUE #327: Make RL index by column name instead of index

bamader:
## Summary
This PR changes the linkage code base to work off of column name rather than column index. Previously a user had to specify numerically in some places which operations they wanted to perform on which columns, but in other places, that specification used the column name. Moreover, indexing on column number rather than name exposed us to brittleness in the future, because if the schema changes all the column mappings are shot. These changes push all conversion between column names and numbers into the codebase so that it can adapt to results returned from the DB, when needed, and more importantly, so that everything is now standard from the user's perspective. All function transformations and feature specifications now use the column name, and all feature evaluation functions now follow the same input signature (before, the enhanced algorithm had different inputs, which caused problems and necessitated this PR before changing the API endpoint).

## Related Issue
Fixes #327

### DIFF
--- a/phdi/linkage/algorithms.py
+++ b/phdi/linkage/algorithms.py
@@ -1,16 +1,3 @@
-# Predefined columns that map to the DIBBs MPI
-IDX_TO_COL = {
-    0: "address",
-    1: "birthdate",
-    2: "city",
-    3: "first_name",
-    4: "last_name",
-    5: "mrn",
-    6: "sex",
-    7: "state",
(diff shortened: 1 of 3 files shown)


m-goggins (approved):
Looks good to me! Thanks for fixing this thorny issue

BradySkylight (commented):
I may be late to the game, but this looks really good Brandon.  Nice work!

PR #651, opened 2023-06-08
Fix name and address helpers

bamader:
## Summary
Fixes dumb column name in helper function dictionaries.


m-goggins (approved):
Looks good to me!

PR #652, opened 2023-06-08
Clear out MPI ahead of testing

### LINKED ISSUE #547: Clear MPI at start of linkage tests

m-goggins:
# PULL REQUEST

## Summary
Clears MPI at the start of `test_linkage_success` to prevent other tests from failing in the case that the test is stopped part way through. This PR implements `DELETE` to remove all rows in the patient and person tables but not the tables themselves.

## Related Issue
Fixes #547 

## Additional Information
n/a

### DIFF
--- a/containers/record-linkage/tests/test_record_linkage.py
+++ b/containers/record-linkage/tests/test_record_linkage.py
@@ -45,6 +45,19 @@ def pop_mpi_env_vars():
     os.environ.pop("mpi_person_table", None)
 
 
+def clean_up_db():
+    dbconn = psycopg2.connect(
+        dbname="testdb", user="postgres", password="pw", host="localhost", port="5432"
+    )
+    cursor = dbconn.cursor()
+    cursor.execute("DROP TABLE IF EXISTS patient")
+    dbconn.commit()
+    cursor.execute("DROP TABLE IF EXISTS person")
+    dbconn.commit()
+    cursor.close()
+    dbconn.close()
+
+
 def test_health_check():
     set_mpi_env_vars()
     actual_response = client.get("/")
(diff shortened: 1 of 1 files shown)


containers/record-linkage/tests/test_record_linkage.py
```diff
@@ -96,6 +96,19 @@ def test_linkage_invalid_db_type():
...
+        dbname="testdb", user="postgres", password="pw", host="localhost", port="5432"
+    )
+    cursor = dbconn.cursor()
+    cursor.execute("DELETE FROM patient;")
+    dbconn.commit()
+    cursor.execute("DELETE FROM person;")
+    dbconn.commit()
+    cursor.close()
+    dbconn.close()
+
```
BradySkylight, line 111: Can this be used in another test besides this one, within the same file?  If yes, maybe we can put this in it's own internal function and then call it from either the start or end of the tests (Maybe both if we want to be VERY safe)?  Just a thought.

bamader:
LOL I literally did this in my API PR right now. And yeah Brady I did exactly what you suggested, putting it in its own function.

containers/record-linkage/tests/test_record_linkage.py
bamader: Apparently Marcelle and I did the same thing in different PRs 😝 . I have this in my API PR as its own function.

m-goggins: Good call, @BradySkylight I can pull this out as its own function.

m-goggins: @bamader it looks like you are dropping the tables, but I'm only deleting the contents of them. Do we want to drop the table and re-create every time, or just remove the content?

BradySkylight (approved):
This is awesome!  I really like it.  Also nice job on centralizing getting the test bundle as well.  I think we have some other tests out there in the wild that could utilize this same clean up.  Great work Marcelle!

PR #655, opened 2023-06-08
Last of the linkage bugs

bamader:
# Oh, list indices, how we curse thee

## Summary
This is the last set of linkage bugs I discovered on the SDK side of record linkage while exposing the enhanced algorithm. We were over-list indexing some places and under-indexing others, leading to us secretly comparing the wrong fields during some parts of the algorithm as well as improperly concatenating joint address objects. This should now be fixed, with updated information in our test assets and our test cases to reflect this.

### DIFF
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -634,6 +634,9 @@ def link_record_against_mpi(
                             blocked_record[j] = ""
                             break
                         blocked_record[j] = blocked_record[j][0]
+                # Name / address come back as lists of one more depth than they should
+                else:
+                    blocked_record[j] = blocked_record[j][0]
 
         clusters = _group_patient_block_by_person(data_block)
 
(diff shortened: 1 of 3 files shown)


bamader:
The asset folder looks like it has a bunch of changes but it's just applying flake formatting to the JSON, it's like all spacing based. There's only a couple meaningful lines of changed code here.

phdi/linkage/link.py
```diff
@@ -1221,8 +1224,8 @@ def _compare_address_elements(
     """
     feature_comp = False
     idx = col_to_idx[feature_col]
-    for r in record[2:][idx]:
-        for m in mpi_patient[2:][idx]:
+    for r in record[idx]:
+        for m in mpi_patient[idx]:
```
m-goggins, line 1228: Wahoo, glad to have resolved this! I think that if anyone had dug into this code, it would have been difficult to understand that this `[2:]` was to skip over `patient_id` and `person_id`

m-goggins (approved):
send it 🚀

PR #657, opened 2023-06-09
fix test to close invalid json file

m-goggins:
# PULL REQUEST

## Summary
I added a context manager to `read_linkage_config` so that the config file would definitely be closed. On Windows machines, without the context manager the config was not closed. This created an issue in `test_read_algo_errors` where reading the `not_valid_json_test.json` as a config file left it open and thus were unable to remove it in the final line of the function. 

Here is the error that would occur when running tests locally on Windows machines:
```
FAILED linkage\test_linkage.py::test_read_algo_errors - PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'not_valid_json_test.json'   
```
## Related Issue
Fixes #

## Additional Information

### DIFF
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -946,12 +946,15 @@ def read_linkage_config(config_file: pathlib.Path) -> List[dict]:
       various parts of linkage pass function.
     """
     try:
-        algo_config = json.load(open(config_file))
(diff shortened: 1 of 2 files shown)


BradySkylight (approved):
Thanks for fixing this!  Awesome!

bamader (approved):

PR #711, opened 2023-07-12
Point back to main

KennethSkylight:
After updating to python 3.10, we need to point back to main because python310 branch doesn't exist anymore


emmastephenson (approved):
Looks good, thanks for the quick fix!

PR #729, opened 2023-07-19
Safely process None-type values

### LINKED ISSUE #715: Handle fields that are `None` in the record linkage code of the SDK

bamader:
## Summary
This PR changes the record linkage module of the SDK to better handle None-type values. When a FHIR bundle came into the system with missing data trying to extract their values was halting the loop over feature comparisons which was breaking the whole linkage function. Now, we safely extract values from the record and handle None by replacing it with an empty string or a list containing the empty string (depending on the field the None occurs in). Further, when using a record with missing information for blocking, previously there were cases where we counted missing data as a blocking stipulation--so if a record came in with a last name of "", then if we blocked on last name, we would require records coming back from the MPI to also have a blank last name. Now, for any values that are either None-type or empty string-types (including a list of the empty string), we don't make this a requirement to be returned from the MPI.

## Related Issue
Fixes #715


phdi/linkage/link.py
```diff
@@ -396,6 +394,24 @@ def extract_blocking_values_from_record(
         except KeyError:
             raise ValueError(f"Field {block} is not a supported extraction field.")
 
+    # Account for any incoming FHIR resources that return no data
+    # for a field--don't count this against records to-block
+    keys_to_pop = []
+    for field in block_vals:
+        # Means the value extractor found no data in the FHIR resource
+        if block_vals[field] == "":
```
emmastephenson, line 402: I think this block would be a little more readable if you extracted the if/elif out to a helper function and made it a more functional block. I think you could do something like

```
def is_empty(field)
  # the if/elif logic of none checking and empty string checks

keys_to_remove = [x for field in block_vals if is_empty(field)]
for k in keys_to_remove:
  block_vals.pop(k)
```

emmastephenson (commented):
Looking good! Just a couple of thoughts/questions.

phdi/linkage/link.py
bamader: Sure, I can do that!

DanPaseltiner (approved):
This LGTM. I like @emmastephenson's suggestion.

phdi/linkage/link.py
```diff
@@ -1313,17 +1329,26 @@ def _flatten_patient_field_helper(resource: dict, field: str) -> any:
...
     elif field == "address":
-        return _condense_extract_address_from_resource(resource)
+        vals = _condense_extract_address_from_resource(resource)
+        return vals if vals is not None else [""]
     else:
-        return extract_value_with_resource_path(
+        val = extract_value_with_resource_path(
             resource, LINKING_FIELDS_TO_FHIRPATHS[field], selection_criteria="first"
         )
+        return val if val is not None else ""
```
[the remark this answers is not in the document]
emmastephenson: Ok perfect! I figured it was something related to the differences in field type, but it's such a subtle thing that I wanted to double-check. Appreciate the explanation! 😄

emmastephenson (approved):
Looks great, thanks for making those changes!

PR #732, opened 2023-07-20
Rollback "Publish Helm Chart"

emmastephenson:
Reverts CDCgov/phdi#727

After discussion, we're going to move the Helm charts to a separate repo. This workflow shouldn't live in the main `phdi` repository.


DanPaseltiner (approved):
LGTM

PR #764, opened 2023-08-09
Add stubbed FHIR validation to the validation service

DanPaseltiner:
# PULL REQUEST

## Summary
This PR introduces stubbed validation for FHIR bundles. Adding this functionality will allow us to simplify the pipeline currently deployed to LAC. Since we are converting the data from the person extract to FHIR and running it through the pipeline to seed it into the MPI we currently need to route the FHIR data around the validation step. This has complicated the pipeline for our users. Merging in these changes will allow us to send all data to the validation service even if for now validation for some message types is only stubbed.


emmastephenson (approved):
Looks good!

containers/validation/app/main.py
BradySkylight, line 1: This looks good, however, we may need to add an additional parameter here for the URL for the IG used to validate the FHIR bundle against.  If it's not specified within the FHIR Bundle in the meta.profile then they would need to supply an IG to validate against.

PR #777, opened 2023-08-17
Add conversion helper functions

m-goggins:
# Add conversion helper functions

## Summary
Update the way that we extract and pre-process given names and dates of birth for seeding data into LAC MPI.

## Related Issue
Fixes [ phdi-azure 273](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-azure/273)

## Additional Information

### DIFF
--- a/phdi/linkage/seed.py
+++ b/phdi/linkage/seed.py
@@ -1,6 +1,36 @@
 # This script converts patient data from parquet to patient FHIR resources.
 from typing import Dict, Tuple
 import uuid
+import datetime
+
+
+def extract_given_name(data: Dict):
+    first_name = data.get("first_name", None)
+    middle_name = data.get("middle_name", None)
+
+    given_names = []
+
+    for name in [first_name, middle_name]:
+        if name is not None:
+            given_names.append(" ".join(n for n in name.split()))
+
+    given_name = " ".join(name for name in given_names)
+
+    if given_name != "":
+        return given_name
+    else:
+        return None
+
+
+def adjust_birthdate(data: Dict):
+    # TODO: remove this function and pass in the `format` parameter to dob
+    # standardization in ReadSourceData for LAC
+    format = "%d%b%Y:00:00:00.000"
+    dob = data.get("birthdate", None)
+    if dob is not None and ":" in dob:
+        datetime_str = datetime.datetime.strptime(dob, format)
+        dob = datetime_str.strftime("%Y-%m-%d")
+    return dob
 
 
 def convert_to_patient_fhir_resources(data: Dict) -> Tuple:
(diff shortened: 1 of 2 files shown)


phdi/linkage/seed.py
```diff
@@ -1,6 +1,36 @@
...
+    for name in [first_name, middle_name]:
+        if name is not None:
+            given_names.append(" ".join(n for n in name.split()))
+
+    given_name = " ".join(name for name in given_names)
+
+    if given_name != "":
+        return given_name
+    else:
+        return None
```
BradySkylight, line 22: I'm wondering if we could use this helper function within our name standardization functionality as well?  Nice!

phdi/linkage/seed.py
```diff
@@ -1,6 +1,36 @@
...
+
+def adjust_birthdate(data: Dict):
+    # TODO: remove this function and pass in the `format` parameter to dob
+    # standardization in ReadSourceData for LAC
+    format = "%d%b%Y:00:00:00.000"
+    dob = data.get("birthdate", None)
+    if dob is not None and ":" in dob:
+        datetime_str = datetime.datetime.strptime(dob, format)
+        dob = datetime_str.strftime("%Y-%m-%d")
+    return dob
```
BradySkylight, line 33: If I'm understanding this correctly, you are receiving a dob in an expected format (line 28) and basically stripping out the time element to just return an actual date, correct?  This way we have a properly formatted date in our patient json column in the MPI.  Am I correct?

phdi/linkage/seed.py
m-goggins: Possibly; it looks like our current name standardization function returns a missing/null name as an empty string rather than a None value.

BradySkylight: Yeah, just trying to re-use where possible.  We don't have to tackle this right now.

phdi/linkage/seed.py
m-goggins: Yes, precisely. Our `_standardize_date` function can accept an existing format to understand what the date's current format is to transform it into YYYY-MM-DD, but we didn't include that in readsourcedata. This was just a quick, LAC-specific fix to get the pipeline started because some of the dates had this wonky format. 

In the future, we could update the `_standardize_date` format to accept multiple existing formats for cases like this where we're ingesting historical data with different date formats.

BradySkylight: Agreed!  That would be a good change to make in the future.  Nice!

BradySkylight (approved):
Nice work!

PR #790, opened 2023-08-24
Upload parsing schemas endpoint

DanPaseltiner:
# PULL REQUEST

## Summary
This PR expands the `/schemas` endpoint on the message parser to allow PUT operations so that users can upload new parsing schemas and update existing ones. Currently the service has the following behavior:

1. Return a `201` when a users uploads a new schema successfully 
2. Return a `200` when a user updates an existing schema. Not that updating existing schemas is only allowed when the `overwrite` parameter is set to `true` as a guardrail to prevent folks from accidentally overwriting data.
3. Return a `400` if a user requests to update an existing schema without setting `overwrite` to `true`.

We have request validation on schemas that get uploaded/updated via this endpoint to ensure they conform to the standard we expect.

## Related Issue
Fixes #https://github.com/CDCgov/phdi-azure/issues/262
Fixes #https://github.com/CDCgov/phdi-azure/issues/261

## Additional Information


emmastephenson (approved):
LGTM! The examples look great too!

containers/message-parser/app/main.py
```diff
@@ -266,3 +266,97 @@ async def get_schema(parsing_schema_name: str, response: Response) -> GetSchemaR
         response.status_code = status.HTTP_400_BAD_REQUEST
         return {"message": error.__str__(), "parsing_schema": {}}
     return {"message": "Schema found!", "parsing_schema": parsing_schema}
+
+
+PARSING_SCHEMA_DATA_TYPES = Literal[
+    "string", "integer", "float", "boolean", "date", "timestamp"
```
BradySkylight, line 272: Do we need to change the ecr.json schema that we have as a default.  I think some of the data types were listed as 'number' instead of 'integer'?

BradySkylight (approved):
This is so nice to have!  Nice work Dan.  I just had one question/concern, that is more related to our default schema.

containers/message-parser/app/main.py
DanPaseltiner: Nice catch. What are your thoughts on `float` v. `number`?

DanPaseltiner: @BradySkylight just push a change to replace `number` with `float` in `ecr.json` based on our conversation. Can you confirm that this has been resolved from your point of view?

containers/message-parser/app/default_schemas/ecr.json
```diff
@@ -203,7 +203,7 @@
          },
          "test_result_code_system": {
             "fhir_path": "Observation.valueCodeableConcept.coding.system",
-            "data_type": "number",
+            "data_type": "float",
```
BradySkylight, line 206: If this a coding system, this should be 'string' instead of float.  Weird I didn't catch this before!

containers/message-parser/app/default_schemas/ecr.json
```diff
@@ -223,22 +223,22 @@
...
          "test_result_ref_range_low": {
             "fhir_path": "Observation.refereneceRange.low.value",
-            "data_type": "number",
+            "data_type": "float",
             "nullable": true
          },
          "test_result_ref_range_low_units": {
             "fhir_path": "Observation.refereneceRange.low.unit",
-            "data_type": "number",
+            "data_type": "float",
```
BradySkylight, line 231: IF this is a unit this should be a 'string'. Sorry I didn't catch this before.

containers/message-parser/app/default_schemas/ecr.json
```diff
@@ -223,22 +223,22 @@
...
          "test_result_ref_range_high": {
             "fhir_path": "Observation.refereneceRange.high.value",
-            "data_type": "number",
+            "data_type": "float",
             "nullable": true
          },
          "test_result_ref_range_high_units": {
             "fhir_path": "Observation.refereneceRange.high.unit",
-            "data_type": "number",
+            "data_type": "float",
```
BradySkylight, line 241: This should be 'string' since it's a unit.  My bad!

BradySkylight (changes requested):
A few bugs that I introduced are still there.  Sorry about that.  I called them out.

BradySkylight (approved):
LGTM

PR #795, opened 2023-08-29
Fix eicr id in FHIR Converter and in ECR Datastore - add additional identifier fields

BradySkylight:
# PULL REQUEST

## Summary
LAC Has requested that we include eICR set id to the ECR datastore and in so doing, we realized we may not have been pulling the correct fields for eICR id.  We have made the following changes to fulfill LAC request and to include more identifiers:

### FHIR Converter Changes

1. Composition.id (FHIR) = ClincalDocument.id.root - This is the unique identifier for each eICR that may be related via an eICR Set Id - this is what should be used in the FHIR server to query for this specific resource.
2. Composition.identifier.value = ClinicalDocument.setId.extension - this is the setId of a specific eICR that may be updated over time with additional/correcting information.
3. Included the relatesTo slice for 'replaced' which includes the code 'replaces', reference to the composition resource being 'updated'/'replaced' (composition.id)
4. Updated the assets files to reflect these changes for the ecr section for the FHIR Converter.

### ECR Datastore Changes

1. eicr_id now pulls the correct position of composition.id
2. Added eicr_set_id which pulls from composition.idenfitifer
3. Added replaced_eicr_id which is the full 'Reference' to the replaced eicr composition resource.  The ID can be extracted from this field as needed.
4. Added an replaced_eicr_version - which shows which version of the eicr is being replaced.

## Related Issue
Fixes #https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-google-cloud/113


m-goggins (approved):
Looks good to me! liquid is 😵

BradySkylight:
I have tested the resulting FHIR Bundle in our AZure FHIR server and it uploaded without any errors.

m-goggins (approved):
🚀

Issue #737, opened 2023-07-21
Move RR to EICR extraction process into FHIR Converter Container


BradySkylight:

emmastephenson:
Pulling this data in from the Azure function application to the FHIR Converter, so that it lives fully within the pipeline.

PR #799, opened 2023-08-31
Refactor fhir converter container to use multiple files

emmastephenson:
# PULL REQUEST

## Summary
Refactors the FHIR converter to run out of three files, instead of all the functionality being kept in `main`.

**No new code here**! Just moving pieces around.

## Related Issue
No related ticket - I started working on #737 and got distracted :) 

## Additional Information
I tried to add the sample eICR request to the api docs, but ran into issues trying to use `Annotated` - guessing it's a versioning problem of some sort. I tried updating the Dockerfile, but it didn't work and I didn't want to spend too much more time going down this rabbit hole. I added the sample request anyways for devs to use as needed, and we can always go back later to read it in the API docs appropriately.


containers/fhir-converter/app/api_contract.py
emmastephenson, line 1: Would welcome other naming suggestions here - `api_contract` was slightly smoother than `request_response_definitions`, but still feels clunky.

nickclyde: Hmm, maybe this could just be `constants`?

nickclyde (approved):
Looks good! Nice refactor

containers/fhir-converter/app/api_contract.py
emmastephenson: I like that! Renamed

BradySkylight (approved):
LGTM!  And the addition of the FHIR Conversion input example/request is great.

PR #800, opened 2023-08-31
Add RR extraction to PHDI SDK

emmastephenson:
# PULL REQUEST

## Summary
Moves the RR extraction logic from the `ReadSourceData` function app in `phdi-azure` to the SDK.

## Related Issue
Part 1 of #737 

## Additional Information
Order of PRs here will be roughly:
1. This function is moved to SDK
2. FHIR Converter and Validation containers are updated to accept both eICR and RR data, then call this SDK function
3. Function app in `phdi-azure` calls this SDK function
4. (Eventually) LAC pipeline is updated to send both eICR and RR data to validation BB/FHIR Converter


pyproject.toml
```diff
@@ -13,7 +13,7 @@ python = "^3.10"
 smartystreets-python-sdk = "^4.10.6"
 pydantic = "^1.10.9"
 fastapi = "^0.96.0"
-httpx = "^0.23.3"
```
emmastephenson, line 16: Had to update HTTPX because otherwise `phdi` couldn't be added as a dependency of the FHIR converter container.

BradySkylight:
Your point 4. (Eventually) LAC pipeline is updated to send both eICR and RR data to validation BB/FHIR Converter may be incorrect.  If the read source data function in the LAC pipeline already grabs both eICR and RR then they can just pass those into the validator, as they do now, but then after that step only a single eICR message with the RR should be passed back, as we don't want to necessarily repeat the extraction process in both BBs, if that makes sense.  Then the FHIR Converter would still just receive a single eICR message for LAC pipeline.  However, we want this extraction process in both BBs to keep them completely modular.  Let me know if that makes sense.

pyproject.toml
DanPaseltiner: We have been trying not to add `phdi` as a dependency in the FHIR converter to avoid making a larger image even larger. If we need to do this for now I think that is fine, but I'd like to have a conversation about breaking the SDK into smaller packages, each published to pypi separately so that each BB can be more selective in terms of what parts of the SDK to include as dependencies.

BradySkylight: This is a good catch actually!  I think breaking up the containers/sdk into smaller packages is a good way to go.

emmastephenson: Happy to chat more about this offline! But for now I'm in favor of importing `phdi` in the FHIR converter as-is. If we decide to break up the SDK, I think that should be a separate effort.

DanPaseltiner: Yeah, my thought is that as the project continues to evolve the single SDK model doesn't really fit our current use case.

DanPaseltiner (approved):
🚀

PR #814, opened 2023-09-06
Add RR extraction logic, phdi dependency to FHIR converter

### LINKED ISSUE #737: Move RR to EICR extraction process into FHIR Converter Container

emmastephenson:
# PULL REQUEST

## Summary
Adds RR extraction logic to the FHIR converter. 

This also adds a dependency on `phdi`, which in turn required some substantial changes to the Dockerfile to ensure we're installing Python 3.10 (the old way installed 3.7, which wasn't compatible with some of the phdi dependencies.) 

## Related Issue
Fixes #737 

## Additional Information
One more planned PR in this effort - adding RR extraction to the validation service.

There's also a bunch of integration test work that I split into a separate PR - #826.


containers/fhir-converter/app/service.py
```diff
@@ -98,14 +98,16 @@ def convert_to_fhir(
     # Process the response from FHIR Converter.
     if converter_response.returncode == 0:
         result = json.load(open(output_data_file_path))
+        old_id = None
```
emmastephenson, line 101: Required for the Python 3.10 update 🤷

containers/fhir-converter/tests/integration/test_FHIR-Converter.py
```diff
@@ -0,0 +1,84 @@
...
+    ).read()
+    request = {
+        "input_data": input_data,
+        "input_type": "vxu",
+        "root_template": "VXU_V04",
+    }
+    vxu_conversion_response = httpx.post(CONVERT_TO_FHIR, json=request)
+
+    assert vxu_conversion_response.status_code == 200
+    assert (
```
emmastephenson, line 28: On all these integration tests, I'm not doing anything except verifying that the status is 200 and that there's a bundle in the body. I had originally hoped to create golden files for these tests and just assert response == golden, but each of the IDs in the bundle will be different on each run :/ Might be something to solve with the FHIR resource library.

containers/fhir-converter/app/constants.py
```diff
@@ -117,3 +117,8 @@ class FhirConverterInput(BaseModel):
         description="Name of the liquid template within to be used for conversion.",
         examples="VXU_V04",
     )
+    rr_data: str = Field(
+        description="If an eICR message, the accompanying Reportability Response data.",
+        example=["sample"],
+        default=None,
+    )
```
DanPaseltiner, line 124: I think we should let this be optional, so that the docs don't show it as `*required`?

containers/fhir-converter/app/main.py
```diff
@@ -52,7 +54,23 @@ async def convert(input: FhirConverterInput, response: Response):
...
+    fhir_converter_input.pop("rr_data")
+
+    # If RR is present, also need input data and conversion type eICR
+    if input.rr_data is not None:
+        if input.root_template != "EICR" or input.input_type != "ecr":
+            response.status_code = status.HTTP_422_UNPROCESSABLE_ENTITY
+            result = {
+                "message": "Reportability Response (RR) data is only accepted "
+                "for eCR conversion requests."
+            }
```
DanPaseltiner, line 67: Nice

containers/fhir-converter/Dockerfile
```diff
@@ -18,15 +18,23 @@ FROM mcr.microsoft.com/dotnet/runtime:3.1 as runtime
...
+# Install python via pyenv
+RUN apt-get update && apt-get install -y curl git build-essential libreadline-gplv2-dev  libncursesw5-dev  libssl-dev  libsqlite3-dev tk-dev libgdbm-dev libc6-dev libbz2-dev libffi-dev zlib1g-dev
+RUN curl https://pyenv.run | bash
+RUN export PYENV_ROOT="$HOME/.pyenv"
+RUN export PATH="$PYENV_ROOT/bin:$PATH"
+RUN eval "$(pyenv init -)"
+RUN /root/.pyenv/bin/pyenv install 3.10.13
+RUN /root/.pyenv/bin/pyenv global 3.10.13
+
+RUN /root/.pyenv/shims/pip install --no-cache --upgrade pip setuptools
```
DanPaseltiner, line 32: Why do we need to use pyenv here? It's fine, I'm just curious.

DanPaseltiner (approved):
Thanks @emmastephenson this looks good to me. Left one minor suggestion and a question below, but reason for either to be blocking.

containers/fhir-converter/app/constants.py
BradySkylight: agreed and good catch.

BradySkylight (approved):
This is great!  Thank you for doing this work.

containers/fhir-converter/Dockerfile
emmastephenson: More or less because there weren't other options that allowed us to install a specific Python version on top of the dotnet image. We originally tried building Python 3.7 from source but that didn't work, so pyenv ended up being the most straightforward way to install everything.

containers/fhir-converter/app/constants.py
emmastephenson: Good catch, thank you! Updated.

PR #823, opened 2023-09-08
Add stubbed HTML Insights Building Block

### LINKED ISSUE #796: Create stubbed endpoint for the HTML insights BB
## Action Requested
Create a new HTML insights building block, and add a stubbed `generate-html` endpoint.

## Acceptance Criteria
The new building block is created, and you can hit the `generate-html` endpoint and receive a token response.

## Background & Context
Given that this building block will be creating/manipulating HTML files directly, it should probably be written in JavaScript of some kind. Another component of this ticket will be figuring out what the best JS framework for this setup is. It needs to be able to export an HTML file, and be containerized like our other building blocks.

## Additional Information
This building block will eventually be used in the orchestration service, so the user journey would be:
1. Upload a batch of eICR/RR data to a DIBBs endpoint
2. Data is converted to FHIR, cleaned, standardized, etc
3. The FHIR bundle of eICR/RR data is sent to this HTML building block
4. The BB returns an HTML file to the orchestration service, and the file is shared back to the end user

nickclyde:
# PULL REQUEST

## Summary
Adds a new HTML Insights Building Block with a stubbed `/generate-html` endpoint that just returns an HTML file that displays whatever data you sent to it. The BB is an ExpressJS server that uses Mustache as a template engine.

## Related Issue
Fixes #796

### DIFF
--- a/containers/html-insights/.gitignore
+++ b/containers/html-insights/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/containers/html-insights/Dockerfile
+++ b/containers/html-insights/Dockerfile
@@ -1,0 +1,7 @@
+FROM node:18-alpine
+
+WORKDIR /code
+COPY . .
+RUN npm install
+EXPOSE 8080
+CMD ["npm", "start"]
--- a/containers/html-insights/description.md
+++ b/containers/html-insights/description.md
@@ -1,0 +1,43 @@
+## Getting Started with DIBBs HTML Insights
+
+### Introduction
+The PHDI HTML Insights app offers a REST API for processing FHIR messages into an HTML page with key insights.
+
+### Running HTML Insights
+
+The HTML Insights app can be run using Docker (or any other OCI container runtime e.g., Podman), or directly from the Node.js source code.
+
+#### Running with Docker (Recommended)
+
+To run the HTML Insights app with Docker, follow these steps.
+1. Confirm that you have Docker installed by running `docker -v`. If you do not see a response similar to what is shown below, follow [these instructions](https://docs.docker.com/get-docker/) to install Docker.
+```
+❯ docker -v
+Docker version 20.10.21, build baeda1f
+``` 
+2. Download a copy of the Docker image from the PHDI repository by running `docker pull ghcr.io/cdcgov/phdi/html-insights:latest`.
+3. Run the service with `docker run -p 8080:8080 html-insights:latest`.
+
+Congratulations, the HTML Insights app should now be running on `localhost:8080`!
+
+#### Running from Node.js Source Code
+
+We recommend running the HTML Insights app from a container, but if that is not feasible for a given use-case, it may also be run directly from Python using the steps below.
+
+1. Ensure that both Git and Node 18.x or higher are installed.
+2. Clone the PHDI repository with `git clone https://github.com/CDCgov/phdi`.
+3. Navigate to `/phdi/containers/html-insights/`.
+5. Install all of the Node dependencies for the HTML Insights app with `npm install`.
+6. Run the HTML Insights app on `localhost:8080` with `npm start`. 
+
+### Building the Docker Image
+
+To build the Docker image for the HTML Insights app from source instead of downloading it from the PHDI repository follow these steps.
+1. Ensure that both [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git) and [Docker](https://docs.docker.com/get-docker/) are installed.
+2. Clone the PHDI repository with `git clone https://github.com/CDCgov/phdi`.
+3. Navigate to `/phdi/containers/html-insights/`.
+4. Run `docker build -t html-insights .`.
+
+### The API 
+
+When viewing these docs from the `/redoc` endpoint on a running instance of the HTML Insights app or the PHDI website, detailed documentation on the API will be available below. 
--- a/containers/html-insights/index.js
+++ b/containers/html-insights/index.js
@@ -1,0 +1,27 @@
+const express = require('express');
+const mustacheExpress = require('mustache-express');
+
+const app = express();
+app.use(express.json())
+app.set('view engine', 'mustache');
+app.engine('mustache', mustacheExpress());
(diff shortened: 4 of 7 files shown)


emmastephenson (changes requested):
This looks great Nick, thank you! Just a couple questions/comments.

- Why Mustache? It seems similar to something Jekyll (although I admit I haven't heard much love for Jekyll). Mostly curious about using a templating engine vs something more robust like React.
- We should check with Ann & Sarah before officially naming this the "HTML Insights" BB - that's what I've been calling it, but it might not be the most user-friendly name 😆 
- Can you add tests? 

Thank you!

nickclyde:
- I chose Mustache since Express supports it and other lightweight template engines like EJS and Pug out of the box, and I think since we are doing server side rendering, it makes more sense to use than React, since React is primarily meant for client-side rendering.
- Sure thing, happy to rename to whatever once we settle on a name
- Will definitely add tests next!

emmastephenson:
Quick update from planning - the name decision for the BB was "HTML Conversion" to keep it consistent with our other BBs.

.github/workflows/test.yaml
```diff
@@ -156,10 +156,12 @@ jobs:
       - name: Install Pytest
         run: pip install pytest
       - name: Install dependencies
+        if: matrix.container-to-test != 'html-converter'
```
emmastephenson, line 159: I think you may be able to just put this if block once in the `list-containers` workflow - the `container-to-test` variable then wouldn't include the html-converter at all.

.github/workflows/test.yaml
```diff
@@ -189,11 +191,30 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Set up Node.js
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14  # Adjust the Node.js version as needed
+    - name: Install dependencies
+      working-directory: ./containers/html-converter  # Navigate to your Node.js app directory
```
emmastephenson, line 216: Nice! 

Definitely out of scope for this PR, but maybe we should consider adding a `/python` and a `/javascript` folder to the containers directory. As written if we write another JS container for whatever reason, it won't be tested by CI. But not something to worry about for this PR. (Refactoring all the "test/assets" paths alone will be kind of a nightmare....)

containers/html-converter/tests/integration/test_html-converter.js
```diff
@@ -0,0 +1,67 @@
...
+
+// Define a function to set up the containers.
+async function setup() {
+    console.log('Setting up tests...');
+  
+    // Define the Docker Compose configuration.
+    const composeFile = 'docker-compose.yaml';
+  
+    // Create a Docker Compose container.
+    const htmlInsights = await new DockerComposeEnvironment('./tests/integration',composeFile)
```
emmastephenson, line 19: Thank you for figuring out the integration tests for JS!! This looks great!

emmastephenson (changes requested):
Looks great! One minor workflow comment but overall this is really solid - thank you!

containers/html-converter/tests/integration/test_html-converter.js
KennethSkylight: Woo!

.github/workflows/test.yaml
KennethSkylight: Done

.github/workflows/test.yaml
```diff
@@ -196,4 +196,21 @@ jobs:
       - name: Run integration tests for containers
         working-directory: ./containers/${{matrix.directories-to-test}}/tests/integration
         run: |
-          python -m pytest -m "integration"
\ No newline at end of file
+          python -m pytest -m "integration"
+          python -m pytest
```
emmastephenson, line 200: I think this second `python -m pytest` should be removed - that will re-run all the tests across the directory, unit and integration.

containers/html-converter/description.md
```diff
@@ -0,0 +1,43 @@
...
+Docker version 20.10.21, build baeda1f
+``` 
+2. Download a copy of the Docker image from the PHDI repository by running `docker pull ghcr.io/cdcgov/phdi/html-converter:latest`.
+3. Run the service with `docker run -p 8080:8080 html-converter:latest`.
+
+Congratulations, the HTML Insights app should now be running on `localhost:8080`!
+
+#### Running from Node.js Source Code
+
+We recommend running the HTML Insights app from a container, but if that is not feasible for a given use-case, it may also be run directly from Python using the steps below.
```
emmastephenson, line 25: Nit - I think this should be "Node" instead of "Python"

containers/html-converter/description.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Building the Docker Image
+
+To build the Docker image for the HTML Insights app from source instead of downloading it from the PHDI repository follow these steps.
+1. Ensure that both [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git) and [Docker](https://docs.docker.com/get-docker/) are installed.
+2. Clone the PHDI repository with `git clone https://github.com/CDCgov/phdi`.
+3. Navigate to `/phdi/containers/html-converter/`.
+4. Run `docker build -t html-converter .`.
+
+### The API
```
emmastephenson, line 41: I don't think this is true anymore, unless Node also auto-generates docs. (Which is a good point - may need to think about generating docs for this BB down the line)

emmastephenson (changes requested):
Looking great, thank you for shepherding this PR across Kenneth! One blocking change on the `test` workflow but otherwise LGTM.

containers/html-converter/description.md
KennethSkylight: changed

.github/workflows/test.yaml
KennethSkylight: changed

emmastephenson (approved):
LGTM! Thanks Kenneth!

PR #825, opened 2023-09-12
Update geocoding census test

emmastephenson:
# PULL REQUEST

## Summary
Updates the expected values for a census geocoding test.

## Related Issue
Test runs for `main` are currently failing: https://github.com/CDCgov/phdi/actions/runs/6153430255

## Additional Information
Two things:
1. This test sends an actual geocoding request to the Census API - I assume that's intended, but didn't really look through the test & related code enough to determine if it is expected behavior
2. I'm guessing the failure came about because the Census API updating their geocoding logic, but don't have a way to confirm - their docs site doesn't list releases that I can find.


m-goggins (approved):
🚀

PR #828, opened 2023-09-14
Separate the DAL and MPI functionality - Add new DDL for multi-table schema for MPI

BradySkylight:
# PULL REQUEST

## Summary

- Created a new DAL (dal.py) in linkage to handle all DB connections and transactions
- Created a new MPI (mpi.py) that acts like postgres.py to handle functionality such as get block data, insert patient, insert person, etc...
- Add tests
- Update all the linkage code in link.py to use the new MPI
- Include .ddl file to be able to create the new multi-table schema for the MPI based upon this [ERD](https://app.diagrams.net/#G1C7RVB-EpUQNrLp0v4vXnGBVs5r3vG7DP)

## Related Issue
Fixes #114 #867


DanPaseltiner:
I think we should probably remove changes to the message parser from this PR.

BradySkylight:
> I think we should probably remove changes to the message parser from this PR.

Agreed!  I pulled out these changes and put them in a separate PR.  Thanks for pointing this out.

phdi/linkage/__init__.py
```diff
@@ -28,8 +28,9 @@
     add_person_resource,
 )
 
-from phdi.linkage.core import BaseMPIConnectorClient
-from phdi.linkage.postgres import DIBBsConnectorClient
+from phdi.linkage.mpi.core import BaseMPIConnectorClient
+from phdi.linkage.mpi.postgres import DIBBsConnectorClient
+from phdi.linkage.mpi.postgres_mpi import PGMPIConnectorClient
```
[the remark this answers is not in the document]
BradySkylight: I think this makes sense.  I'll apply these changes

phdi/linkage/dal.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def get_patient_table(self) -> Table:
+        """
+        Get a session object
+
+        this method returns a session object to the caller
+
+        :return: SQLAlchemy scoped session
+        """
+        return self.PATIENT_TABLE
```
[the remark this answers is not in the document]
BradySkylight: Correct.  These are just ORMs for the table specified, not the actual table and data.  When using reflection the state of the Table is left out.  It was one of the 'things' I was struggling with.

phdi/linkage/postgres.py
```diff
@@ -215,6 +215,9 @@ def _generate_block_query(self, block_vals: dict) -> Tuple[SQL, list[str]]:
             stub for stub in select_query_stubs
         )
 
+        print("HERE STUB:")
+        print(select_query)
+        print(select_query_stubs)
```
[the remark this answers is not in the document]
BradySkylight: Done

phdi/linkage/postgres_mpi.py
```diff
@@ -57,17 +61,23 @@ def block_data(self, block_vals: Dict) -> List[list]:
...
+            # query = session.query(self.dal.PATIENT_TABLE)
+            query = select(self.dal.PATIENT_TABLE)
+            # query = select(self.dal.PATIENT_TABLE)
+            # now tack on the where criteria using the block_vals
+            # while ensuring they exist in the table structure ORM
+            query = self._generate_block_query(
+                block_vals=block_vals,
+                query=query,
+                table=self.dal.PATIENT_TABLE,
+            )
```
m-goggins, line 71: Based on our conversation on Friday, I'm guessing we will need to generate multiple queries (1 per table if the blocking vals fall across multiple tables), right?

phdi/linkage/postgres_mpi.py
```diff
@@ -57,17 +61,23 @@ def block_data(self, block_vals: Dict) -> List[list]:
...
+            results = session.execute(query)
```
m-goggins, line 74: Following multiple queries, we will need a helper function to inner join the results from the multiple table queries before we return the results? Just trying to wrap my head around this 😅

phdi/linkage/postgres_mpi.py
BradySkylight: Great question!  We may be able to leverage 1 query using CTEs and some joins etc....   Also, I want to add some functionality in the MPI that, based upon the block values passed in, it can 'know', dynamically, which tables to generate queries for.  I'll be working on this in the new branch to this Feature branch.

bamader (commented):
The implemented code looks awesome. This is so cool to have a DAL that can handle transaction wrapping and formatting for us, and with the new flattened schema everything just looks like it'll be so much nicer to work with. From what I can see, most conversations/comments have been addressed, but I'm not approving yet because I see several functions in postgres_mpi.py that have comments about using the DAL to implement them (namely insert_match patient and an update function). If those aren't necessary for this PR just let me know and I'll switch to approve.

phdi/linkage/postgres_mpi.py
DanPaseltiner, line 1: How much of  any of this work is actually specific to Postgres? I would think that anything that is specific to the specific flavor of RDBMS we are using would be handled at the DAL level. Could this just become `mpi.py`?

phdi/linkage/postgres_mpi.py
```diff
@@ -0,0 +1,605 @@
...
+    def _sort_mpi_records(self, mpi_records: dict) -> dict:
+        sorted_records = {
+            "patient": mpi_records.get("patient"),
+            "name": mpi_records.get("name"),
+            "given_name": mpi_records.get("given_name"),
+            "identifier": mpi_records.get("identifier"),
+            "phone_number": mpi_records.get("phone_number"),
+            "address": mpi_records.get("address"),
+        }
+        return sorted_records
```
DanPaseltiner, line 482: Why is sorting the MPI records  necessary? Since we are using a map we should be access values by their keys, not indexes, so order shouldn't matter.

BradySkylight: Order does matter since we have FKs at the DB level.  We can't insert an address for a patient, with a FK to the patient_id, without the patient already existing in the DB.  Same goes for given name for a name.

phdi/linkage/postgres_mpi.py
BradySkylight: agreed.  If there would be an area where we would call out the specific DB flavors, that would be in the DAL.

phdi/linkage/postgres_mpi.py
DanPaseltiner: Agreed, but then can't we just access the keys in the order we need them instead of sorting the dict?

DanPaseltiner: It makes sense to me that the DAL would be responsible for "knowing" how to insert data in a way that complies with the schema. Sorting a hash map like this seems like an anti-pattern to me. If we want to be able to iterate over this data while disregarding the keys we should just use a list.

DanPaseltiner: Another way of thinking about this is - If we change the schema should that require changes to the DAL and MPI or just the DAL?

BradySkylight: I totally get it.  I put this in to get tests passing on Friday, but figured we would have a discussion about the best approach.  It makes sense to me to have it over in the DAL.  I'll have a commit coming.

m-goggins (approved):
Tested using `seed_testdb.py` on a couple of different scenarios and feel okay to deploy to Skylight

PR #830, opened 2023-09-15
Updated message parser to handle nulls and then provided sample file

BradySkylight:
# PULL REQUEST

## Summary
The message parser was returning "" instead of null as the value of an element if it wasn't in the FHIR Bundle.  This caused issues downstream, especially with numeric fields.  This PR includes the changes to send back null when necessary as well as a sample response from the message parser with these changes included.

## Related Issue
None

## Additional Information
This change was made in conjunction to @m-goggins work on the ECR Data Store.


DanPaseltiner (approved):
🚀

PR #831, opened 2023-09-15
Add RR extraction to validation service

### LINKED ISSUE #822: Update validation service to accept RR data directly
## Action Requested
Update the validation service to accept either a combined eICR/RR bundle (as it does today) or an eICR and RR separately. This should use the rr extraction method recently added to the SDK.

## Acceptance Criteria
The validation service can take either a single message of eCR data, or a combination of eICR and RR data and validate it either way.

## Background & Context
Long-term, we want to be able to send data to the orchestration service without needing to go through the ReadSourceData function app. Right now, that function app is doing the work of "zipping" the eICR and RR data together.

## Additional Information
This is a parallel of the work happening in #814 for the FHIR converter.

emmastephenson:
# PULL REQUEST

## Summary
Adds RR extraction to the validation service, such that RR data and eCR data can both be passed into the request.

## Related Issue
Fixes #822 

## Additional Information
Included some unrelated updates to the service as well, mostly breaking out a constants file for the request/response types and changing the return type to be the type specified instead of a `dict`. Also added an integration test for the service.


nickbristow (approved):

containers/fhir-converter/app/constants.py
```diff
@@ -119,6 +119,6 @@ class FhirConverterInput(BaseModel):
     )
     rr_data: Optional[str] = Field(
         description="If an eICR message, the accompanying Reportability Response data.",
-        example=["sample"],
+        example=['<ClinicalDocument xmlns="urn:hl7-org:v3"></ClinicalDocument>'],
```
BradySkylight, line 122: I think this is a required 'tag' for the RR being added into eICR, but not sure why we have it here.

containers/validation/app/constants.py
```diff
@@ -0,0 +1,40 @@
...
+            + " included in the return response."
+            + " Valid types are fatal, errors, warnings, information"
+        )
+    )
+    message: str = Field(description="The message to be validated.")
+    rr_data: Optional[str] = Field(
+        description="If validating an eICR message, the accompanying Reportability "
+        "Response data.",
+        default=None,
+    )
```
BradySkylight, line 26: This is great!!

containers/validation/app/utils.py
```diff
@@ -101,5 +103,28 @@ def validate_config(config: dict):
...
+    if input["rr_data"] is not None:
+        if input["message_type"] != "ecr":
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="Reportability Response (RR) data is only "
+                "accepted for eCR validation requests.",
+            )
+
+        try:
+            merged_ecr = add_rr_data_to_eicr(input["rr_data"], input["message"])
```
BradySkylight, line 117: I don't remember, forgive me, but are we ensuring that the RR data hasn't already been added to the EICR before merging them?  If not, we could probably add this as a check at the beginning of that function, just so that we don't duplicate that RR data in the EICR.  If you have already implemented this, awesome!

BradySkylight (approved):
Just had a couple of questions, but overall this is very straightforward and great!  Nice work and thanks for doing this Emma!

containers/fhir-converter/app/constants.py
emmastephenson: This is just for documentation purposes! Right now the docs show the string "sample" for rr_data, so I grabbed the minimum viable RR XML I could find. This snippet came from one of the rr_extraction samples if I recall correctly.

Screenshot 2023-09-29 at 10 36 41 AM

containers/validation/app/utils.py
emmastephenson: Oooh excellent catch! I just checked the function and I don't think we are catching that. Just added a ticket to address it to our backlog: #844

PR #840, opened 2023-09-27
Implement transformations and inserts in the new MPI using the new flattened schema

BradySkylight:
# PULL REQUEST

## Summary
Implement and test the new flattened schema, DAL, and MPI.

## Related Issue
Fixes #[115](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-google-cloud/115)
Fixes #[116](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-google-cloud/116) 
Fixes #[118](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-google-cloud/118)


tests/linkage/test_dal.py
```diff
@@ -150,61 +147,147 @@ def test_bulk_insert_and_transactions():
...
+    dal = _init_db()
+    assert dal.does_table_have_column(dal.GIVEN_NAME_TABLE, "patient_id") is False
+    assert dal.does_table_have_column(dal.NAME_TABLE, "patient_id") is True
+    assert dal.does_table_have_column(None, None) is False
+    assert dal.does_table_have_column(dal.PATIENT_TABLE, None) is False
+    assert dal.does_table_have_column(dal.PATIENT_TABLE, "") is False
+    _clean_up(dal)
+
+
+def test_single_insert():
```
m-goggins, line 242: I think we also want to test the case where someone has multiple entries for a single field and check that for those people, two inserts occur. For example, an incoming record  like `{"city": ["Boston", "Seattle"]}` should insert into the address table twice. 

This may be out of scope for this PR but wanted to flag.

tests/linkage/test_dal.py
```diff
@@ -217,22 +300,104 @@ def test_select_results():
...
+        "sex": "male",
+        "race": "UNK",
+        "ethnicity": "UNK",
+    }
+    pt2 = {
+        "person_id": None,
+        "dob": "1988-01-01",
+        "sex": "female",
+        "race": "UNK",
+        "ethnicity": "UNK",
```
m-goggins, line 383: Have we tried these tests where a value (other than person_id) is `None` instead of "UNK"?

phdi/linkage/dal.py
```diff
@@ -85,67 +116,150 @@ def transaction(self) -> None:
...
+        Perform a bulk insert operation on a table as defined
+        by the 'table' element in the dictionary along with the
+        record(s), a list of record(s) as dictionaries.  This
+        allows for several inserts to occur for different tables
+        along with a single or multiple records for each table.
+
+        :param records_with_table: a dictionary that defines the
+            the SQLAlchemy table object 'table' to insert into
+            along with a list of dictionaries as records to
+            insert into the specified table
```
DanPaseltiner, line 224: This is a dict of dicts rights? Could we include a simple example here of this data structure?

phdi/linkage/dal.py
```diff
@@ -85,67 +116,150 @@ def transaction(self) -> None:
...
+        :return: a dictionary that contains table names as keys
+            along with a list of the primary keys, if requested.
+        """
+        return_results = {}
+        for key, value in records_with_table.items():
+            new_pks = []
+            table = self.get_table_by_name(key)
+            records = value.get("records")
+            if table is not None and records is not None and len(records) > 0:
+                new_pks = self.bulk_insert_list(table, records, return_pks)
```
DanPaseltiner, line 236: Love the code reuse!

DanPaseltiner (commented):
I have to shift gears to prep for an interview, but wanted to get some comments out. I'll pick things back up a little  later on. Looking  really good!

phdi/linkage/core.py
```diff
@@ -11,61 +13,56 @@ class BaseMPIConnectorClient(ABC):
...
+        external person record created for this external person id.
+        If the external person record exists then verify that the person id,
+        either supplied or newly created, exists in the external person record.
+        If the person id supplied, or newly created, does not exist then create
+        a new external person record with a link to the person id.
+        If the external person record does exist and matches the person id,
+        either supplied or newly created, then just return the person id.
+        If the external person record does not exist, then create a new
+        external person record linked to the person id, either supplied or
+        newly created.  Then return the person id.
```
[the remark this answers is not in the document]
DanPaseltiner: Agreed. If there is no external_person_id provided then we don't create one.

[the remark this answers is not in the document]
BradySkylight: I think this is saying that if there isn't an external person record, but an external person id has been supplied, then we would create one.  I may need to reword this a bit, as well as ensure that what I am saying is what the code is doing.  But I'm pretty sure the code won't create an external person record unless an external person id has been supplied and there isn't a record in the DB for it.

tests/linkage/test_dal.py
BradySkylight: added some tests to verify this

phdi/linkage/core.py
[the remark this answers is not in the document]
BradySkylight: I updated the comment here and in the MPI.  The code is doing what I described above.

tests/linkage/test_dal.py
BradySkylight: I'm not really sure, what will happen in that case.  I wonder if it will just try to insert it as a list?  This is an interesting point and is something we can probably tackle in another ticket, if necessary.

phdi/linkage/dal.py
```diff
@@ -23,6 +23,14 @@ def __init__(self) -> None:
         self.Meta = MetaData()
         self.PATIENT_TABLE = None
         self.PERSON_TABLE = None
+        self.NAME_TABLE = None
+        self.GIVEN_NAME_TABLE = None
+        self.ID_TABLE = None
+        self.PHONE_TABLE = None
+        self.ADDRESS_TABLE = None
+        self.EXT_PERSON_TABLE = None
+        self.EXT_SOURCE_TABLE = None
```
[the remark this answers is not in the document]
BradySkylight: This is TOTALLY valid.  I was thinking about this as well and was lazy as I was trying to get it all working.  I will expand these for sure.

tests/linkage/test_dal.py
m-goggins: I'll add this as a separate ticket

phdi/linkage/dal.py
```diff
@@ -157,22 +271,52 @@ def get_session(self) -> scoped_session:
...
-        Get a session object
+        if len(self.TABLE_LIST) == 0:
+            self.initialize_schema()
 
-        this method returns a session object to the caller
+        if column_name is not None and column_name != "":
+            # TODO: I am sure there is an easier way to do this
+            for table in self.TABLE_LIST:
+                if column_name in table.c:
+                    return table
```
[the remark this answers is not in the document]
BradySkylight: That is correct.  This was intended for columns that are specific to the table in question.  If you look at all the fields that we currently 'block' on, this will work.  For other columns that exist in multiple tables (such as id) then it won't be as effective.  It's a good call out and we can put guard rails around it once we arrive at a point where we need to add them.

phdi/linkage/dal.py
```diff
@@ -85,67 +116,150 @@ def transaction(self) -> None:
...
+                        new_pk = session.execute(stmt)
+                        # TODO: I don't like this, but seems to
+                        # be one of the only ways to get this to work
+                        #  I have tried using the column name from the
+                        # PK defined in the table and that doesn't work
+                        new_pks.append(new_pk.first()[0])
+                    else:
+                        stmt = table.insert().values(record)
+                        session.execute(stmt)
+        return new_pks
```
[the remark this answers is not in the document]
BradySkylight: This is a good point.  I think we can just use the insert_bulk  with just a single record in the list.  I'll update.

PR #841, opened 2023-09-27
Added code to handle floats properly when making a list in the message parser

BradySkylight:
# PULL REQUEST

## Summary
Nick identified a bug where if float values are present in a bundle that is processed that they can't be converted to a comma separate list of floats.  The code now translates the floats into str to add them to a comma separated list.  I have added tests to cover this case as well as a updated the test schema to include elements that are floats, additionally I have added lat and long to a sample test bundle for this test case.

My only concern is if this will cause issues downstream where we are reading this data from the message parser into files that LAC can consume.  @m-goggins Can we confirm that this result will not break any of that functionality...if it does I can work with you on coming up with a solution.

See Error below:

image

## Related Issue


nickclyde (approved):
LGTM, thanks for fixing this so quickly Brady!

PR #842, opened 2023-09-27
Add postgres service to container doc generation job

nickclyde:
This should resolve this issue that came up on the latest release workflow:

https://github.com/CDCgov/phdi/actions/runs/6317099298/job/17154276278

Basically, now that the RL container uses Pyway to run migrations, it's attempting to connect to a db at the startup when generating container docs, but we aren't running a DB container for that job. This just adds a DB container so that it can run the migrations and then generate the container docs.


BradySkylight (approved):
This is great!  Thanks for doing this.  This should make it easier to test our RL/MPI etc...

PR #853, opened 2023-10-05
834 message parser meta data

### LINKED ISSUE #834: Spec out ideal payload from message parsing service  for HTML BB
### Overview

Since we are taking in the output of the Message Parsing service for the HTML BB, we also want the important information to be surfaced to the top. 

### Acceptance Criteria
- The message parser building block also returns custom spec filters and fields that the user wants to be surfaced in the html.
- Spec out the needs for an additional endpoint in the message parsing service to support the output format (likely JSON) and additional metadata needed for the HTML BB.

nickbristow:
# PULL REQUEST
Adds an option for metadata to `parse_message` that returns additional metadata for each field

## Summary

## Related Issue
Fixes #834 

## Additional Information


emmastephenson (commented):
From a technical perspective, LGTM. But I'm still a little confused about the underlying need for this work, and I'm not sure the metadata we need to add is the FHIR metadata for a bundle - my interpretation is that it's more like the "constant" fields between conditions? But would like to check with Sarah and Ann on how we intend to use this.

emmastephenson (approved):
LGTM! Thanks for adding the extra tests too.

PR #858, opened 2023-10-11
Add helper function in link.py to convert datetime objects to str

robertmitchellv:
# PULL REQUEST

## Summary
In `phdi/phdi/linkage/link.py` add a helper function that can be used to convert `datetime.date` and `datetime.datetime` objects to `str`. The function will also check if the `input_date` is already of type `str` or not; and if it is of type `str` will ensure it is in the correct format `YYYY-MM-DD`

## Related Issue
Fixes [`phdi-azure #293`](https://github.com/CDCgov/phdi-azure/issues/293)


robertmitchellv:
update branch

tests/linkage/test_linkage.py
```diff
@@ -132,6 +134,36 @@ def _clean_up_postgres_client(postgres_client):
     postgres_client.connection.close()
 
 
+@pytest.mark.parametrize(
```
DanPaseltiner, line 137: Super cool! I didn't know about `mark.parametrize`

DanPaseltiner (approved):
Thanks for putting this together @robertmitchellv. The code and tests looks great. I'm  curious if @m-goggins has thoughts on wether this should be an internal helper function and/or method on the MPI class.

I'll approve with the understanding that we might refactor in the up coming ticket when we begin actually using this function.

phdi/linkage/link.py
```diff
@@ -28,6 +29,34 @@
...
+    # if input is a date or datetime then convert in the expected format
+    elif isinstance(input_date, (date, datetime)):
+        return input_date.strftime("%Y-%m-%d")
+    # if input isn't any of the accepted formats, then return a type error
+    else:
+        raise TypeError(
+            f"Input date {input_date} is not of type date, datetime, or str."
+        )
+
+
```
BradySkylight, line 59: This may be more suited for mpi.py I think.

tests/linkage/test_linkage.py
robertmitchellv: i just read about this in the `pytest` book i'm reading!

phdi/linkage/link.py
```diff
@@ -28,6 +29,34 @@
...
+    check that it follows the 'YYYY-MM-DD' format.
+
+    :param input_date: The input date to convert, which can be of type
+        datetime.date, datetime.datetime, or str.
+    :return: The formatted date as a string in for format 'YYYY-MM-DD'.
+    """
+    # if input is str make sure it follows the expected format
+    if isinstance(input_date, str):
+        try:
+            datetime.strptime(input_date, "%Y-%m-%d")
```
m-goggins, line 44: I realize that we didn't include this in the ticket but we may want to make an additional parameter that allows the user to choose the format of the date that is returned. It can default to what's specified here but it will be useful for this to be flexible when we do fuzzy matching with Jaro-Winkler. 

might be out of scope for this ticket and could be a second ticket

tests/linkage/test_linkage.py
m-goggins, line 137: nice!

phdi/linkage/link.py
BradySkylight, line 44: We may want to retain times, in the future.  At this point, I think this is fine to strip time from Address.Period.start and .end and Phone.Period.start and end.

phdi/linkage/link.py
```diff
@@ -28,6 +29,34 @@
...
+            return input_date
+        except ValueError:
+            raise ValueError(
+                f"Input date {input_date} is not in the format 'YYYY-MM-DD'."
+            )
+    # if input is a date or datetime then convert in the expected format
+    elif isinstance(input_date, (date, datetime)):
+        return input_date.strftime("%Y-%m-%d")
```
BradySkylight, line 52: Nice!

tests/linkage/test_linkage.py
BradySkylight: This is very cool.  I wish I knew about this before!

BradySkylight (approved):
A few comments, but overall looks great!  I would suggest moving this function into mpi.py (the name has changed just recently).  Nice work Robert!

m-goggins (approved):

m-goggins:
@BradySkylight I think we'd want to keep this in link.py since it's only for linkage purposes and not meant to change the underlying data, but would be curious to hear more about inclusion in the MPI side of things.

BradySkylight:
> @BradySkylight I think we'd want to keep this in link.py since it's only for linkage purposes and not meant to change the underlying data, but would be curious to hear more about inclusion in the MPI side of things.

I think it may make sense to have it in the MPI as it provide functionality for the data from the MPI to be transformed into a format that is useful for linkage or other 'packages' to use.  I could go either way, but lean more towards it being in the MPI.

phdi/linkage/link.py
robertmitchellv: since there's still a bit of dicussion that needs to happen about where the function lives--in `link.py` or `mpi.py` i can work on adding this

robertmitchellv: > I realize that we didn't include this in the ticket but we may want to make an additional parameter that allows the user to choose the format of the date that is returned. It can default to what's specified here but it will be useful for this to be flexible when we do fuzzy matching with Jaro-Winkler.
> 
> might be out of scope for this ticket and could be a second ticket

ok, so i added this to the PR since we're still talking about where this function should live

robertmitchellv:
based on what @m-goggins said during stand up it sounds like she and @BradySkylight had a quick chat about where this should live and it sounds like the `containers/record-linkage/app/utils.py` is the best place for this to live so i'll move the function there and then push back shortly

BradySkylight (approved):
Nice work with this!

robertmitchellv:
this is currently blocked due to #861

PR #866, opened 2023-10-13
Check for RR field presence before merging RR to eCR

emmastephenson:
# PULL REQUEST

## Summary
Checks to see if RR data has already been extracted and inserted into the eCR before merging RR data into eCR. (Basically - have we already called this method?)

## Related Issue
Fixes #844 

## Additional Information
Debated for a bit about raising a `ValueError` instead of just returning the eCR values, but this seemed cleaner - there's no need for the caller to know they called the method with a pre-merged eICR.

[//]: # (PR title: Remember to name your PR descriptively!)

PR #868, opened 2023-10-17
Revert "Check for RR field presence before merging RR to eCR"

emmastephenson:
Reverts CDCgov/phdi#866

The integration tests for the containers are failing - can't replicate locally yet, but going to roll back this change until I figure out what's going on.


nickclyde (approved):
🚀

emmastephenson:
Integration tests are failing because they point at the `main` version of the SDK, which is broken...hence the rollback 🤦  But those aren't required, so merging regardless!

PR #873, opened 2023-10-19
Threshold log odds weights

bamader:
This PR makes a small improvement correction to the way we handle log odds computations. Previously, when we fuzzy matched, we just multiplied the string similarity by the log-odds weights for that field's score contribution. However, this allowed the edge case behavior that two records which are fairly dissimilar but have all fields present can, over the course of four or five fields, still accumulate a meaningfully positive log-odds score. This behavior isn't desirable because it makes our cutoff not something that's a true separation boundary. This PR gates the similarity score behind a fuzzy match threshold: only if the field comparison has an actually similar value does the weight get added.

### DIFF
--- a/phdi/linkage/link.py
+++ b/phdi/linkage/link.py
@@ -1212,7 +1212,7 @@ def _compare_name_elements(
     """
     idx = col_to_idx[feature_col]
     feature_comp = feature_funcs[feature_col](
-        [record[idx][0]],
+        [" ".join(record[idx])],
         [mpi_patient[idx]],
         feature_col,
         {feature_col: 0},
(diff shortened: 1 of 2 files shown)


phdi/linkage/link.py
```diff
@@ -503,6 +503,9 @@ def feature_match_log_odds_fuzzy_compare(
     """
     if "log_odds" not in kwargs:
         raise KeyError("Mapping of columns to m/u log-odds must be provided.")
+    threshold = 0.7
+    if "threshold" in kwargs:
+        threshold = kwargs["threshold"]
```
BradySkylight, line 508: If I'm reading this correct, our default for the log_odds threshold is .7 and if one is configured/passed in, that is what is used.  Correct?

m-goggins (approved):

BradySkylight (approved):
LGTM.  Just had a question for my own understanding, but shouldn't hold up the PR.

phdi/linkage/link.py
bamader: Yep that's it exactly! We decided a while ago the default for experimenting would be 0.7 but we always give the user the option to override.

BradySkylight (approved):

PR #874, opened 2023-10-19
Added an additional observation that contains quant. values for testing

BradySkylight:
# PULL REQUEST

## Summary
Added an additional observation in the EICR, Combined EICR/RR CDA documents as well as in the FHIR Reponses, for the converted EICR into FHIR using our FHIR Converter.  This way we can test the scenarios for converting from EICR (CDA) to FHIR and be able to ensure both qual. and quant. observation values are present.


emmastephenson (approved):
LGTM, thanks for adding tis!

PR #880, opened 2023-10-20
Update Message Parser to return data types other than strings

nickclyde:
# PULL REQUEST

## Summary
Updates the message parser to be able to return data types other than strings. Currently just converts all data to strings. This refactors out a helper function that returns data without transformation for values of length 1, otherwise joins them in a comma separated string.

## Related Issue
Fixes https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-azure/306

### DIFF
--- a/containers/message-parser/app/main.py
+++ b/containers/message-parser/app/main.py
@@ -128,6 +128,18 @@ class ParseMessageResponse(BaseModel):
     )
 
 
+# Helper function for coalescing values. Returns None if the list is empty, the value
+# if the list has one element, and a comma separated string of values if the list has
+# more than one element.
+def coalesce_values(value):
+    if len(value) == 0:
+        return None
+    if len(value) == 1:
+        return value[0]
+    else:
+        return ",".join(map(str, value))
+
+
 @app.post("/parse_message", status_code=200, responses=parse_message_response_examples)
 async def parse_message_endpoint(
     input: Annotated[ParseMessageInput, Body(examples=parse_message_request_examples)],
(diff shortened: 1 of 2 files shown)


containers/message-parser/app/main.py
```diff
@@ -128,6 +128,18 @@ class ParseMessageResponse(BaseModel):
...
+# Helper function for coalescing values. Returns None if the list is empty, the value
+# if the list has one element, and a comma separated string of values if the list has
+# more than one element.
+def coalesce_values(value):
+    if len(value) == 0:
+        return None
+    if len(value) == 1:
+        return value[0]
+    else:
+        return ",".join(map(str, value))
```
DanPaseltiner, line 140: Could we move these comments into a doc string within the function definition and add some type hints?

containers/message-parser/app/main.py
```diff
@@ -184,11 +196,7 @@ async def parse_message_endpoint(
     for field, parser in parsers.items():
         if "secondary_parsers" not in parser:
             value = parser["primary_parser"](input.message)
-            if len(value) == 0:
-                value = None
-            else:
-                value = ",".join(map(str, value))
-            parsed_values[field] = value
+            parsed_values[field] = coalesce_values(value)
```
DanPaseltiner, line 199: 🚀

containers/message-parser/app/main.py
```diff
@@ -197,12 +205,9 @@ async def parse_message_endpoint(
...
                 ].items():
-                    if len(secondary_parser(initial_value)) == 0:
-                        value[secondary_field] = None
-                    else:
-                        value[secondary_field] = ",".join(
-                            map(str, secondary_parser(initial_value))
-                        )
+                    value[secondary_field] = coalesce_values(
+                        secondary_parser(initial_value)
+                    )
```
DanPaseltiner, line 210: 🚀

DanPaseltiner (changes requested):
Thanks for putting this together @nickclyde! I have 3 thoughts:

1. I'd like to tidying up the documentation a bit for this new helper functions (see comments)
2. I agree a slightly more robust test would be good here. LMK if you want to pair. 
3. Based on our conversation at stand up yesterday I think we might want to hold of on merging this until post LAC to avoid needing to patch their eCR datastore. Maybe we checking in at backlog refinement and/or next sprint to determine if we have the bandwidth.

PR #895, opened 2023-10-31
Add MPI Indexes to Improve Performance

DanPaseltiner:
# PULL REQUEST

## Summary

This PR introduces indexing on the MPI in order to optimize performance of the new flattened schema. After profiling we have found that two categories of index improve performance.

1. Postgres [does not automatically create indexes on foreign keys ](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK). Therefore introducing indexes on foreign key columns significantly improves performance of JOINS between tables.
2. Indexing based on the specific filter criteria for each block improves performance of identifying the records that belong in each block. We compared simply indexing each column versus being more specific e.g. using the first or last 4 characters of a given column. The later offered the best performance.

## Related Issue
Fixes #https://github.com/CDCgov/phdi-azure/issues/296

## Checklist


m-goggins (approved):
Looks good to me!

PR #896, opened 2023-10-31
Select person id from patient table

DanPaseltiner:
# PULL REQUEST

## Summary
This PR makes a small modification to the base of the blocking query in the MPI. Currently we select `person_id` from the `person` table which requires that we do a JOIN between `patient` and `person`. However, since `person_id` is already included in `patient` as a foreign key we can simply select `person_id` from `patient`. Testing does not indicate that we will see a significant performance improvement from this change, but this is a simple change to make and eliminates a JOIN from our blocking query.

## Related Issue
Fixes https://github.com/CDCgov/phdi-azure/issues/323

## Additional Information

## Checklist


m-goggins (approved):
🚀

PR #920, opened 2023-11-09
Don't add person key to Patient resource

nickclyde:
# PULL REQUEST

## Summary
`link_record_against_mpi` adds a person key to the Patient resource, but this shouldn't actually be returned because it isn't valid FHIR. This is preventing uploads to the new FHIR server. This should resolve it by passing in a copy of the record instead.


DanPaseltiner (approved):
🚀

PR #927, opened 2023-11-14
Update download_object() method to return bytes if encoding is set to None

DanPaseltiner:
# PULL REQUEST

## Summary
This PR allows `AzureCloudContainerConnection.download_object()` to return blob contents as either a string when encoding is specified, or bytes when the encoding is not specified. Backwards compatibility is preserved by maintaining `UTF-8` as the default value for encoding. 

This change is necessary now in order for us to work with ZIP files in blob storage since these files cannot be directly decoded. 

Additionally we migrate to using the `readall()` method on the Azure blob client because `download_as_text` and `download_as_bytes` have been deprecated.   

## Checklist

### DIFF
--- a/tests/cloud/test_cloud.py
+++ b/tests/cloud/test_cloud.py
@@ -322,9 +322,8 @@ def test_azure_download_object(mock_get_client):
     object_path = "output/path/some-bundle-type/some-filename-1.fhir"
     object_content = {"hello": "world"}
 
-    mock_downloader = mock.Mock(
-        content_as_text=lambda encoding: json.dumps(object_content)
-    )
+    mock_downloader = mock.Mock()
+    mock_downloader.readall.return_value = json.dumps(object_content).encode("utf-8")
     mock_blob_client.download_blob.return_value = mock_downloader
 
     phdi_container_client = AzureCloudContainerConnection(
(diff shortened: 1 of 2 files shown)


tests/cloud/test_cloud.py
```diff
@@ -364,9 +363,8 @@ def test_azure_download_object_cp1252(mock_get_client):
     ) as cp1252file:
         object_content = cp1252file.read()
 
-    mock_downloader = mock.Mock(
-        content_as_text=lambda encoding: object_content.decode(encoding=encoding)
-    )
+    mock_downloader = mock.Mock()
+    mock_downloader.readall.return_value = object_content.decode("cp1252")
```
nickclyde, line 367: Somehow I had never heard of this encoding! [TIL](https://en.wikipedia.org/wiki/Windows-1252)

nickclyde (approved):
Looks good! Is the title of this PR incorrect though? Isn't this returning a string by default? (UTF-8)?

DanPaseltiner:
> Looks good! Is the title of this PR incorrect though? Isn't this returning a string by default? (UTF-8)?
Good point. I'll modify to "if encoding is set to None". Thanks!

tests/cloud/test_cloud.py
DanPaseltiner: Yeah, it was new to me at the time when this came up. I honestly can't remember why this was an issue.

PR #939, opened 2023-11-27
Create elr.json for use in `message-parser`

robertandremitchell:
# PULL REQUEST

## Summary
Based on work parsing sample ELR messages. A few considerations:

* There was no `Person` record, which was a required field for ECR. I did not see an equivalent person ID in these fields:

```python
"""I put the full json into python to get a count of reach resourceType; there are numerous repeating ones to try to make sense of
Counter({'Location': 68,
         'Organization': 26,
         'Practitioner': 23,
         'PractitionerRole': 14,
         'Device': 6,
         'DiagnosticReport': 3,
         'Specimen': 3,
         'ServiceRequest': 2,
         'Observation': 2,
         'MessageHeader': 1,
         'Provenance': 1,
         'Patient': 1,
         'EpisodeOfCare': 1,
         'Encounter': 1,
         'DocumentReference': 1,
         'ResearchStudy': 1,
         'ChargeItem': 1,
         'Procedure': 1,
         'MedicationDispense': 1})
 """
```

* Do we need to dedupe duplicate Location/Organizations in some way? These are the largest, broadly duplicated records. Another thought I had, based at least on the sample I was working with, was to try to link "backwards"; for example, a lot of recordtypes had a link to organization, practitioner records, etc.

I am wondering if for ELR data is makes sense to essentially work backwards: start with a specimen and then link back to the relevant observation, location recordtype with the relevant info there. I was not entirely sure the best way to do that within the json, or if that would make sense for all ELR data. I'm happy to keep experimenting with sample ELR / see the best way to do that kind of dynamic link-back.

## Related Issue
Fixes #
- Could not find an existing issue; this work is intended to augment message-parser.

## Additional Information
These linkages were based on sample ELR data; there may be more modifications necessary based on real-life data. Further, as noted above, the way the message is parsed could changed based on readability/usability by practitioners.

## Checklist


DanPaseltiner:
@robertandremitchell this is great! Thanks for putting this together. Here are some thoughts:

1. RE `person_id` - Our record linkage service inserts a `Person` resource into the bundle to reflect the results of linkage, so we would not expect a `Person` resource to be present in your  test data, however one should be present by the time we call the message parser in production. To simulate this in your dummy data you can spin up the RL service on your local and `POST` the FHIR bundle you have been working with. The service will return the FHIR bundle including a `Person` resource.
2. I like your thinking about how to model this data. Could you put together an ERD to show the tables that your are envisioning. We can then review as a team.

containers/message-parser/app/default_schemas/elr.json
```diff
@@ -0,0 +1,326 @@
...
+        "fhir_path": "response.FhirResource.entry.resource.where(resourceType = 'Patient').address.extension.where(url='http://hl7.org/fhir/StructureDefinition/geolocation').extension.where(url='latitude').valueDecimal",
+        "data_type": "float",
+        "nullable": true
+    },
+    "longitude": {
+        "fhir_path": "response.FhirResource.entry.resource.where(resourceType = 'Patient').address.extension.where(url='http://hl7.org/fhir/StructureDefinition/geolocation').extension.where(url='longitude').valueDecimal",
+        "data_type": "float",
+        "nullable": true
+    },
+    "specimen": {
```
DanPaseltiner, line 62: I think we'll probably need `specimen_id` or something similar as a primary key for this table.

containers/message-parser/app/default_schemas/elr.json
```diff
@@ -0,0 +1,326 @@
...
+                "nullable": true
+            },
+            "specimen_type_system": {
+                "fhir_path": "Specimen.type.coding.system",
+                "data_type": "string",
+                "nullable": true
+            }
+        }
+    },
+    "labs": {
```
DanPaseltiner, line 84: We'll probably need `lab_id` or `accession_id`

DanPaseltiner (commented):
Looking good. I think we need to make sure that we are parsing out fields that can be used as primary and foreign keys for each of the tables.

PR #949, opened 2023-11-28
Adding harmonization datetime function to fhir-converter container

### LINKED ISSUE #928: Add cleaning of DateTimes to HL7 batch cleaning function
After the discovery of this issue with processing ELRs from LAC:

```
The errors all seem to have to do with incorrectly formatted date strings:
The format of Date or DateTime is invalid: 20230820185501-080000000.

This has 5 extra zeroes after a valid datetime string. We inspected a failed ELR message, and it contained a number of correctly formatted datetimes alongside this incorrect one with five extra zeroes.
```

We've decided to add a simple cleaning function to fix the issue. We should match strings in the format (14digits-9digits) and strip off the last 5 digits to get a proper datetime. The place we should do this is in the [_clean_hl7_batch](https://github.com/CDCgov/phdi/blob/main/phdi/harmonization/hl7.py#L272) function, which gets called by the ReadSourceData function app on initial message ingestion.

robertmitchellv:
# PULL REQUEST

## Summary
Ensure that the harmonization datetime function is being called in the `fhir-converter` container. This is specifically designed for HL7v2 message types. Since we're setting templates to HL7v2 for all requests using `input_type` equal to either `"elr"` or `"vxu"`, we're going to also run the datetime function from harmonization to clean up dates for message segments that dates are supposed to be located. I also added logic in the Dockerfile to safely remove a file so that the container is able to build on Linux.

## Related Issue
Fixes #928

## Additional Information
Tested locally with ELRs structured similarly to those failing for LAC

## Checklist

### DIFF
--- a/containers/fhir-converter/Dockerfile
+++ b/containers/fhir-converter/Dockerfile
@@ -4,10 +4,10 @@ FROM mcr.microsoft.com/dotnet/sdk:3.1 as build
 RUN git clone https://github.com/microsoft/FHIR-Converter.git --branch v5.0.4 --single-branch /build/FHIR-Converter
 WORKDIR /build/FHIR-Converter
 RUN echo test > src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/VXU_V04.liquid
-RUN rm src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid
+RUN if [ -f src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid ]; then rm src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid; fi
 RUN mkdir bin && \
-    wget https://github.com/deislabs/oras/releases/download/v0.12.0/oras_0.12.0_windows_amd64.tar.gz -O ./bin/oras.tar.gz && \
-    tar -xvf ./bin/oras.tar.gz -C ./bin
+  wget https://github.com/deislabs/oras/releases/download/v0.12.0/oras_0.12.0_windows_amd64.tar.gz -O ./bin/oras.tar.gz && \
+  tar -xvf ./bin/oras.tar.gz -C ./bin
 
 # Build Microsoft FHIR Converter
 RUN dotnet build -c Release -o output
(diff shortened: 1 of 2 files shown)


DanPaseltiner (approved):
LGTM! Thank you @robertmitchellv and @nickclyde for digging through the codebase to figure this one out.

containers/fhir-converter/Dockerfile
```diff
@@ -4,10 +4,10 @@ FROM mcr.microsoft.com/dotnet/sdk:3.1 as build
 RUN git clone https://github.com/microsoft/FHIR-Converter.git --branch v5.0.4 --single-branch /build/FHIR-Converter
 WORKDIR /build/FHIR-Converter
 RUN echo test > src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/VXU_V04.liquid
-RUN rm src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid
+RUN if [ -f src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid ]; then rm src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/TestData/DecompressedFiles/.wh.VXU_V04.liquid; fi
 RUN mkdir bin && \
-    wget https://github.com/deislabs/oras/releases/download/v0.12.0/oras_0.12.0_windows_amd64.tar.gz -O ./bin/oras.tar.gz && \
-    tar -xvf ./bin/oras.tar.gz -C ./bin
+  wget https://github.com/deislabs/oras/releases/download/v0.12.0/oras_0.12.0_windows_amd64.tar.gz -O ./bin/oras.tar.gz && \
+  tar -xvf ./bin/oras.tar.gz -C ./bin
```
DanPaseltiner, line 10: 💯 @robertmitchellv could you confirm that these changes don't impact building on MacOS? Could we also get one of the folks on the team running windows to confirm things work for them as well? Just don't want to fix one issue here at the expense of creating another.

nickclyde: Looks like I am able to build just fine! Also tested with one of the failing messages, and succeeded conversion 👍🏻

nickclyde: Oh it looks like this is actually just a whitespace change @DanPaseltiner 😆

nickclyde (approved):

containers/fhir-converter/Dockerfile
robertmitchellv: for visibility this failed building on windows but it was due to `uvloop` being explicitly named in the `requirements.txt` for the `fhir-converter` container. i removed it and rebuilt and re-ran sample data through it and it worked and Marcelle was able to build and spin up the container so we have good coverage now. it might be good to test build the containers using runners that make sure it works on all OSs

PR #951, opened 2023-11-28
Fix FHIR converter sample requests

### LINKED ISSUE #937: Correct "root_template" in PHDI API docs for `/convert-to-fhir`
## Action Requested
Need to correct the "root_template" in the [API docs](https://cdcgov.github.io/phdi/latest/containers/fhir-converter.html#operation/convert_convert_to_fhir_post); rather than:

`"root_template": "V"`

It should say:

`"root_template": "VXU_V04"`

As a user of the API, I want to be able to run the sample message so that it succeeds

## Testing
It might be good to test the sample messages in the docs but that might be better for another ticket

## Background & Context
We don't want users to try the sample code and get an error

emmastephenson:
# PULL REQUEST

## Summary
Adds dropdown options for different sample requests to the API docs for our FHIR converter. All 3 sample messages were tested and work for the FHIR converter.

## Related Issue
Fixes #937 

## Additional Information
This method of setting multiple API docs probably will break when we upgrade to the latest version of FastAPI - some changes were introduced in 0.99 (we're on 0.96 at time of writing). See more [here](https://fastapi.tiangolo.com/tutorial/schema-extra-example/#__tabbed_1_2).

## Screenshots
Screenshot 2023-11-28 at 10 55 14 AM
Screenshot 2023-11-28 at 10 55 26 AM

## Checklist


containers/fhir-converter/app/main.py
```diff
@@ -43,7 +45,10 @@ async def health_check():
     status_code=200,
     responses=sample_response,
 )
-async def convert(input: FhirConverterInput, response: Response):
+async def convert(
+    input: Annotated[FhirConverterInput, Body(examples=sample_request)],
```
robertmitchellv, line 49: this is a neat way to do this 🚀

robertmitchellv (approved):
looks great 🏆

PR #956, opened 2023-11-30
Check to see if container is running & if image exists before building in local-dev.sh

emmastephenson:
# PULL REQUEST

## Summary
Tweaks to the local-dev version of the orchestration service - will now check to see if the container is running, then if the image is built, and either just run the service or build/run as a result.

## Related Issue
No related issue; just something I kept hitting locally (I'd have images laying around so it wouldn't run the containers under the old version)

## Additional Information
Should hopefully be a non-breaking change, but lmk if there are issues! 

### DIFF
--- a/containers/orchestration/local-dev.sh
+++ b/containers/orchestration/local-dev.sh
@@ -7,7 +7,7 @@ elif docker ps -a | grep -q "validation"; then
 else
     echo "validation Container does not exist, building and running..."
     # Build and run your Docker container here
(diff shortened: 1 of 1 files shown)


robertandremitchell:
@emmastephenson thanks for updating this. @zdeveloper and I observed this while I was getting set up yesterday. I tested and the containers started up properly with this change for me, fwiw.

nickbristow (approved):

PR #363, opened 2023-03-30
Update tutorials description

mattmgoldberg:
# PULL REQUEST

## Summary
Edits to tutorial copy

[//]: # (PR title: update-tutorials-description)

PR #972, opened 2023-12-07
Improve Base Query Performance

### LINKED ISSUE #362: Add the Record Linkage Service to the pipeline
Currently the Record Linkage service is deployed as part of the starter kit, but it is not used in the ingestion pipeline. This service should replace the `add_patient_hash` step in the pipeline.

AC:

GIVEN: The DIBBS starter kit pipeline
WHEN: the pipeline is executed
THEN: the record linkage service is called after the geocoding step and the FHIR bundle that it returns is passed on to the message parsing and FHIR bundle upload steps.

robertmitchellv:
# PULL REQUEST

## Summary
Remove the aggregate in the subquery for `_get_base_query` to improve performance; at the same time modify record linkage to accept this new data structure. Fix tests based on these updates.

## Related Issue
Fixes #362 #363

### DIFF
--- a/phdi/linkage/__init__.py
+++ b/phdi/linkage/__init__.py
@@ -24,6 +24,7 @@ from phdi.linkage.link import (
     read_linkage_config,
     link_record_against_mpi,
     add_person_resource,
+    aggregate_given_names_for_linkage,
 )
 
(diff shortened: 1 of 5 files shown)


m-goggins (approved):
😎

PR #982, opened 2023-12-11
Create stubbed standardization service

robertmitchellv:
# PULL REQUEST

## Summary
This is the beginning of an epic to create a new building block for just standardization. This service should allow users to apply all current DIBBs standardizations (DOB, name, phone number, address) with a single request

## Related Issue
Fixes https://github.com/CDCgov/phdi-azure/issues/366

## Checklist


containers/standardization/app/main.py
```diff
@@ -0,0 +1,53 @@
...
+from pydantic import BaseModel, Field
+
+from app.config import get_settings
+
+
+# read settings immediately to fail fast in case there are invalid values.
+get_settings()
+
+# instantiate FastAPI via PHDI's BaseService class
+# TODO:
```
emmastephenson, line 13: Not related to this PR, but I thought the plan was to create an entirely new `dibbs` folder that would have the shared util functions immediately? Rather than ever importing `phdi` to this new service.

CC @DanielPaseltiner

containers/standardization/app/main.py
```diff
@@ -0,0 +1,53 @@
...
+# - at a later point, we will import this from a non-SDK source, e.g., a dibbs
+#   library with only util functions that are building block agnostic and provide
+#   true utility to our services
+app = BaseService(
+    service_name="DIBBs Ingestion Service",
+    description_path=Path(__file__).parent.parent / "description.md",
+).start()
+
+
+class StandardizeInput(BaseModel):
```
emmastephenson, line 24: I don't think we've ever had a broad conversation about this, and we should...but my strong preference would be to split out the models into a separate file (either `models.py` or `constants.py` is what we've been using in other services). Then `main.py` would _just_ be the app setup/endpoint definitions for a given service.

containers/standardization/requirements.txt
```diff
@@ -0,0 +1,6 @@
+typing_extensions==4.6.3
+pydantic==1.10.9
+fastapi==0.96.0
+phdi @ git+https://github.com/CDCgov/phdi@main
```
emmastephenson, line 4: Same as the above comment - could we copy this over to a new `dibbs` library instead of importing `phdi`?

containers/standardization/tests/test_standardization.py
```diff
@@ -0,0 +1,13 @@
+from fastapi.testclient import TestClient
```
emmastephenson, line 1: This test looks good! But could you add integration tests as well? We're using `testcontainers-compose` in a few other containers (i.e. [FHIR Converter](https://github.com/CDCgov/phdi/tree/main/containers/fhir-converter/tests/integration)) and I'd highly recommend sticking with it. I expect for now the test won't do a thing beyond what this one already does (ping the health check endpoint) but it'll be really nice to have the infrastructure already set up for later. 
There's a [wiki on integration tests](https://github.com/CDCgov/phdi/wiki/Integration-Testing), but I'm also happy to pair if you have questions or issues with it!

emmastephenson (commented):
So speedy ⚡ Couple of questions, happy to talk offline if that's best!

containers/standardization/app/main.py
robertmitchellv: @DanielPaseltiner can probably speak to this more, but the thinking is that we'd use `phdi` only for the `BaseService`  class until we've solved how we'll get the `dibbs` package installable for the container since our repo is set up for `phdi` already and this is just for the stubbed version so we can work on downstream tickets

containers/standardization/app/main.py
robertmitchellv: i like this idea 💡 i think `models.py` feels like it makes sense--i can look at one of the other services as an example and if @DanielPaseltiner or @m-goggins have other thoughts open to those too

containers/standardization/tests/test_standardization.py
robertmitchellv: this is something i can work on now; the integration test will just be for the health check but it should probably be added into this; 🙏🏻 for linking the wiki

containers/standardization/app/main.py
m-goggins: I like @emmastephenson's idea as well and agree that models.py makes sense.

containers/standardization/app/main.py
robertmitchellv: as a POC i tried to build a wheel and see if i could get `pip` to install it from a branch and it just can't do it. it could do that if we had a release for `dibbs` but that would probably interfere with everything else already in place for `phdi`

DanPaseltiner: I don't this will be terribly hard, but I'd like to leave it out of scope for this ticket.

containers/standardization/requirements.txt
DanPaseltiner: My preference would be to set it up so that it can be installed with pip. This will make it easier to run this services outside of a container by running `uvicorn` directly. This can be a really valuable tool for development especially if you need to de-bug or want access to a fast hot reload feature. We can set up a lightweight "pip installable" `dibbs` package in another directory pretty easily. Actually if memory serves we did this in `phdi-gcp`.

containers/standardization/tests/test_standardization.py
DanPaseltiner: @robertmitchellv let's leave the integration tests out for now. We have this work captured in tickets 376 and 377.

containers/standardization/app/main.py
robertmitchellv: i added models and constants; constants has files loaded from assets just as an example for later. i think this one is resolved now

Issue #879, opened 2023-10-20
Show validation failures in the Demo UI


emmastephenson:
## Action Requested
Show failures from the validation service within the Demo UI.

## Acceptance Criteria
Errors returned from the validation service are formatted and displayed according to [the designs](https://www.figma.com/file/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Demo-Site?type=design&node-id=212-1997&mode=design&t=QJqOS9RF0ULUPXSa-0).

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHhjQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--70be42783a50d2bb54f9f6ab4629a78c0af42e63/Screenshot%202023-10-20%20at%2012.34.23%20PM.png)

## Testing
Unit tests are required

## Background & Context
This ticket is just for errors on the validation service, not warnings.

PR #991, opened 2023-12-14
show validation failures

zdeveloper:
# PULL REQUEST

## Summary
Adds validation response to `orchestration`
related demo-ui PR https://github.com/CDCgov/phdi-playground/pull/38

## Related Issue
#879

## Additional Information
- this will change the validation service to use the `sample_ecr_config.yaml` file instead of `ecr_config_lac.yaml` 

## Checklist


containers/validation/config/sample_ecr_config.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 fields:
 - fieldName: Status
-  cdaPath: "//hl7:ClinicalDocument/hl7:component/hl7:structuredBody/hl7:component/hl7:section/hl7:entry/hl7:act/hl7:code"
+  cdaPath: "//hl7:ClinicalDocument/hl7:component/hl7:structuredBody/hl7:component/hl7:section/hl7:entry/hl7:encounter/hl7:code"
```
zdeveloper, line 4: This was changed to better natch the test_zip.zip, let me know if that makes sense

containers/validation/app/utils.py
```diff
@@ -37,7 +37,7 @@ def load_ecr_config(file_path: pathlib.Path = None) -> dict:
     # if not, then just use the default sample_ecr_config.yaml
     for file in pathlib.Path(DEFAULT_CONFIG_PATH).glob("*.yaml"):
         file_name = pathlib.Path(file).stem
-        if file_path is None and re.search("ecr_config_[a-z]+", file_name.lower()):
+        if file_path is None and re.search("^ecr_config_[a-z]+", file_name.lower()):
```
zdeveloper, line 40: this is changing the config to use the `sample_ecr_config.yaml` file instead of the lac config

containers/validation/tests/test_utils.py
```diff
@@ -18,7 +18,7 @@ def test_load_ecr_config():
             config_fields,
         )
     )
-    assert value_for_ecr_version[0].get("errorType") == "warnings"
+    assert value_for_ecr_version[0].get("errorType") == "fatal"
```
zdeveloper, line 21: changed to match the `sample_ecr_config.yaml` file

containers/validation/tests/test_validation.py
```diff
@@ -79,32 +79,25 @@ def test_validate_ecr_invalid():
...
-                "Attribute: 'code' "
-                + "not in expected format. Field name: 'Sex' Attributes: "
-                + "attribute #1: 'code' with the required value pattern: "
-                + "'F|M|O|U' actual value: 't', attribute #2: 'codeSystem'"
-                + " actual value: '2.16.840.1.113883.5.1'",
+                "Attribute: 'code' not in expected format."
+                " Field name: 'Sex' Attributes: "
+                "attribute #1: 'code' with the required value "
+                "pattern: 'F|M|O|U' actual value: 't', "
+                "attribute #2: 'codeSystem' actual value: '2.16.840.1.113883.5.1'"
```
zdeveloper, line 100: changed to match the `sample_ecr_config.yaml` file

containers/orchestration/app/services.py
```diff
@@ -100,10 +110,14 @@ async def call_apis(
             print(f"Status Code: {response.status_code}")
 
             if websocket:
```
emmastephenson, line 112: We should probably think more broadly about how to keep the websocket/non-websocket endpoints in sync 😬 but for this particular PR, should we also update the regular (non-websocket) response to a failed validation? I.e., stop pipeline progression if the Validation service returns anything but a 200?

containers/validation/config/sample_ecr_config.yaml
emmastephenson: @gordonfarrell do you happen to know the correct path here? I'm not sure if the path was originally incorrect or if our testing file is off. Seems a bit odd to me that the `encounter` would have the status of the eCR in it

emmastephenson (commented):
Looking good! But a couple of starting questions

containers/orchestration/app/services.py
zdeveloper: good point, what are the non-websocket endpoints used for?

emmastephenson: They will eventually be used for actual pipeline processing by users (i.e., in the pilot of the eCR Viewer). The websockets power the demo UI, but all other requests would go through the non-websocket endpoints.

robertandremitchell (commented):
replicated. the part I initially struggled with was that the tests differed slightly based on the root directory level at which you ran them. I'm not sure if that's worth fixing / making agnostic, but overall makes sense.

containers/validation/config/sample_ecr_config.yaml
zdeveloper: ran this by @gordonfarrell and @nickbristow and are both ok with the change

emmastephenson (approved):
LGTM!

PR #1007, opened 2023-12-28
add preliminary db connection to orchestration

### LINKED ISSUE #977: Allow orchestration service to connect to local postgres database
## Action Requested
Set up the orchestration service to connect to a local Postgres database to write FHIR bundles. (This database must have the same format as the local database used in the eCR Viewer).

## Acceptance Criteria
* ~There's a `docker-compose` file to start up both the orchestration service and a Postgres database~
* There's also an option to connect to an already-running database (this will be needed when doing local development with the eCR viewer and orchestration service in tandem). Could be done by passing environment variables to the service.
* There's a DAL on the service that manages connections given the database connection information
* The database connection service is written such that a local Postgres DB is *one* way of saving data, but not the only way. (Pilot partners may want to use a different kind of data storage, so we should keep the overall design flexible)

## Testing
Integration and e2e testing will be required.

## Background & Context
We'll need this for local development of the orchestration service.

## Additional Information
This ticket does NOT need to manage writing FHIR data to the database - that will happen in a followup ticket.

BobanL:
# PULL REQUEST

## Summary
- Create a DB model of the data
- Create a repository for db connection

## Related Issue
Fixes #977 

## Additional Information
- This is just part one of the process to begin saving to the db

## Checklist


containers/orchestration/app/DAL/FhirDataModel.py
```diff
@@ -0,0 +1,11 @@
+from sqlalchemy import Column, VARCHAR
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import declarative_base
+
+Base = declarative_base()
+
+
+class FhirDataModel(Base):
```
emmastephenson, line 8: Could you rename this to something more specific to Postgres? It's pretty likely that during the pilot we'll use another data source (like cloud blob storage) and we'll probably need another DataModel class for that.

containers/orchestration/app/DAL/SqlFhirRepository.py
```diff
@@ -0,0 +1,13 @@
+from sqlalchemy.orm import Session
+
+from app.DAL.FhirRepository import FhirRepository
+
+
+class SqlAlchemyFhirRepository(FhirRepository):
+    def __init__(self, session: Session):
+        self.session = session
```
emmastephenson, line 8: Do we need to close the session at the end of use, or does SQL Alchemy do that automatically?

emmastephenson (changes requested):
Looks great, thanks Boban! Just some minor changes to request.

containers/orchestration/app/DAL/SqlFhirRepository.py
BobanL: So the session will usually be closed just at the end of where this is be instantiated. If you look at the [test](https://github.com/CDCgov/phdi/pull/1007/files#diff-71057b7766e450052c4b98c2bfe715a4b2a8f4963ba35a6ff737d2f3a15622d5R16-R21):
```py
    with Session(engine) as session:
        repo = SqlAlchemyFhirRepository(session)
        repo.persist(expected)
        actual = session.query(FhirDataModel).get(str(ecr_id))
        assert expected.ecr_id == actual.ecr_id
        assert expected.data == actual.data
```
The session will just [closes at the end of the `with Session(engine) as session:` block.](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#basics-of-using-a-session:~:text=Above%2C%20the%20Session%20is%20instantiated%20with%20an%20Engine%20associated%20with%20a%20particular%20database%20URL.%20It%20is%20then%20used%20in%20a%20Python%20context%20manager%20(i.e.%20with%3A%20statement)%20so%20that%20it%20is%20automatically%20closed%20at%20the%20end%20of%20the%20block%3B%20this%20is%20equivalent%20to%20calling%20the%20Session.close()%20method.)

emmastephenson: Ahh got it, thank you!

containers/orchestration/app/DAL/SqlFhirRepository.py
```diff
@@ -0,0 +1,13 @@
...
+
+    def persist(self, entity):
+        self.session.add(entity)
+        self.session.flush()
```
emmastephenson, line 12: I think we may need to use [`commit` instead of `flush`](https://stackoverflow.com/a/4202016) here. Otherwise the DB changes won't be permanently peristed. (I think the reason this is currently working in the integration test is that the assertions are all happening inside the `with Session` block)

emmastephenson (approved):
LGTM!

nickbristow (approved):

PR #1009, opened 2024-01-01
Kenneth/encounter

### LINKED ISSUE #995: Add Encounter Info to eCR Viewer
## Action Requested
Add "Encounter Info" section to the eCR viewer. 

## Acceptance Criteria
"Encounter Info" section is included in all eCRs, and has the following data:
* ecounter date/time
* Encounter Type
* Encounter Facility Name
* Facility Address
* Facility Contact
* Facility Type
* Facility ID
* Provider Name
* Provider Contact

There should be two sub-sections: "Encounter Details" and "Provider Details".

## Testing
Does this ticket have unusual testing requirements, or new tests that will need to be created?

## Background & Context
So what? Why is this functionality important to the user?

## Additional Information
What else does the person picking up this work need to know about?

KennethSkylight:
# PULL REQUEST

## Summary
Adding the encounter info. It also loops in provider info because they are on the same page. Also fixed some of the stylings. 

Please note the changes to unavailable data 

## Related Issue
Fixes #995 

With the available data:
image

With unavailable data
image


sarahtress:
@KennethSkylight thanks for sending! few questions:

- the section title should be "Encounter Info" not "Encounter Data"
- looks like the fields are out of order a bit, the order should match the design exactly. (for ex, I'm seeing Facility ID at the top, but it should be at the bottom. also looks like Encounter Type is below Facility Name and Address when it should be above)
- do you have any other example data you could run through to show more fields being available? I'm surprised facility address isn't available for the eCR for ex
- Where is "facility type" coming from? I'm just surprised to see "healthcare provider" come up as the facility type, but that just might be me being confused lol
- under "Unavailable Info" it looks like Provider Name and Provider Contact are coming up under "Encounter Details" but they should only be under "Provider Details"

I also have 2 comments that I must have missed on the design review for Patient Info, so apologies on missing this:
- any clue what's happening with the blank "preferred language" field? is it unavailable? 
- just realized the ordering is off for the Patient Info section. Patient ID should be at the very bottom of "Demographics" and Social History section is all jumbled in order

KennethSkylight:
# With some unavailable info 

image


image



# With all the info 
image

sarahtress:
@KennethSkylight sorry might have been confusing with my comments - the formatting should still match the design here: 
Screenshot 2024-01-03 at 1 08 59 PM

so the changes that still to be made would be:
- overall title should be Encounter Info, not Provider Info
- there should be a Provider Details section at the bottom with Provider Name and Provider Contact under it. Provider Name and Provider Contact should not be duplicated fields under Encounter Details (but they both live under Encounter Info)
- ordering of fields still looks off to me - Encounter Date/Time should be before Encounter Type
- also was it intentional to turn Encounter Date/Time into just Encounter Date? is the time available to display?

also noticed one formatting issue that might need to be a different ticket - the gray for the divider lines looks like it's too dark - could you either match the shade to the designs here or add it to the other ticket?

thanks! and sorry for not catching everything on the first go

PR #1010, opened 2024-01-02
Add pre-commit.ci; update linting

emmastephenson:
# PULL REQUEST

## Summary
4 major changes here:
* Run pre-commit automatically when PRs are opened (will run our `pre-commit` config against all files in the PR)
* Add prettier and eslint to pre-commit
* Add python-reorder-imports to pre-commit
* Add a new Github action, `lint`, to run all linters (instead of running it in `test`)

## Related Issue
n/a - couple of Slack conversations prompted this!

## Additional Information
There are only 2 files that really changed here, `pre-commit-config.yaml` and `lint.yaml`. All the other files are the result of running `pre-commit run --all-files` on the repo.

Once this PR is merged, I'll update the repo to require passing the `lint` check before merge. I'll also remove `code-check-python` from the `test` GH action.

## Checklist


DanPaseltiner (approved):
Looks good to me. I don't think this needs to be blocking, but should we also pull linting out of `test`? Unless I'm missing something it now looks like we are running some checks twice.

emmastephenson:
> Looks good to me. I don't think this needs to be blocking, but should we also pull linting out of `test`? Unless I'm missing something it now looks like we are running some checks twice.

@DanPaseltiner good call, just removed it from `test`. Once this is in I'll make the linters a required check.
I also just added yaml formatting in prettier, so even more files changed 🙃

BobanL (approved):
Changes look good to me! Thanks for looking into this :D

containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
```diff
@@ -1,152 +1,184 @@
...
+                    </div>
+                  </div>
                 </div>
+              </div>
+            </div>
+            <div className={"section__line"} />
+            <div className="grid-row">
+              <h4>Clinical sections relevant to reportable condition</h4>
+              <div className={"padding-top-05"}>
+                No matching clinical data found in this eCR (temp){" "}
```
BobanL, line 174: The tests are failing because of this space added. It shouldn't have originally been there anyways so it's safe to remove.

DanPaseltiner (dismissed):
Oops, meant to approve before, not just comment. Sorry if this was blocking.

emmastephenson:
> Oops, meant to approve before, not just comment. Sorry if this was blocking.

@DanPaseltiner nah don't worry, you're not blocking - eslint isn't playing nice with pre-commit locally, so I'm trying to figure that out before I merge.

containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
emmastephenson: Thanks for finding that - will remove the space!

emmastephenson:
@DanPaseltiner and @BobanL, ready for re-review! Should be all set with pre-commit now ✔️

BobanL (approved):
Changes look good to me! Thank you for adding this 😄

PR #1017, opened 2024-01-05
Add config and steps to save to DB in orchestration

### LINKED ISSUE #978: Save FHIR bundles at the end of the orchestration pipeline
## Action Requested
Save FHIR bundles and their associated eICR ID to local Postgres storage.

## Acceptance Criteria
When running orchestration with the appropriate configuration file and a local database, output from the pipeline is saved in the format `ecr_id : FHIR bundle`.

## Testing
This will need unit and integration tests.

## Background & Context
This is the expected flow for NBS users - data will be saved after pipeline output to data storage. Since we don't know the exact flavor of data storage, for now we'll work with local Postgres databases.

nickbristow:
# PULL REQUEST

## Summary
In the orchestration, save to DB as a step.
## Related Issue
Fixes #978

## Additional Information

## Checklist


BobanL (commented):
Looks good to me, but as one of my first time looking at python I don't feel comfortable approving it!

containers/orchestration/app/default_configs/sample-orchestration-config.json
```diff
@@ -48,6 +51,10 @@
     {
       "service": "message_parser",
```
emmastephenson, line 52: We can make this a follow-up PR, but since we're saving the FHIR bundles to the DB, message parsing shouldn't be a step in the default (eCR processing) config anymore.

containers/orchestration/app/services.py
```diff
@@ -81,6 +90,31 @@ def message_parser_payload(**kwargs) -> dict:
...
+        return Response(content=e, status_code=500)
+
+
+def save_to_db_payload(**kwargs) -> dict:
+    response = kwargs["response"]
+    r = response.json()
+    if r.get("parsed_values", {}).get("eicr_id"):
+        ecr_id = r["parsed_values"]["eicr_id"]
+    else:
+        ecr_id = uuid.uuid4()
```
emmastephenson, line 114: I'd actually argue there should be different behavior here - if the eicr ID can't be found, I think the service should throw an exception rather than generate a random UUID.

The product thinking here is that we'll only be able to find the eCRs users are looking at in NBS through the original eICR ID - they won't know what the generated ID is.

containers/orchestration/local-dev.sh
```diff
@@ -51,6 +51,10 @@ else
     docker run -d -p 8083:8080 ingestion
 fi
 
+cd ../ecr-viewer
```
emmastephenson, line 54: nit: Is it possible to check if a database is already running? In the same way we check for other Docker containers before trying to start them up.

emmastephenson (commented):
Mostly looking great! My feedback is more about expected behavior than implementation. Thank you both for tackling this! 🙏

emmastephenson (changes requested):
Meant to request changes, not comment 🤦

containers/orchestration/local-dev.sh
gordonfarrell: Gooood ideeaaaoooood

containers/orchestration/app/services.py
gordonfarrell: Good idea, added an exception here and a test for it

containers/orchestration/app/default_configs/sample-orchestration-config.json
gordonfarrell: Cool we'll leave it out of this one and then take a look

emmastephenson (approved):
Looks great! Thank you for making those changes!

PR #1037, opened 2024-01-09
Support Referenced Resources in Message Parser Schema

### LINKED ISSUE #1021: Update message parser to handle referenced resources in schema

bamader:
# PULL REQUEST

## Summary
This PR adds some basic support for the message parser service to accept referenced resources in its parsing schemas. These referenced resources will now use a special DIBBs placeholder token `#REF#` to denote when a resource must be located elsewhere in the bundle. Validation, unit testing, a new schema, and quality test data has also been developed according to the FHIR spec.

## Related Issue
Fixes #1021 

## Additional Information
Some of the code here is in a temporary location due to us wanting to add a separate `extract-phdc-from-fhir` endpoint, so don't worry that it and the validation code currently lives within the message parser. This change set is meant to cover the logic rather than the location, so that code will get moved once this lands and we begin work on the final endpoints.


containers/message-parser/app/default_schemas/test_reference_schema.json
```diff
@@ -0,0 +1,44 @@
...
+        "nullable": true
+      },
+      "test_result_code_display": {
+        "fhir_path": "Observation.valueCodeableConcept.coding.display",
+        "data_type": "string",
+        "nullable": true
+      },
+      "ordering_provider": {
+        "fhir_path": "Bundle.entry.resource.where(resourceType='Organization').where(id='#REF#').name",
+        "reference_lookup": "Observation.performer.first().reference",
```
DanPaseltiner, line 32: This is fine for now, but we should keep in mind that at some point we may need to handle multiple references here.

containers/message-parser/app/main.py
```diff
@@ -120,6 +120,36 @@ def require_schema_or_schema_name(cls, values):
             )
         return values
 
+    @root_validator
```
DanPaseltiner, line 123: I added #1044 for adding doc strings to all of these validator functions. @bamader you maintained consistency with the existing code here, so I think adding this documentation can be out of scope, but it is something we should address.

containers/message-parser/app/main.py
```diff
@@ -136,6 +166,12 @@ class ParseMessageResponse(BaseModel):
     )
 
 
+# TODO
+@app.post("/fhir_to_phdc", status_code=200, responses=parse_message_response_examples)
```
DanPaseltiner, line 170: These response examples will be wrong for this endpoint, could we just remove `response=parse_message_response_examples`? We'll need a ticket down the road to add examples of both requests and responses to this endpoint.

DanPaseltiner: Added #1045 and #1046 for eventually adding request and response examples for this endpoint.

containers/message-parser/app/utils.py
```diff
@@ -90,9 +92,14 @@ def get_parsers(extraction_schema: frozendict) -> frozendict:
...
-                    secondary_field_definition["fhir_path"]
-                )
+                if not secondary_field_definition["fhir_path"].startswith("Bundle"):
+                    secondary_parsers[secondary_field] = fhirpathpy.compile(
+                        secondary_field_definition["fhir_path"]
+                    )
+                else:
+                    secondary_parsers[secondary_field] = secondary_field_definition[
+                        "fhir_path"
+                    ]  # noqa
```
DanPaseltiner, line 102: This change makes sense to me, but in the spirit of not repeatedly computing the same parser could we compile a parse for the value of `reference_lookup` and store that in conjunction with fhirpath that will need the value of the reference inserted?

DanPaseltiner (changes requested):
@bamader thanks for getting started on this work so quickly! I like the changes you have introduced to the parsing schema config. I have also identified a few other tickets we'll need eventually, these have been added to our backlog and are not blocking for this PR. There are 2 areas that I still think we need a little work:

1. See my comment about computing parsers for the reference lookup fhirpaths as part of `get_parsers`.
2. The additional unit tests are great, but I don't see any tests or modifications to existing tests that prove we are ensuring backwards compatibility. What happens in one of these new schemas is passed to `/parse_message` do things work correctly?

containers/message-parser/app/main.py
bamader: I agree that it's out of scope but I think it'll be a great way for someone else to walk through this code and add some documentation around how the logic works. That way we can help knowledge share and it's not just me knowing our reference logic.

containers/message-parser/app/main.py
bamader: The example requests and responses should now be correct for reference objects and schemas. I basically repurposed some of the stuff I use for our testing so it should be good to go now.

containers/message-parser/app/utils.py
bamader: Yep you're totally right here. I've moved the compilation logic back to `get_parsers` where we can now compile everything that isn't the insertion of the referenced ID into the external resource's reference path.

bamader:
@DanPaseltiner Ultimately it seemed like the best way to address your comments was to directly point to the places in my other PR where I'd implemented the logic, but that seemed like it'd be really clunky and ineffective. I ended up merging my other feature branch directly into this one because it was a direct work-off anyway, and I think this now clears things up a bit more effectively. Logic is correctly located, references are all compiled in `get_parsers`, and we have appropriate sample data, examples, and unit tests to show validation and successful reference schema parsing.

The one big thing to note is that with the fully implementation of the separate `fhir-to-phdc` endpoint, the `parse-message` endpoint is now left completely unaltered. I know we talked this morning about how that endpoint might also make use of some of the reference extraction stuff, but for the time being that's fully contained in the fhir to phdc endpoint. The parsing compilation is still present in `get_parsers` so theoretically the message parser _could_ do the same thing (if we moved the logic in the endpoint up into `parse-message`), but we had said we wanted to do the development insulated from parse-message in the beginning. Let me know how much, if any, logic you think should be ported backward into `parse-message`, or we can make that a separate ticket to expand the functionality.

containers/message-parser/app/models.py
DanPaseltiner, line 1: 🚀

containers/message-parser/app/utils.py
```diff
@@ -90,9 +92,19 @@ def get_parsers(extraction_schema: frozendict) -> frozendict:
...
+                        secondary_field_definition["fhir_path"]
+                    )
+
+                # Reference case: secondary field is located on a different resource,
+                # so we can't compile the fhir_path proper; instead, compile the
+                # reference for quick access later
+                else:
+                    secondary_parsers[secondary_field] = fhirpathpy.compile(
+                        secondary_field_definition["reference_lookup"]
+                    )
```
DanPaseltiner, line 107: I'm thinking about how this would work in practice and not quite sure I follow. We'll need the parser and the fhir path string to insert the reference into. What if we make the value of `secondary_parsers[secondary_field]` a dict and set it equal to something like

```
{
"reference_parser": fhirpathpy.compile(
                        secondary_field_definition["reference_lookup"],
"secondary_fhir_path": FHIRPATH-TO-INSERT-REFERENCE-INTO  
}
```

This way when we run through our parsers everything we need is available in one data structure.

DanPaseltiner (commented):
I'll finish reviewing later on, but here is one initial comment. Thanks for your continued work on this!

containers/message-parser/app/utils.py
DanPaseltiner: I realize this will break `/parse-message` in its current form if it receives a schema containing references. We should be able to address this though, potentially in another ticket.

containers/message-parser/tests/test_fhir_to_phdc_endpoint.py
```diff
@@ -0,0 +1,139 @@
...
+
+def test_parse_message_success_external_schema():
+    request = {
+        "parsing_schema": test_schema,
+        "message": fhir_bundle,
+    }
+
+    actual_response = client.post("/fhir_to_phdc", json=request)
+    assert actual_response.status_code == 200
+    assert actual_response.json() == expected_successful_response
```
DanPaseltiner, line 68: Nice!

containers/message-parser/app/main.py
```diff
@@ -218,23 +130,120 @@ async def parse_message_endpoint(
...
+                                return {
+                                    "message": "Provided bundle does not contain a "
+                                    "resource matching reference criteria defined "
+                                    "in schema",
+                                    "parsed_values": {},
+                                }
+                            else:
+                                value[secondary_field] = ",".join(map(str, ref_res))
+                values.append(value)
+            parsed_values[field] = values
```
DanPaseltiner, line 240: This is great. I think we should refactor this our into it's own function and then call it here and in `/parse-message`. For now both endpoints can simply return the parsed values. Eventually, both endpoints will start by running this code, `/parse-message` will continue to just return the parsed values while `fhir-to-phdc` will pass the parsed values to the PHDC builder and then return PHDC to the client.

bamader:
@DanPaseltiner Everything we talked about yesterday during the review should now be in place! Major changes include:

* Modified `get_parsers` to create a data structure within the `secondary_fields` dictionary, where we can hold the pre-compiled reference lookups as well as the FHIR paths we need to insert into
* Refactoring the repeated code of both endpoints that extracted and applied the `get_parsers` into its own helper function that both endpoints now call--both are therefore set up to handle reference resources
* Refactoring the validation that enforces a parsing schema that uses reference fields having all required attributes so that it can be called by both endpoints without duplication
* Added unit tests to both endpoints to confirm that they can handle reference-based schemas while still maintaining backwards compatibility with `parse-message`
* Documentation and comments throughout that indicate where changes might need to be made in the future based on other tickets we've created and on work we intend to do (e.g. make the phdc endpoint return actual phdc)

I think this addresses everything that was outstanding, so feel free to review again when you have the chance!

containers/message-parser/tests/test_fhir_to_phdc_endpoint.py
```diff
@@ -0,0 +1,141 @@
...
+    / "app"
+    / "default_schemas"
+    / "test_reference_schema.json"
+)
+
+with open(test_schema_path, "r") as file:
+    test_schema = json.load(file)
+
+# TODO: Once we complete M2, the response type of the phdc endpoint will
+# become phdc rather than justs JSON, so update the expected response
```
DanPaseltiner, line 33:
```suggestion
# become PHDC rather than just JSON, so update the expected response
```

DanPaseltiner (dismissed):
LGTM! Thanks for bearing with me on this as we figured out the right path forward here. 🚀

DanPaseltiner (approved):

PR #1043, opened 2024-01-10
Side Nav

### LINKED ISSUE #994: Add side navigation bar to eCR Viewer
## Action Requested
Add the side nav bar to the eCR Viewer.

## Acceptance Criteria
Side navigation populates on every eCR Viewer render. 

## Testing
Usual unit testing requirements. 

## Background & Context
Navigating through eCR data is a lengthy process - we want users to be able to jump between sections easily.

## Additional Information
This navigation bar should be sticky/always visible on the page no matter where the user is in the document.

JNygaard-Skylight:
# PULL REQUEST

## Summary

## Related Issue
Fixes #994 

## Additional Information

## Checklist


JNygaard-Skylight:
@sarahtress, I still have some behind-the-scenes work to do for this ticket, but could you take a look at what I currently have?
image

sarahtress:
> @sarahtress, I still have some behind-the-scenes work to do for this ticket, but could you take a look at what I currently have? image

thanks for sending @JNygaard-Skylight few things to match the designs:
- the width of the navigation menu should be 170 px, looks wider than that here
- I just realized that the styling of my navigation in the designs doesn't match the USWDS default, so I'll work on changing that in the designs!
- is it possible to make the main section headings (Patient Info, Encounter Info, Clinical Info, Lab Info, eCR Metadata, Unavailable Info) bolded in the side navigation?
- the design also has eCR Summary and eCR Document as items in the side navigation, could those be added? they should be in black text, rather than gray
- right now everything is left justified, but the side navigation and the eCR data should be centered on the page
- also, does the side navigation stay in place when you scroll? (as in, you should always be able to see the side navigation, it shouldn't disappear as you scroll)

let me know if you have any questions!

JNygaard-Skylight:
image
image
@sarahtress, some new screenshots, how do things look now?

nickbristow (approved):

sarahtress:
> image image @sarahtress, some new screenshots, how do things look now?

thanks for the new screenshots @JNygaard-Skylight. it's looking closer, but still some things to be adjusted. you may also need to send a video for this one bc it's interaction heavy and I can't totally tell what's going on as you scroll

- there should be three levels of hierarchy in the navigation, so only eCR Summary and eCR Document should be fully left justified. I attached a screenshot which labels the 3 levels within the navigation in the design, but let me know if you have more questions
- the side navigation shouldn't go above the content on the right. when at the top of the page, the top of the side navigation and the top of the "eCR Summary" text should be at the same height. I also attached a screenshot to show this alignment
- in the screenshots you sent, I'm not seeing any active sections on the side navigation. as you scroll the page, it should be responsive to show where you are on a page with bolded blue text. good example is on this page for the component, the right side navigation is responsive as you scroll the page (... I see the styling of that component is different though and matches my original styling... perhaps there is a different component with this functionality built in??)


Screenshot 2024-01-29 at 2 52 00 PM

Screenshot 2024-01-29 at 2 59 20 PM

containers/ecr-viewer/src/app/view-data/page.tsx
```diff
@@ -16,14 +16,77 @@ import Demographics from "./components/Demographics";
...
+    {
+      content: null,
+      expanded: true,
+      id: "1",
+      headingLevel: "h2",
+      link: [
+        <a href="#patient-info">Patient Info</a>,
+        <SideNav
+          isSubnav={true}
+          items={[<a href="#demographics">Demographics</a>]}
```
emmastephenson, line 60: I think there should also be a "Social History" sub-section here

containers/ecr-viewer/src/app/view-data/page.tsx
```diff
@@ -16,14 +16,77 @@ import Demographics from "./components/Demographics";
...
 import EcrMetadata from "./components/EcrMetadata";
-import EncounterDetails from "./components/Encounter";
 
 const ECRViewerPage = () => {
   const [fhirBundle, setFhirBundle] = useState<Bundle>();
   const [mappings, setMappings] = useState<PathMappings>({});
   const [errors, setErrors] = useState<Error | unknown>(null);
   const searchParams = useSearchParams();
   const fhirId = searchParams.get("id") ?? "";
+  const accordionItems: any[] = [
```
emmastephenson, line 26: I'm wondering if there's a way to do this more dynamically...because as-is, I think we'll need to manually update the side nav with every new section that's added, and I can see that getting hard to maintain somewhat quickly. 

My idea would be to generate the side nav based on the `h2` or `h3` elements within the accordion. Anything on the page with h2 or h3 tags is auto-included in the side nav, and any heading levels below that are ignored. 

I'm not sure if this would be practical with the `Accordion` components that were recently added, though - happy to chat more offline.

containers/ecr-viewer/src/app/view-data/page.tsx
```diff
@@ -134,14 +183,32 @@ const ECRViewerPage = () => {
...
-          <h1 className={"page-title"}>EZ eCR Viewer</h1>
+          <h1 className="page-title">EZ eCR Viewer</h1>
         </header>
-        <div className={"ecr-viewer-container"}>
-          <h2 className="margin-bottom-3">eCR Summary</h2>
-          <EcrSummary fhirPathMappings={mappings} fhirBundle={fhirBundle} />
-          <div className="margin-top-6">
-            <h2 className="margin-bottom-3">eCR Document</h2>
-            {renderAccordion()}
+        <div className="main-container">
```
emmastephenson, line 188: We have snapshot tests for specific sections/components, but I don't think we have any test coverage for the side nav itself at the moment - could you add one? Either a snapshot test of the entire page or unit tests for the side nav specifically would work.

emmastephenson (changes requested):
Thanks Josh! Some comments, happy to chat through offline if that's best.

containers/ecr-viewer/src/app/view-data/page.tsx
JNygaard-Skylight: I moved the side nav code to its own file and added unit tests.

containers/ecr-viewer/src/app/view-data/page.tsx
JNygaard-Skylight: Added.

containers/ecr-viewer/src/app/view-data/page.tsx
JNygaard-Skylight: I created a `SectionConfig` class that I think should reduce the technical load for adding a new section.

containers/ecr-viewer/src/app/view-data/components/EcrSummary.tsx
```diff
@@ -23,7 +29,9 @@ const EcrSummary = ({ fhirPathMappings, fhirBundle }: EcrViewerProps) => {
         aria-labelledby="summary-box-key-information"
       >
         <div className="usa-summary-box__body">
-          <h3 className="summary-box-key-information">About the Patient</h3>
+          <h3 className="summary-box-key-information" id="about-the-patient">
+            About the Patient
```
emmastephenson, line 33: Would it be possible to reference the title as defined in the `eCRSummaryConfig` above? You might need to tweak the `SectionConfig` class definition, but I think then you could do something like

```html
<h3 className="summary-box-key-information" id={eCRSummaryConfig.subNav[1].id}>
            {eCRSummaryConfig.subNav[1].title}
</h3>
```

The reason I suggest it is that right now it'd be easy to only update the title in one place and have the side nav mismatch the actual content. I think this also applies to the `EcrMetadata` file, but `Demographics` and `SocialHistory` look good already :)

emmastephenson (changes requested):
This looks great Josh, thanks for making those changes! Just one other small request for changes.

containers/ecr-viewer/src/app/view-data/components/tests/SideNav.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+describe("SectionConfig", () => {
+  it("should create an instance with correct title and id", () => {
+    const section = new SectionConfig("Test Section");
+    expect(section.title).toBe("Test Section");
+    expect(section.id).toBe("test-section");
+  });
+
+  it("should handle subNavItems as strings and convert them to SectionConfig instances", () => {
+    const section = new SectionConfig("Parent Section", ["Child Section"]);
+    // @ts-ignore
```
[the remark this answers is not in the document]
JNygaard-Skylight: It was concerned that `subNavItems` could be null, but it should never be null, and if it is, we want the test to fail anyway. Just rewrote it to get rid of the TS ignores.

emmastephenson (approved):
Looks good to me, thank you Josh! 🙌 This will be great to have on the viewer.

The failing `build-containers` job should resolve itself if you merge with `main`, btw - that issue was from a bad PR I put out earlier this week, but it's resolved now.

Issue #1001, opened 2023-12-19
Extract patient travel history from eICR; add to eCR Viewer


emmastephenson:
## Action Requested
Add patient travel history to the eCR Viewer (currently it's stubbed).

## Acceptance Criteria
* When eCRs go through conversion to FHIR, patient travel history is extracted (if present)
* When eCRs are shown in the eCR Viewer, patient travel history is shown according to the designs (if present)

## Testing
We'll need some sample eCRs to test with that include travel history.

## Background & Context
The eCR Viewer designs call for including patient travel history, but it's not currently included in our FHIR conversion process.

## Additional Information
This is likely the file that will need to be updated within the converter: [SocialHistory/_entry.liquid](https://github.com/CDCgov/phdi/blob/main/containers/fhir-converter/Templates/eCR/Entry/SocialHistory/_entry.liquid)
And this file within the eCR Viewer will need the travel history FHIR path added: [fhirPath.yml](https://github.com/CDCgov/phdi/blob/main/containers/ecr-viewer/public/fhirPath.yml)

We will need to figure out where in the eCR CCDA spec patient travel history is found. The [FHIR implementation guide](https://build.fhir.org/ig/HL7/case-reporting/StructureDefinition-eicr-composition-definitions.html#diff_Composition.section:sliceSocialHistorySection.entry:sliceTravelHistory) for eICR composition may be helpful here.

PR #1048, opened 2024-01-10
resolve the value of references in fhir converter

BobanL:
# PULL REQUEST

## Summary
- In the eICR document, there are references that reference fields. These changes will set the text value of references to the matching ID's value.

Example:
There may be an observation for travel history that references `#trvhx-1`, but that data is located in a `table` outside of the entry
```XML
<entry>
    <act classCode="ACT" moodCode="EVN">
        <code code="420008001" codeSystem="2.16.840.1.113883.6.96" codeSystemName="SNOMED CT" displayName="Travel"/>
        <text>
            <reference value="#trvhx-1"/>
        </text>
        <effectiveTime>
            <low value="20180118"/>
            <high value="20180218"/>
        </effectiveTime>
    </act>
</entry>
```

```XML
<table>
    <thead>
        <tr>
            <th>Date of Travel</th>
            <th>Location</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>January 18th, 2018 - February 18th, 2018</td>
            <td ID="trvhx-1">Traveled to Singapore, Malaysia and Bali with<br/>my family.</td>
        </tr>
    </tbody>
</table>
```

Outcome:
```XML
<reference xmlns="urn:hl7-org:v3" xmlns:sdtc="urn:hl7-org:sdtc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" value="#trvhx-1">Traveled to Singapore, Malaysia and Bali with my family.</reference>
```

## Related Issue
Part of #1001

## Additional Information
- Using liquid templates, it is not possible as far as I am aware to fetch the value for that ID
- Set `build=True` for our integration test so the image rebuilds each time its run.

### DIFF
--- a/containers/fhir-converter/app/service.py
+++ b/containers/fhir-converter/app/service.py
@@ -2,6 +2,7 @@ import json
 import subprocess
 import uuid
 from pathlib import Path
+
 from lxml import etree
 
 from phdi.harmonization import standardize_hl7_datetimes
@@ -41,10 +42,10 @@ def add_data_source_to_bundle(bundle: dict, data_source: str) -> dict:
 
 def resolve_references(input_data: str):
     ecr = etree.fromstring(input_data)
-    ns = {'hl7': 'urn:hl7-org:v3'}
-    refs = ecr.xpath('//hl7:reference', namespaces=ns)
+    ns = {"hl7": "urn:hl7-org:v3"}
+    refs = ecr.xpath("//hl7:reference", namespaces=ns)
     for i in range(len(refs)):
-        ref_id = refs[i].attrib['value'][1:]
+        ref_id = refs[i].attrib["value"][1:]
         value = " ".join(ecr.xpath("//*[@ID='" + ref_id + "']/text()"))
         refs[i].text = value
 
--- a/containers/fhir-converter/app/service.py
+++ b/containers/fhir-converter/app/service.py
@@ -2,6 +2,7 @@ import json
 import subprocess
 import uuid
 from pathlib import Path
+from lxml import etree
 
 from phdi.harmonization import standardize_hl7_datetimes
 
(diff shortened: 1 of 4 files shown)


containers/fhir-converter/app/service.py
```diff
@@ -38,6 +40,22 @@ def add_data_source_to_bundle(bundle: dict, data_source: str) -> dict:
...
 
+def resolve_references(input_data: str):
+    try:
+        ecr = etree.fromstring(input_data.encode())
+    except etree.XMLSyntaxError:
+        return input_data
+
+    ns = {"hl7": "urn:hl7-org:v3"}
+    refs = ecr.xpath("//hl7:reference", namespaces=ns)
+    for i in range(len(refs)):
```
emmastephenson, line 51: Very minor nit: I think you could change this to a regular for-each loop instead (`for i in refs`)

emmastephenson (dismissed):
Looks good! And I appreciate the testing updates. 

One quick question - why the namespaces on the reference object in your example? All the `xmlns` tags didn't appear to be there before.
`<reference xmlns="urn:hl7-org:v3" xmlns:sdtc="urn:hl7-org:sdtc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" value="#trvhx-1">Traveled to Singapore, Malaysia and Bali with my family.</reference>`

BobanL:
> Looks good! And I appreciate the testing updates.
> 
> One quick question - why the namespaces on the reference object in your example? All the `xmlns` tags didn't appear to be there before. `<reference xmlns="urn:hl7-org:v3" xmlns:sdtc="urn:hl7-org:sdtc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" value="#trvhx-1">Traveled to Singapore, Malaysia and Bali with my family.</reference>`

Good call out! I looked into this and it seems like it puts all of those additional tags on the top level when searching with lxml, so something like `etree.tostring(tree.xpath("//hl7:originalText", namespaces={"hl7": "urn:hl7-org:v3"})[0])` will produce this:
```XML
<originalText xmlns="urn:hl7-org:v3" xmlns:sdtc="urn:hl7-org:sdtc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><reference value="#birthsex">Female</reference></originalText>
```

emmastephenson (approved):
LGTM!

PR #1049, opened 2024-01-10
Add PHDCBuilder skeleton and `_build_telecom` method to PHDCBuilder

m-goggins:
# PULL REQUEST

## Summary
This PR sets up the PHDCBuilder class and adds an internal method for building out the telecom section of a `recordTarget`. 

## Related Issue
Fixes # [oops, put it in phdi-azure](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi-azure/392)

## Additional Information

- This PR assumes a single phone is added; we should probably have another PR for handling and building multiple telecom instances
- The phone `uses` were based on examples in the PHDC spec but there could be additional example of `use` and I am not 100% certain that `MC` is for mobile phones (Mobile Cell?) but it was my best guess.

## Checklist


containers/message-parser/app/phdc/phdc.py
```diff
@@ -0,0 +1,37 @@
...
+        self.header = builder.header
+
+
+class PHDCBuilder:
+    def __init__(self):
+        self.header = None
+
+    def _build_telecom(
+        phone: str,
+        use: Literal["HP", "WP", "MC"] = None,
```
DanPaseltiner, line 17: At some point we'll need to map between the FHIR and CDA values for `use`. Would we want to do that in this function or address elsewhere? I don't think this is blocking just trying to keep this on our radar.

containers/message-parser/app/phdc/phdc.py
DanPaseltiner, line 1: Are we skipping the director class for now?

containers/message-parser/app/phdc/phdc.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        telecom_data = ET.Element("telecom")
+
+        if use is not None:
+            telecom_data.set("use", use)
+        telecom_data.set("value", phone)
+
+        return telecom_data
+
+    def build(self):
```
DanPaseltiner, line 36: Flagging that we don't have test coverage here yet, but probably okay for now. We should add it soon when things progress a little further.

DanPaseltiner (approved):
Really exciting to see code for building PHDCs begin to land in the codebase. I left a few comments on things I noticed, but I realize this is suppose to be a simple lightweight framework to get us started and I don't think any of them are blocking at this point.

containers/message-parser/app/phdc/phdc.py
m-goggins: I think elsewhere. There will also be `use` for address which appears to have similar but slightly different values than `telecom`. I can add a ticket for this into the backlog.

containers/message-parser/app/phdc/phdc.py
m-goggins: I had skipped it for now given that it wasn't necessary for this helper function but I was still thinking that it should be eventually included.

m-goggins: Happy to include it now or add as a ticket in the backlog.

DanPaseltiner: Let's put a ticket in the backlog. Once are building at least a "base" PHDC that includes the required boiler plate I think adding this as well as the additional tests mentioned in the other conversation will be easier.

containers/message-parser/app/phdc/phdc.py
DanPaseltiner: Oh good call about use for address. Should we make that ticket a spike to determine how we should handle `use`?

m-goggins: Yep, I can add that as well!

DanPaseltiner (approved):

PR #1061, opened 2024-01-11
Demographics Unavailable Info

### LINKED ISSUE #1015: Fix the Unavailable Info and N/A rendering
In the previous iteration of the design, the `N/A` was not rendered on the section they were supposed to be. In the latest design they are supposed to render in the section they are supposed to be.

Example:

https://www.figma.com/file/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?type=design&node-id=435-671&mode=design&t=S6kvOiZl3UJ6ypyy-0

If the facility type does not have an entry. It displays `Facility Type: N/A` twice. The current implementation does not do that.

The sections to fix are:
Demographics
Encounter
Social
Provider

KennethSkylight:
# PULL REQUEST

## Summary
~~Previously when there was some unavailable data, it would not render where it's supposed to. It would only appear in Unavailable Info. Following this PR, all fields will appear where they are supposed to with an N/A in it's place.~~ Also demographics did not follow the pattern laid out in social history and encounter.

I have also made it so that the evaluates and formatting happen before it is passed to the components. I'm not sure if it would be better to have the components know what data they have automatically?

## Related Issue
Fixes #1015 

## Screenshots

image

### DIFF
--- a/containers/ecr-viewer/src/app/utils.tsx
+++ b/containers/ecr-viewer/src/app/utils.tsx
@@ -262,7 +262,7 @@ export const evaluateDemographicsData = (
       value: evaluate(fhirBundle, mappings.patientEthnicity)[0],
     },
     {
-      title: "Tribal",
+      title: "Tribal Affiliation",
       value: evaluate(fhirBundle, mappings.patientTribalAffiliation)[0],
     },
     {
--- a/containers/ecr-viewer/src/app/utils.tsx
+++ b/containers/ecr-viewer/src/app/utils.tsx
@@ -361,14 +361,15 @@ export const evaluateProviderData = (
 };
 
 const evaluateData = (data: DisplayData[]) => {
-  let evaluatedData: DisplayData[] = [];
+  let availableData: DisplayData[] = [];
   let unavailableArray: DisplayData[] = [];
   data.forEach((item) => {
     if (item.value == undefined) {
       unavailableArray.push(item);
       item.value = "N/A";
+    } else {
+      availableData.push(item);
     }
-    evaluatedData.push(item);
   });
-  return { evaluated_data: evaluatedData, unavailable_data: unavailableArray };
+  return { availableData: availableData, unavailable_data: unavailableArray };
 };
--- a/containers/ecr-viewer/src/app/utils.tsx
+++ b/containers/ecr-viewer/src/app/utils.tsx
@@ -97,13 +97,10 @@ export const extractFacilityContactInfo = (
     fhirBundle,
     fhirPathMappings,
   );
-  console.log("location resource", locationResource);
   const phoneNumbers = locationResource.telecom?.filter(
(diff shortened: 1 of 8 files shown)


sarahtress:
@KennethSkylight oh shoot I'm sorry I misunderstood what you meant with the Unavailable Info fix. the current design has the unavailable info available within all of the sections just so that you all know what order to put the fields in - it's not how it should actually work though. if the data is unavailable, the field should *only* show up in the unavailable info section (basically, N/A sections should only be at the very bottom) sorry for the unnecessary work!!

KennethSkylight:
@sarahtress all good, that's an easy revert. Most of this PR is getting demographics info in unavailable info anyways

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -97,10 +97,13 @@ export const extractFacilityContactInfo = (
     fhirBundle,
     fhirPathMappings,
   );
-  const phoneNumbers = locationResource.telecom.filter(
+  console.log("location resource", locationResource);
```
BobanL, line 100: Can you remove this log?

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -97,10 +97,13 @@ export const extractFacilityContactInfo = (
...
+  const phoneNumbers = locationResource.telecom?.filter(
     (contact: any) => contact.system === "phone",
   );
-  return phoneNumbers[0].value;
+  return phoneNumbers && phoneNumbers.length > 0
+    ? phoneNumbers[0].value
+    : undefined;
```
BobanL, line 106: I think this can be simplified by using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_expressions)
`a?.[0]?.value`

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -123,8 +126,11 @@ export const formatPatientContactInfo = (
     .map((email) => `${email.value}`)
     .join("\n");
 
-  return `${phoneNumbers}
+  const formattedContactInfo = `${phoneNumbers}
     ${emails}`;
+
+  // if it's an empty string, return undefined
+  return formattedContactInfo.trim().length ? formattedContactInfo : undefined;
```
BobanL, line 133: I think it might make more sense here to return `formattedContactInfo.trim()` and then in the [evaluate data function](https://github.com/CDCgov/phdi/pull/1061/files#diff-0048e1f03ec47b0b2e09b3981316a88166682838cdf495fd5b369f375ad25183R367), just check all falsey values instead of undefined.

BobanL (commented):
Mostly looks good, but just a couple of things in the utils.

BobanL (approved):

PR #1067, opened 2024-01-11
add a catch all error

### LINKED ISSUE #877: Implement the catch-all error for the upload page
## Action Requested
Implement the generic error for the upload page.

## Acceptance Criteria
If an unexpected error is thrown on the upload page (can't connect to the backend, etc) the error message "We could not upload your file at this time, please try again" is [shown](https://www.figma.com/file/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Demo-Site?type=design&node-id=212-2043&mode=design&t=QJqOS9RF0ULUPXSa-0).

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2hiQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ba935b395adb35464de34fe396baa9b8550605b5/Screenshot%202023-10-20%20at%2011.56.17%20AM.png)

## Testing
Unit testing

## Background & Context
This is the catch-all error for the upload page - if something goes wrong unexpectedly, this is what should be shown.

zdeveloper:
# PULL REQUEST

## Summary
What changes are being proposed? catch all error handling

## Related Issue
Fixes #877
frontend PR https://github.com/CDCgov/phdi-playground/pull/46

### DIFF
--- a/containers/orchestration/app/main.py
+++ b/containers/orchestration/app/main.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 from pathlib import Path
 from typing import Annotated
@@ -15,6 +16,7 @@ from app.models import ProcessMessageRequest
 from app.models import ProcessMessageResponse
 from app.models import PutConfigResponse
 from app.services import call_apis
+from app.services import validate_response
 from app.utils import load_config_assets
 from app.utils import load_processing_config
 from app.utils import unzip_http
@@ -30,6 +32,8 @@ from fastapi import WebSocket
 from fastapi import WebSocketDisconnect
 
 from phdi.containers.base_service import BaseService
+
+logger = logging.getLogger(__name__)
 
 # Read settings immediately to fail fast in case there are invalid values.
 get_settings()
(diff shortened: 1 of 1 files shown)


containers/orchestration/app/main.py
```diff
@@ -83,26 +87,29 @@ async def process_message_endpoint_ws(
             response, responses = await call_apis(
                 config=processing_config, input=initial_input, websocket=websocket
             )
-            if response.status_code == 200:
+            if validate_response(response=response):
```
emmastephenson, line 90: I think this `validate_response` function is actually only for the processing the response from the validation service, not the responses of services more generally - I don't think all of our services use `message_valid` in the response. (Could be wrong, though!)

zdeveloper: that is correct, since the `response` is the last response we get from the services, 

in case of a validation failure, it will be a validation response, in that case, it will be incorrect if we only check the status code.

`validate_response` is used on all services, the `message_valid` check is optional, it will check the status code if it doesnt exisit
https://github.com/CDCgov/phdi/blob/303922c3af8b6e17c1b33fed77a26daa211f963d/containers/orchestration/app/services.py#L27-L34

emmastephenson: Gotcha, thanks for clarifying!

emmastephenson (approved):
LGTM 👍

PR #1094, opened 2024-01-18
Convert all ELR OBX segments to FHIR

### LINKED ISSUE #1059: Convert all ELR OBX segments to FHIR
We recently received feedback from LAC that the FHIR converter is not converting the data from all OBX segments to FHIR. To address this we should create our own ELR specific template that builds on the `ORU_R01` template provided by Microsoft.

Acceptance Criteria:
GIVEN an ELR containing multiple non-consecutive OBX segments associated with multiple OBR segments
WHEN we convert to FHIR
THEN data from all OBX segments are included in the resulting FHIR bundle as `Observation` resources

robertmitchellv:
# PULL REQUEST

## Summary
Edit the Microsoft FHIR liquid template for the `ORU_R01` HL7v2 message type to be able to handle multiple OBX segments.

## Related Issue
Fixes #1059 

## Additional Information

## Checklist


DanPaseltiner (commented):
Since we only changed the root template, I wonder if we should just add `ELR.liquid` into `Templates/hl7v2` and not duplicate all of the supporting templates which have not required modification.

robertmitchellv:
@DanPaseltiner i think we'd need to rework `convert_to_fhir` in `service.py` since `root_template` there is just the name of the `.liquid` template we want to use in the `template_directory_path`. because the templates reference the other files they need to be in a similar directory/path structure.

at some point we could add only the templates we care about here and remove the copying of all the templates that is happening in the Dockerfile--no matter what we're still cloning and copying all of these templates. also building this container cold locally without any previous builds takes a good bit of time.

DanPaseltiner (approved):
LGTM. Thanks for formalizing and testing this. 🚀

PR #1129, opened 2024-01-22
Wicked path accessor

bamader:
# PULL REQUEST

## Summary
This PR creates a demo schema for use extracting a toy FHIR bundle into some internally represented pre-PHDC JSON. More importantly, though, this PR introduces some cool accessor functionality into our get parsers function. By default, `fhirpathpy` only allows compilation and evaluation of paths on literal FHIR resources, such as `Patient` and `Observation`. This gets really messy if we want bulk operations on multiple values of a data type, though, such as a `HumanName` or `Address` (both of which we do want to preserve for PHDC). This PR introduces some instance variable manipulation behind the scenes to treat the compiled FHIR path as a list-based accessor if we're trying to access a FHIR `DataType` and not a `Resource`, which allows us to maintain our pattern of `SecondarySchema` definitions for multiple elements of an Array. This schema, the sample bundle, and the new accessor pattern all based local tests on my machine, and I was able to build and successfully use a docker image of this new code with insomnia.


robertandremitchell (approved):
looks good to me. tested the demo FHIR bundle without issue.

Issue #1020, opened 2024-01-05
Design a sample input to the PHDC module


DanPaseltiner:
In order to move forward on development of the `phdc` module that will build a PHDC given data extracted from a FHIR bundle we need to understand the format of these extract data. Based on the existing capabilities of the message parser, the PHDC specification, and the 2 sample PHDCs we have let's create a first pass at defining the schema of the `dict` that will be used to pass values extracted from FHIR to the PHDC builder.

Acceptance Criteria:
A JSON file containing a sample input for the `phdc` module has been added to the test assets of the message parser.

zdeveloper:
Proposed input schema
```json
{
    "type": "Case Report | Contact Record | Lab Report | Morbidity Report",
    "patient.address": {
        "streetAddressLine1": "14 WILLOW GLEN",
        "streetAddressLine2": "Apt. 2",
        "city": "SANDY SPRINGS",
        "state": "TX",
        "postalCode": "30029",
        "county": "Fulton",
        "country": "USA",
        "type": "H: Home address | WP: Workplace"
    },
    "patient.name": {
        "prefix": "Mr.",
        "first": "Surma",
        "middle": "J",
        "family": "Singh",
        "suffix": "Jr.",
        "type": "L: Legal | P: Pseudonym"
    },
    "patient.gender": "Female",
    "patient.race": "White",
    "patient.ethnicity": "Hispanic or Latino",
    "patient.birthDate": "YYYY-MM-DD",
    "patient.telcom": [
        {
            "value":"tel:55354257 | fax:56542558 | mailto:opr@opr-sem.no",
            "type": "H: Home | WP: Work | MC: Mobile phone",
            "useablePeriod.low": "YYYY-MM-DD",
            "useablePeriod.high": "YYYY-MM-DD"
        }
    ],
    "author.time": "20150304",
    "author.assignedPerson": {
        "prefix": "Mr.",
        "first": "Mike",
        "middle": "J",
        "family": "Smith",
        "suffix": "Jr."
    },
    "custodian.representedCustodianOrganization": {
        "name": "STATE DEPT OF HEALTH",
        "phone": "1234567890",
        "streetAddressLine1": "14 WILLOW GLEN",
        "streetAddressLine2": "Apt. 2",
        "city": "SANDY SPRINGS",
        "state": "13",
        "postalCode": "30029",
        "county": "Fulton",
        "country": "840"
    },
    "clinicalInformation.observation": {
        "code": "",
        "code.system": "",
[...]

PR #1140, opened 2024-01-24
Add remaining fields, example data

bamader:
# PULL REQUEST

## Summary
This PR adds the rest of the fields for the PHDC schema specified [here](https://github.com/CDCgov/phdi/issues/1020). The sample data file is also expanded to showcase this new retrieval. We now have the full range of data type accessors, referenced resources, secondary schemas, and some additional flattening. 

## Additional Information
Two changes had to get made in main for this work:
1. we now type control the initial values for schema fields that rely on a secondary schema; checking if they're a list (and casting them if they're not) allows us to pull out information from non-array fields without duplicative retrieval
2. had to remove an error message if the fully built and compiled reference path returns an empty list, because this wasn't accounting for the possibility that one or more fields on the secondary referenced resource might not be present or have the desired information; e.g., for the extraction of the custodian organization's address, if street address line 2 doesn't exist, the whole process will fail. This is now no longer the case, and we just set this value to null as with all the other missing fields we encounter.

### DIFF
--- a/containers/message-parser/app/default_schemas/demo_phdc.json
+++ b/containers/message-parser/app/default_schemas/demo_phdc.json
@@ -138,12 +138,12 @@
         "data_type": "string",
         "nullable": true
       },
-      "usable_period_low": {
+      "useable_period_low": {
         "fhir_path": "ContactPoint.period.start",
         "data_type": "string",
         "nullable": true
       },
-      "usable_period_high": {
+      "useable_period_high": {
         "fhir_path": "ContactPoint.period.end",
         "data_type": "string",
         "nullable": true
--- a/containers/message-parser/app/default_schemas/demo_phdc.json
+++ b/containers/message-parser/app/default_schemas/demo_phdc.json
@@ -122,5 +122,187 @@
     "fhir_path": "Bundle.entry.resource.where(resourceType = 'Patient').birthDate",
     "data_type": "date",
     "nullable": true
+  },
+  "patient_telecom": {
+    "fhir_path": "Bundle.entry.resource.where(resourceType = 'Patient').telecom",
+    "data_type": "array",
+    "nullable": true,
+    "secondary_schema": {
+      "value": {
+        "fhir_path": "ContactPoint.value",
+        "data_type": "string",
+        "nullable": true
+      },
+      "type": {
+        "fhir_path": "ContactPoint.use",
+        "data_type": "string",
+        "nullable": true
+      },
+      "usable_period_low": {
+        "fhir_path": "ContactPoint.period.start",
+        "data_type": "string",
+        "nullable": true
+      },
+      "usable_period_high": {
+        "fhir_path": "ContactPoint.period.end",
+        "data_type": "string",
+        "nullable": true
+      }
+    }
+  },
+  "author_time": {
+    "fhir_path": "Bundle.entry.resource.where(resourceType = 'Composition').date",
+    "data_type": "datetime",
+    "nullable": true
+  },
+  "author_assigned_person": {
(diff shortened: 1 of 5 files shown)


containers/message-parser/app/default_schemas/demo_phdc.json
```diff
@@ -122,5 +122,187 @@
...
+        "data_type": "string",
+        "nullable": true
+      }
+    }
+  },
+  "author_time": {
+    "fhir_path": "Bundle.entry.resource.where(resourceType = 'Composition').date",
+    "data_type": "datetime",
+    "nullable": true
+  },
```
DanPaseltiner, line 157: This is completely fine for now. I want to flag that we need to determine if `author_time` is whenever the case data is exported from the source system, or when we produce the PHDC. If the later then we can set this internally. Again, completely fine for now just an implementation details we'll need to get some guidance on from our users.

containers/message-parser/app/default_schemas/demo_phdc.json
```diff
@@ -122,5 +122,187 @@
...
+      "country": {
+        "fhir_path": "Bundle.entry.resource.where(resourceType = 'Organization').where(id = '#REF#').contact.address[0].country",
+        "reference_lookup": "Composition.custodian.reference",
+        "data_type": "string",
+        "nullable": true
+      }
+    }
+  },
+  "clinical_information_observations": {
+    "fhir_path": "Bundle.entry.resource.where(resourceType='Observation')",
```
DanPaseltiner, line 257: I think we may need to filter more narrowly than just all `Observations`, but since we aren't sure how all the data will be represented in FHIR yet we can cross this bridge later.

DanPaseltiner (approved):
LGTM. Thanks for putting this together. I realize this kind of a work can be tedious. Left a couple comments flagging things we may need to consider in the future.

containers/message-parser/app/default_schemas/demo_phdc.json
bamader: Yeah I was torn here on whether to just filter for labs like we've done in the past, but since I'm not as familiar with what we want the PHDC to represent or what the users are looking for I left it open ended in case there might be clinical data contained in non-lab observations.

robertandremitchell (commented):
lgtm; test data worked without issue. I did see some small typo differences between your and Zedd's code (telcom v telecom, then of course underscore versus camelCase). Do we want these to have parity, or does it not matter much?

One field I did see missing was `"type": "case report"` from Zedd's sample data, but everything else looked good.

bamader:
@robertandremitchell We always want to be using snake casing with underscores in Python, at least for things that we can control. Camel casing is a no-no convention best left for other languages. The exception of course is the FHIR server and FHIR paths, which unfortunately use camel cased variable names, but we've made a very conscious choice to not do that in our own code. If you encounter any other camel cases during reviews or PRs, def flag them for changing to snake case!

As far as typos go, everything in this PR here should align as closely as possible to names actually derived from FHIR spec. Telecom is the appropriate variable to use, for example, and I even changed all instances of "usable" (which is the correct spelling in standard American English) to "useable" (with the e, which is reserved for British English usage) because FHIR responses utilize the with-e spelling.

Finally, you're totally correct that there's no type field in this schema yet. I don't know if there's a ticket for it yet, but Dan and I decided last week that this isn't something the extraction parser can determine on its own, so we're going to alter the endpoint to ask the user to supply this parameter rather than try to experimentally determine it and then just pass it through. This also gives us the advantage of allowing the user's specification of this value to choose which PHDC template we load, once we implement multiple types. All good callouts though!

robertandremitchell (approved):

PR #1141, opened 2024-01-25
Refactor phdc builder

DanPaseltiner:
# PULL REQUEST

## Summary
This PR cleans up and refactors our FHIR to PHDC work after the push to get things working end to end. Notably we:

1. Add doc strings to all methods and functions
2. No longer define functions within other functions
3. Clean up the interaction between `PHDC` and PHDCBuilder
4. Moving our parsing and transformation into the PHDCBuilder `InputData` class logic into `utils`
5. Reorganize the `phdc` module

There is probably still work to be done here, but this is hopefully a good start.

## Related Issue
Fixes #

## Checklist


containers/message-parser/app/main.py
```diff
@@ -275,32 +146,16 @@ async def fhir_to_phdc_endpoint(
...
-            input_data.patient.birth_time = value
+    # 3. Transform to PHDCbuilder data classes
+    input_data = transform_to_phdc_input_data(parsed_values)
 
-    # Build the PHDC
-    print(input_data)
+    # 4. Build PHDC
     builder = PHDCBuilder()
     builder.set_input_data(input_data)
+    builder.build_header()
```
zdeveloper, line 158: I feel like this is should be handled inside the build method of the builder but its not blocking

zdeveloper (approved):
LGTM! thank you for the cleanup.

DanPaseltiner:
> From a technical standpoint, this all looks good to me. I'll just offer the following overall feedback, which I definitely don't feel is a must for us, but is still a perspective I'd like to raise.
> 
> To me, the PHDC builder is a bit long, cumbersome, and not particularly pythonic. If we take a step back from the fact that we're working with XML (meaning some of each of those traits is frankly unavoidable) and just consider the general question of "what do we actually want and need to do", our objective boils down to: extract a bunch of information from a FHIR bundle and format that data in one particular way. The objective itself (and our users) won't really care how we accomplish that, just that we do and that it's clean and robust.
> 
> What that means is that patterns (i.e. the Builder pattern) and specific instantiations of patterns only have value insofar as they directly advance that objective. What's written here definitely works (and no matter was still a big lift and so a major win for us), but I think it tries to go above and beyond what our objective actually calls for as necessary. We have two classes that are essentially interdependent on one another, meaning it's not clear how to just _do_ the thing we want to do (when I was briefly working on altering the endpoint to return PHDC, before the pairing call, it wasn't clear at all what the most efficient way to proceed was--do I instantiate a builder, then call PHDC? Do I call PHDC first and pass it some data and then separately call a builder? That complexity is still present here). We have a suite of Data Classes that to my eyes aren't particularly adding any value to the process of formatting data. Sure, they group input fields into related clusters, but if we plan to have a method to build each element anyway, why do we also need to create a specific structure for that method's input?
> 
> All of this adds up to something that I think gets away from what a pythonic expression of a builder pattern might be. In the abstract, something that Builds something else transforms a host of raw materials into a finished product. There are individual processes for building the parts of that finished entity--with a house, laying the foundation, putting up the walls, installing the electric, etc.--but the actual process of _building_ from 0 to 100 will naturally do that. I think this "intuitive" conception of what it means to be a Builder is also the naturally Pythonic way to do things.
> 
> What I mean by that is as follows. I would have a single class called `PHDCBuilder`. This class has a constructor that does the equivalent of building the "base" phdc (as a function in this PR does) document, with nothing else inserted into it. Then, the class has a series of private helper methods, each of which is responsible for building one element (or type of element) that appears in the output. We largely have this already coded up in our "_build_address" and "_build_patient" and the like. The difference to me is that we have no need of data classes whatsoever. Each private helper method receives as input the entirety of the extracted data from the parsing service--we already have it stored in a structure, so there's no reason to allocate another object whose data is useful for one function. This also fulfills the more pythonic and intuitive convention of just having one method of a class be responsible for transforming portions of those raw materials into a finished good. These helper methods are all united by a single publicly exposed function, `.build_phdc`, which takes as input the extracted data from the parsing service and hands it off to each private helper in sequence, where logic will transform it into a finished element.
> 
> To me, this reduces the overall number of classes, unifies the output into a single obvious method that creates and returns PHDC, does away with a bunch of middleman data structures, and puts the logic for going from raw to finished in the place I as an engineer would most expect to look (e.g. if I want to see how to build a patient PHDC element, I need only look at `._build_patient()`, I don't have to check out `.build_patient` but then look up what a `Patient data class` is and go to `create_phdc_input_structs` to figure out how we grab certain fields).
> 
> All of this again comes with the caveat that I don't feel super strongly this way and the above is just me trying to explain my position rather than actively ranting. This was a ton of work to put into our service, and I know it's much easier to propose edits from the outside than it is to actually get this written and produced. Just wanted to offer some food for thought on how we might be able to clean this up even more, if we're so inclined. Either way, though, this PR and the work it's refactoring are great work that I'm glad we can build on!

Thanks for this @bamader I'll do my best to respond.

The reason for the two separate classes is that there are things that a PHDC should "do" that are unrelated to it's construction e.g. producing a serialized version of itself. We did have a circular dependency issue between these two classes that @zdeveloper pointed out which has now been resolved. We use the `PHDCbuilder` class to construct a PHDC object  returned by the `.build()` method. After we have constructed a PHDC there is no longer a need to lug around all of the methods related to it construction, just those that do the things we would like to do with the PHDC now that we have it.

The reason for not implementing a`.build_phdc` method that calls all of the appropriate internal methods is that we will eventually need to build different kinds of PHDC documents we are starting with case reports, but there are also contact records, lab reports, and morbidity reports. Although we have not implemented this yet, a director class will have the knowledge to build different types of PHDCs. The benefit to this is that we can have a separation of concerns between the builder and the director. The builder just knows how to construct different parts of the PHDC the director knows which parts are necessary for different types of PHDC.

containers/message-parser/app/main.py
DanPaseltiner: I hadn't focused too much on this, because I figured this is what would eventually be handled by a director class, but I'm going to move it insider as you suggest. This is certainly cleaner and we can refactor later as needed.

bamader:
@DanPaseltiner Great point about the different types of PHDC. I forgot about the eventual use of a director to make internal logic decisions about which template format we'll want to output. While I think the data classes are still a bit clunky, I'm convinced that not having a singular `.built_phdc()` function is completely fine now, as is the separation of things like serializing from construction. Thanks for the response!

robertmitchellv (approved):
i think this continues to move us forward in a good direction. the thing that remains a _known unknown_ for us is the complexity as we get closer to live prod data ➡️ our FHIR to PHDC conversion ➡️ into a live prod system. so this builder pattern does give us some flexibility in a couple of different ways:

1. **the work is already done** and we have input ➡️ output, so it makes sense to leave it as is for the time being and to gradually add complexity to our implementation
2. **we don't know how complex our implementation _will_ become**, so if we simplify + reduce the pattern now and it turns out we actually need some of these seemingly verbose structures it'll be a painful refactor. for example, we may end up needing to implement more than just a simple case report PHDC.

this feels like a really good foundation to continue to build off of 🏗️

containers/message-parser/app/main.py
DanPaseltiner: Done!

PR #1148, opened 2024-01-25
add script to convert eicr and rr data to fhir

BobanL:
# PULL REQUEST

## Summary
- Found and added all of the documents to the `baseEcr` folder
- Created a script that will start the fhir-converter and convert all baseEcr documents to fhir. 
- Run it with `sh ./containers/ecr-viewer/seed-scripts/convert.sh` or `npm run convert-seed-data`

## Related Issue
Fixes #

## Additional Information
- Each time the script is run some IDs get regenerated causing additional changes to be made.


nickbristow (approved):
Do we need to update readme for this?

emmastephenson (approved):
:D Thank you! This will be really helpful for new changes!

Issue #1142, opened 2024-01-25
Update _build_base_phdc to include and empty structuredBody element


DanPaseltiner:
When we produce an empty PHDC during initialization of `PHDCbuilder` we do not currently created an empty body. We will need this so that the soon to be implemented `build_clinical_information` method has a place to put the data it will produce.

GIVEN the `PHDCbuilder` class
WHEN the `_build_base_phdc()` method is called 
THEN  a ET.ElementTree object representing the following XML is returned.

```
<?xml version='1.0' encoding='utf-8'?>
<?xml-stylesheet type="text/xsl" href="PHDC.xsl"?>
<ClinicalDocument xmlns="urn:hl7-org:v3" xmlns:sdt="urn:hl7-org:sdtc" xmlns:sdtcxmlnamespaceholder="urn:hl7-org:v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:hl7-org:v3 CDA_SDTC.xsd">
	<component>
		<structuredBody>
		</structuredBody>
	</component>
</ClinicalDocument>

```

DanPaseltiner:
This may require tweaking the `builder_header()` method it ensure it inserts header data in the correct location. Failing tests should indicate is this becomes an issue.

robertmitchellv:
i implemented this and worked through some of the tests locally that were failing and it would've required more tweaking since the `<ClinicalDocument>` is the root and when we append to the root in `build_header` it places the elements _after_ the `<component>` and `<structuredBody>` tags from `_build_base_phdc`, which doesn't follow the structure from our example PHDC documents. 

it makes sense to me to have the `_build_base_phdc` simply create the root and then have the `build_header` and `build_body` create those sections as factories to ensure that the methods are being applied to the correct sections. i was in the middle of doing this when i looked more closely at Zedd's PR while helping with diagnosing the issue with the pre-commit bot and the linting workflow. i'm not sure what your thoughts are on this; maybe you disagree completely, but an alternative would be to let Zedd's PR merge to main and then work on either: 
* retrofitting the `<component>` and `<structuredBody>` into `_build_base_phdc`; or,
* a combination of that _plus_ picking the Director back up

let me know what you think

robertmitchellv:
Solved in #1163

PR #1163, opened 2024-01-26
add body and case_report section without observations

### LINKED ISSUE #1055: Develop PHDC `clinicalInformation` builder
The clinical information section is required for building a PHDC case report. We need to construct a template data class that can be populated with data coming out of the message parser.

Acceptance Criteria:
A clinical information section can be added to a PHDC case report that minimally includes the following elements with repeatable entries:
- id 
- code
- entry with observation, code, and optional value(s)

Example:
```
	<component>
		<structuredBody>
			<component>
				<section>
					<id extension="0610000005-79CAS" assigningAuthorityName="LR"/>
					<code code="55752-0" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" displayName="Clinical Information"/>
					<title>Clinical Information</title>
					<entry typeCode="COMP">
						<observation classCode="OBS" moodCode="EVN">
							<code code="INV180" codeSystem="2.16.840.1.114222.4.5.1" displayName="Investigator"/>
							<value xsi:type="II" root="2.3.3.3.322.23.34" extension="Provider-101"/>
						</observation>
					</entry>
					<entry typeCode="COMP">
						<observation classCode="OBS" moodCode="EVN">
							<code code="INV515" codeSystem="2.16.840.1.114222.4.5.1" displayName="Binational Reporting Criteria"/>
							<value xsi:type="CE" code="PHC1140" codeSystem="1.2.3.5" displayName="Exposure to suspected product from Canada or Mexico"/>
							<value xsi:type="CE" code="PHC1139" codeSystem="1.2.3.5" displayName="Has case contacts in or from Mexico or Canada"/>
							<value xsi:type="CE" code="PHC1141" codeSystem="1.2.3.5" displayName="Other situations that may require binational notification or coordination of response"/>
						</observation>
					</entry>
				</section>
			</component>
		</structuredBody>
	</component>
```

zdeveloper:
# PULL REQUEST

## Summary
What changes are being proposed? add PHDC body and case_report section without observations

## Related Issue
Fixes #1055, #1142

## Checklist

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip
-          pip install black flake8
+          pip install black=23.12.1 flake8=7.0.0
       - name: Check format (black)
         run: |
           black --check --verbose .
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip
-          pip install black=23.12.1 flake8=7.0.0
+          pip install black==23.12.1 flake8==7.0.0
       - name: Check format (black)
         run: |
           black --check --verbose .
--- a/containers/message-parser/app/phdc/builder.py
+++ b/containers/message-parser/app/phdc/builder.py
@@ -187,8 +187,6 @@ class PHDCBuilder:
 
         self.phdc.getroot().append(body)
 
-        # return body
-
     def _build_case_report(self):
         component = ET.Element("component")
         section = ET.Element("section")
--- a/containers/message-parser/app/phdc/builder.py
+++ b/containers/message-parser/app/phdc/builder.py
@@ -170,6 +170,48 @@ class PHDCBuilder:
             )
         )
 
+    def build_body(self):
+        body = ET.Element("component")
+        structured_body = ET.Element("structuredBody")
+        body.append(structured_body)
+
+        match self.input_data.type:
+            case "case_report":
+                body.append(self._build_case_report())
+            case "contact_record":
+                pass
+            case "lab_report":
+                pass
+            case "morbidity_report":
+                pass
+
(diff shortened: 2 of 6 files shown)


containers/message-parser/app/phdc/builder.py
```diff
@@ -170,6 +170,46 @@ def build_header(self):
             )
         )
 
+    def build_body(self):
+        body = ET.Element("component")
+        structured_body = ET.Element("structuredBody")
```
robertmitchellv, line 175: technically this solves #1142 which doesn't quite work if we merely insert a `<component>` and `<structuredBody>` into the `_build_base_phdc` method like the ticket says (it causes a lot of tests to fail related to header and inserts those elements in the wrong places). 

i was in the middle of implementing something similar with a "build_body" to work alongside "build_header" before i wanted to see what you and Rob were doing to be inline when i saw this! so i think it's best continue in this direction and to have your PR also fix #1142

.github/workflows/lint.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip
-          pip install black flake8
+          pip install black==23.12.1 flake8==7.0.0
```
robertmitchellv, line 36: glad this worked!

robertmitchellv (approved):
i like the approach of separating a `build_header` and `build_body`. i also think we should add #1142 to the fixes; i like adding both the `<component>` and `<structuredBody>` within a `build_body` method so we can better construct parent and child relationships in the right order.

looks good Zedd 🚀

containers/message-parser/app/phdc/builder.py
DanPaseltiner: If we update #1142 to have an "empty body", then the method below that is currently `_build_case_report` can be called `build_clinical_information` and handle appending directly to `self.phdc`. Then in either the `build` method or a director class in the future we handle building the appropriate sections for a given type of case report.

PR #1186, opened 2024-01-30
Add `<title>` to `build_header()`

### LINKED ISSUE #1160: Add title to PHDC header
Currently our PHDC header does not include `title`. We need to add this element and for now we can hard code the value to `"Public Health Case Report - Data from the DIBBs FHIR to PHDC Converter"`.

Acceptance Criteria:

GIVEN the `build_header()` method 
WHEN it is called
THEN `title` is set as shown in the example below

```
<ClinicalDocument
	xmlns="urn:hl7-org:v3"
	xmlns:sdt="urn:hl7-org:sdtc"
	xmlns:sdtcxmlnamespaceholder="urn:hl7-org:v3"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:hl7-org:v3 CDA_SDTC.xsd">
	<realmCode code="US"></realmCode>
	<typeId root="2.16.840.1.113883.1.3" extension="POCD_HD000040"></typeId>
	<id root="2.16.840.1.113883.19" extension="CAS10124000GA01" assigningAuthorityName="LR"></id>
	<code code="55751-2" codeSystem="2.16.840.1.113883.6.1" codeSystemName="LOINC" displayName="Public Health Case Report - PHRI"></code>
	<title>Public Health Case Report - Data from the DIBBs FHIR to PHDC Converter</title>
	<effectiveTime value="20240124151751"></effectiveTime>
	<confidentialityCode code="N" codeSystem="2.16.840.1.113883.5.25"></confidentialityCode>
	<setId extension="ONGOING_CASE" displayable="true"></setId>
	<versionNumber value="1"></versionNumber>
```

robertmitchellv:
# PULL REQUEST

## Summary
Add `<title>` to the PHDC header's `build_header()` method. This value is hard-coded to: `"Public Health Case Report - Data from the DIBBs FHIR to PHDC Converter"`. This adds a minimum level of provenance to the PHDC creation process, as it denotes that the data was processed via the FHIR to PHDC Converter endpoint.

## Related Issue
Fixes #1160

### DIFF
--- a/containers/message-parser/app/phdc/builder.py
+++ b/containers/message-parser/app/phdc/builder.py
@@ -148,6 +148,16 @@ class PHDCBuilder:
         code.set("displayName", "Public Health Case Report - PHRI")
         return code
 
+    def _get_title(self):
+        """
+        Returns the title element of the PHDC header.
+        """
+        title = ET.Element("title")
+        title.text = (
+            "Public Health Case Report - Data from the DIBBs FHIR to PHDC Converter"
+        )
+        return title
+
     def build_header(self):
         """
         Builds the header of the PHDC document.
@@ -156,6 +166,7 @@ class PHDCBuilder:
         root.append(self._get_type_id())
         root.append(self._get_id())
         root.append(self._get_case_report_code())
+        root.append(self._get_title())
         root.append(self._get_effective_time())
         root.append(self._get_confidentiality_code(confidentiality="normal"))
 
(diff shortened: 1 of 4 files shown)


m-goggins (commented):
Looks good to me! Wondering if you think it's necessary to add a specific `_get_title` test or not. I'm in the middle of a similar PR.

robertmitchellv:
> Wondering if you think it's necessary to add a specific `_get_title` test or not. I'm in the middle of a similar PR.

i figure since there are three separate files within `assets` that are all used to test this piece as well as the test for `build_header()` that it's a part of, that it would be a bit redundant but i'm open to adding one if other feel it's important

m-goggins:
I would add it if it's not too much work, just because it could be employed differently in a different PHDC document type in the future.

m-goggins (approved):

PR #1193, opened 2024-01-31
Add path rewrite middleware to BaseService

nickclyde:
# PULL REQUEST

## Summary
This adds a middleware to the BaseService that strips out the service name at the beginning of any request made to the container. EG, if I make a request to the FHIR converter container at `/fhir-converter/convert-to-fhir`, it will rout that request to `/convert-to-fhir`. This is needed in order to use Application Load Balancer Controllers as an ingress controller for EKS in AWS, because unlike App Gateways in Azure, they don't support path rewrites.


phdi/containers/base_service.py
```diff
@@ -64,6 +66,23 @@ def __init__(
...
+    def add_path_rewrite_middleware(self):
+        """
+        Add middleware to the FastAPI instance to strip the service_path
+        from the URL path if it is present. This is useful when the service
+        is behind a gateway that is using a path-based routing strategy.
+        """
+
+        @self.app.middleware("http")
+        async def rewrite_path(request, call_next):
+            if request.url.path.startswith(self.service_path):
```
emmastephenson, line 79: From what we talked about earlier, Azure automatically strips these leading strings already, right? So in those cases this `if` statement just wouldn't trip at all.

containers/alerts/app/main.py
emmastephenson, line 1: If we need to add this middleware to all the services, I think the FHIR converter will need it too. (It's not extending the BaseService IIRC, so maybe we can add the `add_path_rewrite_middleware` function directly?

emmastephenson (approved):
Looks good!

phdi/containers/base_service.py
nickclyde: Exactly!

containers/alerts/app/main.py
nickclyde: Aha! Good catch. Is there any reason we don't extend the BaseService with it? I seem to recall something about the license but I can't remember. If there's not a good reason not to, it would be "DRY"er to just switch it to use the BaseService.

emmastephenson: I can't remember either...I think @robertmitchellv was closest to that work, and might recall why we didn't use it for the FHIR converter. May have also been some wonkiness with the Docker image, but maybe I'm misremembering?

We'll also have to figure out how to do this for the eCR Viewer, but BaseService is a bit of a moot point there since it's not a Python server anyways.

DanPaseltiner (commented):
LGTM. Thanks for putting this together @nickclyde! What is our testing plan for AWS and Azure once this merges?

nickclyde:
@DanPaseltiner I can test on AWS manually using my branch that I still haven't pushed on `phdi-playground`. To test on Azure, we'll need to bump the version on `phdi-playground` and let it deploy to test everything is still working fine.

PR #1236, opened 2024-02-01
Add config name param

### LINKED ISSUE #1227: Modify the `/process` endpoint to accept a config_file_name parameter

bamader:
# PULL REQUEST

## Summary
This PR modifies the input to the orchestration service's `process` and `process-message` endpoints to enforce the user providing a configuration file name. This name is used to look up the relevant workflow on the back end for application to the data the user provided.

## Related Issue
Fixes #1227 

## Additional Information
This work will not touch the `process-ws` endpoint, since that involves directly sending zipped messages as filebytes and will need other work to alter the input.

### DIFF
--- a/containers/orchestration/app/main.py
+++ b/containers/orchestration/app/main.py
@@ -190,12 +190,17 @@ async def apply_workflow_to_message(
     """
     try:
         processing_config = load_processing_config(config_file_name)
-    except FileNotFoundError:
-        return {
-            "status_code": 400,
-            # "message": error.__str__(),
-            # "processed_values": {},
-        }
+    except FileNotFoundError as error:
+        response = Response(
+            content=json.dumps(
+                {
+                    "message": error.__str__(),
+                    "processed_values": {},
+                }
+            ),
+            status_code=status.HTTP_400_BAD_REQUEST,
+        )
+        return response
 
     api_input = {
         "message_type": message_type,
--- a/containers/orchestration/app/main.py
+++ b/containers/orchestration/app/main.py
@@ -190,11 +190,11 @@ async def apply_workflow_to_message(
     """
     try:
         processing_config = load_processing_config(config_file_name)
-    except FileNotFoundError as error:
+    except FileNotFoundError:
         return {
             "status_code": 400,
-            "message": error.__str__(),
-            "processed_values": {},
+            # "message": error.__str__(),
+            # "processed_values": {},
         }
 
     api_input = {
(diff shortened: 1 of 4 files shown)


robertandremitchell (approved):
looks good to me. incredibly ticky-tacky, but with how the pathing works, `pytest` at the orchestration-level does not work, but `pytest tests` does, or running `pytest` within `orchestration/tests`.

99% sure it's just a difference in pathing for two of the tests; not sure if we need to really tackle as part of this work, though.

bamader:
@robertandremitchell Oh that's weird, I didn't notice that until you called it out. Interestingly, the tests that are throwing issues in pathing are actually tests where the `process-message` call is supposed to succeed (`test_process_success` and `test_process_message_success`), which are two tests that I neither wrote nor touched aside from bringing their inputs in line.

I would probably vote to ignore it tbh lol. I typically run pytest from within the tests directory anyway, and this feels like something super not worth digging into right now. Def weird though!

containers/orchestration/app/models.py
```diff
@@ -17,6 +17,13 @@ class ProcessMessageRequest(BaseModel):
     message_type: Literal["ecr", "elr", "vxu", "fhir"] = Field(
         description="The type of message to be validated."
     )
+    config_file_name: str = Field(
```
emmastephenson, line 20: Will this config_file_name always be required as an input, or can it be optional and we use the existing file as the default value?

containers/orchestration/app/main.py
```diff
@@ -77,6 +77,7 @@ async def process_message_endpoint_ws(
             # Hardcoded message_type for MVP
             initial_input = {
                 "message_type": "ecr",
+                "config_file_name": "sample-orchestration-config.json",
```
emmastephenson, line 80: I don't think this line is necessary - this should just be the input we're passing along to the other building blocks. The configuration is loaded separately on line 86. But if I'm missing something, just let me know!

emmastephenson (commented):
Looks good - my only comment would be, have we considered making `sample-orchestration-configuration` the default config? Then we wouldn't need to update existing/local requests.

containers/orchestration/app/models.py
DanPaseltiner: From the standpoint of maintaining backwards compatibility of the API whenever possible I agree we should make this optional and default to the existing config file. However, we need to think through how not all incoming data formats will work with a single default file. I don't think this is a reason not to proceed, but maybe more of a consideration when we move over to working on the Demo UI.

PR #1237, opened 2024-02-01
Moving URL variable reference in orchestration-config to local-dev.env

### LINKED ISSUE #1194: Move Service URLs from service configs to environment variables in orchestration service
Service URLs should be specified as environment variables, not within workflow configs
Update existing config file accordingly and confirm tests still work.

https://github.com/CDCgov/phdi/blob/main/containers/orchestration/app/default_configs/sample-orchestration-config.json

GIVEN: Multiple services that have service URLs
WHEN: A .env specifies  those URLs
THEN: The config file will pull those URLs and continue to process data.

robertandremitchell:
# PULL REQUEST

## Summary
Moving URL references in https://github.com/CDCgov/phdi/blob/main/containers/orchestration/app/default_configs/sample-orchestration-config.json to be done via `local-dev.env`.

## Related Issue
Fixes #1194 .

## Additional Information
Curious if we also just rename this to be base as `.env`
Also, whether we should load_dotenv exclusively in the load_processing_config section. can see arguments for both?

## Checklist


robertandremitchell:
@nickbristow and @gordonfarrell , tagging more for visibility

bamader (commented):
This looks good to me, but I'd like an approval from either Nick or Gordon as well to make sure it doesn't mess with anything(which is why I'm just commenting instead of approving) (also might be worth tagging @DanielPaseltiner on this for more visibility). I would also be in favor of your proposed change, Rob, of just renaming it to `.env` and dropping the `local-dev`. Once we migrate to a docker compose file this can get cleaned up more but for now this is a good start.

containers/orchestration/app/utils.py
```diff
@@ -38,9 +48,32 @@ def load_processing_config(config_name: str) -> dict:
...
+    # more generalizable in the future
+    for settings in config.get("configurations", {}).values():
+        if "url" in settings and isinstance(settings["url"], str):
+            has_placeholders = True
+            break
+
+    if has_placeholders:
+        for settings in config.get("configurations", {}).values():
+            if "url" in settings and isinstance(settings["url"], str):
+                settings["url"] = os.path.expandvars(settings["url"])
```
DanPaseltiner, line 74: Having the same for loop in here twice `for settings in config.get("configurations", {}).values()` and checking this condition twice `if "url" in settings and isinstance(settings["url"], str)` seems like a code smell to me. I understand the theory of the first block to see if we can just skip this entire process, but I think we are over optmizing a bit here. So far at least one step will require `url` and so far all of these values need to be interpolated based on environment variables.

What do you think about something like this?

```suggestion
        for settings in config.get("configurations", {}).values():
            if "url" in settings:
                settings["url"] = os.path.expandvars(settings["url"])
```

DanPaseltiner, line 74: Having the same for loop in here twice `for settings in config.get("configurations", {}).values()` and checking this condition twice `if "url" in settings and isinstance(settings["url"], str)` seems like a code smell to me. I understand the theory of the first block to see if we can just skip this entire process, but I think we are over optmizing a bit here. So far at least one step will require `url` and so far all of these values need to be interpolated based on environment variables. I also don't think we need to ensure that the value of `url` is a string. This in theory should always be the case and if we are concerned about ensuring this we should handle it during a step where we validate config files before allowing the service to accept them.

What do you think about something like this?

```suggestion
        for settings in config.get("configurations", {}).values():
            if "url" in settings:
                settings["url"] = os.path.expandvars(settings["url"])
```

DanPaseltiner (changes requested):
I really like this approach! However, I do think there is a little room to simplify. I also think we now need to look at ensuring these env var are provided when the service spins up. We can do this by removing `Optional` from all of these values in `config.py`. Then when we call `get_settings()` in `main.py` we will check to see if all necessary urls have been specified and shut the service down if they have not. I think we need to do this because if we make these changes and the environment variables are not set we will not be able to update the config file appropriately leaving the service running but in a broken state.

containers/orchestration/app/utils.py
robertandremitchell: makes sense to me. I'm definitely guilty of over-optimizing at times. Fixed in latest.

robertandremitchell:
> This looks good to me, but I'd like an approval from either Nick or Gordon as well to make sure it doesn't mess with anything(which is why I'm just commenting instead of approving) (also might be worth tagging @DanielPaseltiner on this for more visibility). I would also be in favor of your proposed change, Rob, of just renaming it to `.env` and dropping the `local-dev`. Once we migrate to a docker compose file this can get cleaned up more but for now this is a good start.

makes sense. Updated to use .env.

DanPaseltiner (approved):
Nice work. I didn't know about `os.path.expandvars` previously. Nice find!

PR #1241, opened 2024-02-02
Rob/1196 write a docker compose file for the orchestration service

### LINKED ISSUE #1196: Write a docker compose file for the Orchestration service
BLOCKED: First confer with the Streamline team to determine why there may be a reason to continue to not use docker-compose.yml logic.

GIVEN: We have an orchestration service calling in multiple building blocks
WHEN: There is a docker-compose.yml determining how to call in those building blocks
THEN: The service will start

robertandremitchell:
# PULL REQUEST

## Summary
This would functionally replace the `local-dev.sh` as the primary way to start the orchestration service and tie together all the services.

## Related Issue
Fixes #1196 

## Additional Information
Still puzzling through ecr-viewer; we might be able to postpone work on this aspect in the short run.

edit: ecr-viewer _seems_ to be working fine, but I am not entirely sure the best way to test/confirm, so will be very interested in review here.

The previous testing protocol already used docker: https://github.com/CDCgov/phdi/blob/main/containers/orchestration/tests/integration/docker-compose.yaml. 

I am not entirely sure, reviewing the code, if there are any other changes or modifications that ought to be considered, given that.

## Checklist


robertandremitchell:
weirdly, the integration tests seem to be passing on git, but aren't for me locally... @bamader or @DanielPaseltiner , I'm curious if you have any trouble running either test, locally. In particular, because it deals with DB things, I'm wondering if it's because I'm lacking any particular local set up?

cc @gordonfarrell @emmastephenson

robertandremitchell:
If the tests work for others, before merging, I will remove `test.env` and the `tests/integration/docker-compose.yaml`. I've preserved them for now for ease of comparison, review.

containers/orchestration/docker-compose.yml
```diff
@@ -0,0 +1,72 @@
...
+  message-parser:
+    build:
+      context: ../message-parser/
+      no_cache: true
+    logging:
+      driver: "json-file"
+    ports:
+      - "${MESSAGE_PARSER_PORT_NUMBER}:8080"
+
+  ecr-viewer-db:
```
emmastephenson, line 41: I think there's an opportunity to DRY this up with the ecr-viewer's docker-compose file. What if instead of copying the job, we created a separate docker-compose just for the database? It would include both the database setup (this job) as well as the seed-db job currently included in the viewer. Then this file and the ecr-viewer compose could both reference the database-setup compose.

containers/orchestration/docker-compose.yml
```diff
@@ -0,0 +1,72 @@
...
+      retries: 20
+
+  orchestration:
+    build: .
+    command: python -m uvicorn app.main:app --host 0.0.0.0 --port 8080 --reload
+    ports:
+      - "${ORCHESTRATION_PORT_NUMBER}:8080"
+    logging:
+      driver: "json-file"
+    depends_on:
```
emmastephenson, line 66: If we do go the route of having a separate database-setup docker compose, we should make the database setup jobs another dependency here - I think you can [reference those jobs directly](https://docs.docker.com/compose/multiple-compose-files/include/#example).

containers/orchestration/tests/integration/conftest.py
```diff
@@ -10,8 +9,8 @@
 @pytest.fixture(scope="session")
 def setup(request):
     print("Setting up tests...")
-    dotenv_path = join(dirname(__file__), "test.env")
-    load_dotenv(dotenv_path)
+    dotenv_path = Path(__file__).resolve().parent.parent.parent / ".env"
```
emmastephenson, line 12: Probably would be best to delete the existing `docker-compose` file in the `test` folder as well - no need to have it around as dead code!

emmastephenson (changes requested):
I have a suggestion to tweak the database setup a bit, happy to chat more offline if you'd like!

DanPaseltiner (commented):
Looking really good. Thank @robertandremitchell. I think @emmastephenson sounds like the right approach here -- not that what you did was wrong.

containers/orchestration/docker-compose.yml
robertandremitchell: nice! yeah, I was struggling to get the logic for stringing together docker-composes, thank you for this resource. I agree that separating out the database work and seed-db work would work better separately. I will work to implement.

robertandremitchell:
@bamader , I know you've puzzled through these integration tests recently. I have 3 failing for what I suspect is the same reason. I can get `http://localhost:8080/process-message` to run after running `docker compose up` locally. however, I've been stuck on this assertionerror problem since yesterday afternoon.

As far as I can tell, the payload for the three tests failing should be up to date. Any clues why it might not be working?

cc @DanPaseltiner

robertandremitchell:
New docker-compose seems to be working. I am not entirely sure why the `include` logic does not seem to work on GitHub actions but does locally? https://docs.docker.com/compose/compose-file/14-include/

Curious if we want to try to get it to work or just delete `docker-compose-ecr-viewer.yml` for now in favor of the super-file? I can do some additional digging to see if there's any reason include would not work.

DanPaseltiner:
> New docker-compose seems to be working. I am not entirely sure why the `include` logic does not seem to work on GitHub actions but does locally? https://docs.docker.com/compose/compose-file/14-include/
> 
> Curious if we want to try to get it to work or just delete `docker-compose-ecr-viewer.yml` for now in favor of the super-file? I can do some additional digging to see if there's any reason include would not work.

@robertandremitchell I think we want `extends` instead of `include`, [these docs](https://docs.docker.com/compose/multiple-compose-files/extends/) might be helpful. Could you attempt to get this working? LMK if you'd like to pair.

robertandremitchell:
Argh! Still not working, but a promising clue, this time around:
```
Cannot extend service 'seed-db' in /home/runner/work/phdi/phdi/containers/orchestration/docker-compose-ecr-viewer.yml: services with 'depends_on' cannot be extended
```

robertandremitchell:
@DanPaseltiner  and @emmastephenson I got the `extend` logic to work!

LMK if this looks good. 🤞🏽

DanPaseltiner (approved):
LGTM. Is the intention here that we would eventually point this to a different docker compose file for the DB so we can eliminate `docker-compose-ecr-viewer.yml`?

Nice work here. I know this was a real slog with a lot of hair pulling. Thanks for sticking with it!

robertandremitchell:
> LGTM. Is the intention here that we would eventually point this to a different docker compose file for the DB so we can eliminate `docker-compose-ecr-viewer.yml`?
> 
> Nice work here. I know this was a real slog with a lot of hair pulling. Thanks for sticking with it!

Yeah. I don't think we can point to the `ecr-viewer` docker-compose.yml, at least in its current form, because you can't have an `extend` that also has a `depends_on`, which that file currently does.

containers/orchestration/docker-compose.yml
robertandremitchell: @emmastephenson , curious your thoughts on the current state with `extends`. As noted during our pairing, unfortunately at least at the moment, `includes` doesn't seem to play nicely with GitHub Actions.

emmastephenson: Interesting! I think this will definitely work for now, and it's easy enough to follow, so 👍 from me on the PR overall.

I do wonder if @rin-skylight has run into this `includes` issue with GH actions before?

PR #1244, opened 2024-02-05
Adding Vital Signs

### LINKED ISSUE #1113: Add Vital Signs to eCR Viewer
## Action Requested
Add the Vital Signs field to the eCR Viewer.
>eCR Viewer
>> Clinical Info
>>> Diagnostics and Vital Signs
>>>> Vital Signs

## Acceptance Criteria
Vital Signs are displayed for eCRs that have the field.

## Testing
Unit testing will be required.

## Background & Context
This field may be a list of observations rather than a single data point. If this is the case, please chat with Sarah about the best display options.

## Designs
[Figma link](https://www.figma.com/file/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?type=design&node-id=1024-136062&mode=design&t=sxIe5Umchzr49IHv-0)

## Additional Information
We'll need to test to see whether or not this field is already being converted to FHIR, or if we'll need to complete both FHIR conversion and adding to the eCR Viewer.

KennethSkylight:
This PR is a recreation of my first vital signs PR
# PULL REQUEST

## Summary
Adding the Vital Signs info as well as the Clinical Info section. You'll notice that the clinical info section sometimes has nothing but it's because we haven't added the diagnostic info yet either. 

It uses the same unavailable info pattern as other sections.

Note that the design has info such as pulse, blood pressure, and temperature. From sleuthing the LAC data, only Height, Weight, and BMI exist on the **vitals** section

## Related Issue
Fixes #1113 
## Screenshots

Showing that this has already been design reviewed

image


containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -226,6 +226,45 @@ const formatStartEndDateTime = (
...
+  weightMeasurementType: string,
+  bmi: string,
+) => {
+  let heightString = "";
+  let weightString = "";
+  let bmiString = "";
+
+  let heightType = "";
+  let weightType = "";
+  if (heightAmount != undefined && heightMeasurementType != undefined) {
```
BobanL, line 242: you could just check [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy#:~:text=That%20is%2C%20all%20values%20are%20truthy%20except%20false%2C%200%2C%20%2D0%2C%200n%2C%20%22%22%2C%20null%2C%20undefined%2C%20NaN%2C%20and%20document.all.) here and on line 251 to ensure that empty string and null values don't pass this check. 
```js
if (heightAmount && heightMeasurementType ) {
```

containers/ecr-viewer/src/app/utils.tsx
BobanL, line 1: I think just to keep things even everywhere we should probably be using the [triple equals to enforce strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness). I don't think it would actually impact anything here, but it's probably just better practice to do that.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -226,6 +226,45 @@ const formatStartEndDateTime = (
...
+    }
+    weightString = `Weight: ${weightAmount} ${weightType}\n\n`;
+  }
+
+  if (bmi != undefined) {
+    bmiString = `Body Mass Index (BMI): ${bmi}`;
+  }
+
+  const combinedString = `${heightString} ${weightString} ${bmiString}`;
+  return combinedString.trim().length ? combinedString : "";
```
BobanL, line 265: We could just return `combinedString.trim()` here to ensure that no extra spaces are being added at the end. I believe right now as is if there wasn't BMI there would be two extra lines.

containers/ecr-viewer/src/app/view-data/components/Clinical.tsx
```diff
@@ -46,10 +47,24 @@ const ClinicalInfo = ({ activeProblemsDetails }: ClinicalProps) => {
...
+          </div>
+        </AccordianDiv>
+      </>
+    );
+  };
+
   return (
     <AccordianSection>
-      <div className="margin-top-3">
+      <div>
```
BobanL, line 65: I think we could remove this div,  I don't think it adds any value.

BobanL (commented):
Looks good, just a couple of comments!

containers/ecr-viewer/src/app/utils.tsx
KennethSkylight: changed

containers/ecr-viewer/src/app/utils.tsx
KennethSkylight: changed

containers/ecr-viewer/src/app/utils.tsx
KennethSkylight: changed

containers/ecr-viewer/src/app/view-data/components/Clinical.tsx
KennethSkylight: removed

BobanL (approved):
🎉 Thanks for making those changes 😺

Issue #1167, opened 2024-01-29
Add social history section to PHDC


DanPaseltiner:
Let's add support to `PHDCBuilder` for the Social History section by adding a `build_social_history` method.

PR #1261, opened 2024-02-05
Build social history info & separate social history from clinical info

### LINKED ISSUE #1240: Add method for building social history information section

m-goggins:
# PULL REQUEST

## Summary
This PR creates the `_build_social_history_section` method and introduces logic for parsing observations into the appropriate section of the PHDC (social history or clinical info). We keep the observations together when parsed and separate out the observations into the different sections in the builder by checking each observations `obs_type` (category). For now, the logic is such that observations with the category == `social-history` are included in the social history section and every other category (or if there is no category in the observation because it's not required by FHIR) is put into clinical information.

## Related Issue
Fixes #1240 
, #1167 

## Additional Information

## Checklist


containers/message-parser/app/utils.py
```diff
@@ -445,6 +445,17 @@ def transform_to_phdc_input_data(parsed_values: dict) -> PHDCInputData:
...
+                input_data.clinical_info = []
+                input_data.social_history_info = []
+                for obs in value:
+                    if obs["obs_type"] == "social-history":
+                        input_data.social_history_info.append(Observation(**obs))
+                    elif (
+                        "obs_type" not in obs.keys()
+                        or obs["obs_type"] != "social-history"
+                    ):
+                        input_data.clinical_info.append(Observation(**obs))
```
DanPaseltiner, line 458: The conditions for this if block seem a little suspect to me. As currently written we will always run `obs["obs_type"]` which will raise a `KeyError` if `"obs_type"` is not a key in `obs`. Then in the `elif` we check if `"obs_type"` is not in `obs`, but if we get to this point we already know that it must be in `obs` because if it wasn't an exception would have already been raised. I have two questions:

1. Since `"obs_type"` is part of `Observation` and has a default value why are we concerned about it potentially being missing?
2. Is there any reason this can't just be a simple if, else block? If we have a social history observation include in social history otherwise include in clinical information?

Also, for future reference if you want to check if a key is in a dictionary you can just use `key in dict`. `"obs_type" not in obs.keys()` is equivalent to `"obs_type" not in obs`

containers/message-parser/app/phdc/builder.py
```diff
@@ -233,28 +233,27 @@ def build_header(self):
...
             case "contact_record":
                 pass
             case "lab_report":
                 pass
             case "morbidity_report":
                 pass
-
-        self.phdc.getroot().append(body)
+        self.phdc.getroot().append(social_history_info)
+        self.phdc.getroot().append(clinical_info)
```
DanPaseltiner, line 251: I think we need to reconsider our approach calling `.append()`. This works for now because we have not implemented any other kind of PHDC than case report. However, if we create a type of PHDC that doesn't for instance have `social_history_info` then we will run into a referenced before assignment issue on line 250. I think there are two choice here:

1. Append sections as we build them, something like this: `self.phdc.getroot().append(self._build_social_history_info())`
2. Collect all components of the body in a list as we build them for a given PHDC type and then append for each afterwards.

Also, have we considered what happens if we build a section like Social History, but we have not observations to add to it? Addressing this might be beyond the scope of this PR, but it just occurred to me.

DanPaseltiner (changes requested):
Thanks for all of your work here. Left two comments below that I think we need to address before merging.

containers/message-parser/app/utils.py
m-goggins: Good catch! I forgot we set a default value and was operating under the assumption that `obs_type` may or may not be present for a given `Observation`. I will change to if/else.

containers/message-parser/app/phdc/builder.py
m-goggins: It looks like I removed the `body.append(clinical_info)` on line 243 and I'm inclined to add it back (option #1 that you provided). 

_Also, have we considered what happens if we build a section like Social History, but we have not observations to add to it? Addressing this might be beyond the scope of this PR, but it just occurred to me_

We have not addressed this yet. Right now it will add the title/code/etc. parts of the section but no observations if there aren't any. I'm wondering if it would be better (or allowed) to return 1) the title/code/etc. parts of the section but no observations, or 2) omit the section entirely.

DanPaseltiner: Going with option 1 makes  sense to me. Can we do this for both `clinical_info` and `social_history`?

> We have not addressed this yet. Right now it will add the title/code/etc. parts of the section but no observations if there aren't any. I'm wondering if it would be better (or allowed) to return 1) the title/code/etc. parts of the section but no observations, or 2) omit the section entirely.

Completely agreed on the current behavior and I have the same question. I think this depends on the needs of NBS and/or what happens with the validation that we do. For now I think we can leave things the way they are, understand how the builder currently functions and change as needed in the future.

@robertmitchellv if you have a spare moment could you try running a PHDC through the validator your have build that has a clinical information section with no observations?

DanPaseltiner (approved):
Thanks for making those changes! 🚀

PR #1262, opened 2024-02-05
Handle FHIR data in orchestration service

### LINKED ISSUE #1226: Modify the `/process` endpoint of the orchestration service to accept multiple data types

bamader:
# PULL REQUEST

## Summary
This PR expands the orchestration service to handle more data types. Specifically, it adds functionality to handle FHIR data and only selectively uses zipping and unzipping when appropriate.

Also, don't worry about the diff size! 400 lines of that is from one sample FHIR bundle that's been created for testing purposes.

## Related Issue
Fixes #1226 

## Additional Information
One note, and two areas I'd like some deliberate extra attention / focus on. 

Note: HL7v2 has deliberately been left out of this work, since bringing it in will involve invoking the FHIR converter.

Focus Areas:
1. Right now, the input to the orchestration service expects the `message` parameter to be a string. That means in a couple places in the code, before we send a FHIR bundle to the orhcestration service, we have to serialize it as a JSON string with `json.dumps` and then use `json.loads()` for some validation testing. I don't know how we feel about stringifying and decoding JSON so much, so if there's another option or people want to change the input type I'm all ears.
2. Testing. I _think_ I got testing situated enough to prove that this service can handle FHIR data (particularly the integration test that uses a new FHIR workflow), but I'd love folks to be particularly attentive to the tests to make sure I'm proving what I think I am. Just calling that out because I know sometimes tests are hard to read carefully!


containers/orchestration/app/main.py
```diff
@@ -118,28 +118,32 @@ async def process_message_endpoint_ws(
...
-
-    zip_content = unzip_http(upload_file)
+    rr_content = None
+    if upload_file.content_type == "application/zip":
+        unzipped_file = unzip_http(upload_file)
+        message = unzipped_file.get("ecr")
+        rr_content = unzipped_file.get("rr")
+    elif upload_file.content_type == "application/json":
+        message = load_json_from_binary(upload_file)
+    else:
```
emmastephenson, line 137: Do we want to accept data that's not a zip or json?

containers/orchestration/app/main.py
```diff
@@ -118,28 +118,32 @@ async def process_message_endpoint_ws(
...
-        raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail="Only .zip files are accepted at this time.",
-        )
-
-    zip_content = unzip_http(upload_file)
+    rr_content = None
+    if upload_file.content_type == "application/zip":
+        unzipped_file = unzip_http(upload_file)
+        message = unzipped_file.get("ecr")
```
emmastephenson, line 133: Way out of scope for this PR, but I'd really like to get to the point that we pass actual objects around instead of dictionaries - my eyes hurt every time we do `variable.get("field")` instead of just `variable.field` 😭

emmastephenson (commented):
All this looks good, and tests seem through too! Thanks! Just one question about desired functionality.

containers/orchestration/app/main.py
bamader: Lol. To be honest I'd never used `variable.get("")` until probably last year when another dev threw it into the tabulation service I think. It got scrapped but the convention somehow stuck around. I think normal pythonic convention is just using the bracketed accessor `variable[property]` and not bothering with `.get()`. Sometimes it can be hard to toss around objects in more functional programming, but I'm totally with you that the getter is super annoying!

containers/orchestration/app/main.py
bamader: I think our goal is to eventually broaden out the service to be able to accept HL7v2 / ELR data which would come in as just a string. And actually, it looks like that's much easier to implement than I thought because we've already got template identifying functions in the PHDI library, so I'm going to go ahead and put that in because I don't think that's enough work to add a whole other ticket.

DanPaseltiner (commented):
Thanks for putting this together. There is a lot going on here and I'm noticing a need for additional tickets. I think we should be able to talk through everything and get this merged during our pairing session shortly.

containers/orchestration/app/main.py
```diff
@@ -118,28 +118,32 @@ async def process_message_endpoint_ws(
...
     message_type: str = Form(None),
+    data_type: str = Form(None),
     config_file_name: str = Form(None),
     include_error_types: str = Form(None),
     upload_file: UploadFile = File(None),
 ) -> ProcessMessageResponse:
     """
-    Processes an uploaded zip file through a series of microservices.
+    Processes an uploaded file through a series of microservices.
     """
```
DanPaseltiner, line 128: Update doc string

containers/orchestration/app/main.py
```diff
@@ -167,7 +172,7 @@ async def process_message_endpoint(
 
 
 async def apply_workflow_to_message(
-    message_type, config_file_name, include_error_types, ecr_content, rr_content
+    message_type, data_type, config_file_name, include_error_types, message, rr_content
 ):
```
DanPaseltiner, line 176: Let's add type hints

containers/orchestration/app/models.py
```diff
@@ -42,15 +48,44 @@ class ProcessMessageRequest(BaseModel):
...
+    @root_validator()
+    def validate_fhir_message_is_dict(cls, values):
+        message = values.get("message")
+        data_type = values.get("data_type")
+        if data_type == "fhir" and type(json.loads(message)) is not dict:
+            raise ValueError(
+                "A `data_type` of FHIR requires the input message "
+                "to be a valid dictionary."
+            )
+        return values
```
DanPaseltiner, line 87: Let's add doc strings and type hints

containers/orchestration/app/services.py
```diff
@@ -79,13 +79,27 @@ def message_parser_payload(**kwargs) -> dict:
     response = kwargs["response"]
```
DanPaseltiner, line 79: Let's add a doc string

containers/orchestration/app/utils.py
```diff
@@ -84,6 +84,10 @@ def unzip_http(upload_file: UploadFile) -> Dict:
     return search_for_ecr_data(zipped_file)
 
 
+def load_json_from_binary(upload_file: UploadFile) -> Dict:
+    return json.load(io.BytesIO(upload_file.file.read()))
+
+
```
DanPaseltiner, line 90: Let's add a doc string

containers/orchestration/app/services.py
```diff
@@ -79,13 +79,27 @@ def message_parser_payload(**kwargs) -> dict:
...
-    data = {
-        "message_format": config["configurations"]["message_parser"]["message_format"],
-        "parsing_schema_name": config["configurations"]["message_parser"][
-            "parsing_schema_name"
-        ],
-        "message": r["bundle"],
-    }
+
+    # We're going to be hitting the fhir_to_phdc side of the message parser
+    if "phdc_report_type" in config["configurations"]["message_parser"]:
```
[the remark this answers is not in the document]
bamader: Done. We now check for the endpoint itself rather than a configuration param.

containers/orchestration/app/main.py
```diff
@@ -118,28 +118,32 @@ async def process_message_endpoint_ws(
...
 
-    if upload_file.content_type != "application/zip":
-        raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail="Only .zip files are accepted at this time.",
-        )
-
-    zip_content = unzip_http(upload_file)
+    rr_content = None
+    if upload_file.content_type == "application/zip":
```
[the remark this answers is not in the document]
bamader: Our model right now assumes that all zipped data right now is eCR (since why would you zip an HL7 message or a FHIR-formatted JSON file on its own, whereas eCR might come with RR), but I guess in the future we could expand to other data types being zipped. That doesn't feel super important on its own here, though, unless you think otherwise?

bamader:
HL7v2 processing functionality is added, along with a host of new unit tests, integration tests, and docstrings!

containers/orchestration/app/services.py
```diff
@@ -43,16 +55,41 @@ def validate_response(**kwargs) -> bool:
...
+    with default permissive EICR templates (needed to preserve
+    compatibility with demo UI viewers.)
+    """
     input = kwargs["input"]
+    msg = str(input["message"])
+    # Template will depend on input data formatting and typing, so try
+    # to figure that out. If we can't, use our default EICR settings
+    # to preserve backwards compatibility
+    try:
+        conversion_settings = _get_fhir_conversion_settings(msg)
```
emmastephenson, line 73: Didn't know about this `_get_fhir_conversion_settings` method, it's cool!

containers/orchestration/app/default_configs/sample-fhir-test-config.json
```diff
@@ -0,0 +1,42 @@
...
+      }
+    },
+    "message_parser": {
+      "phdc_report_type": "case_report",
+      "url": "${MESSAGE_PARSER_URL}"
+    }
+  },
+  "steps": [
+    {
+      "service": "validation",
```
emmastephenson, line 22: Do you want to include the validation service on non-eCR messages? My recollection is that the other endpoints are just stubbed, only the eCR endpoint actually does anything.

containers/orchestration/app/services.py
```diff
@@ -76,16 +113,34 @@ def ingestion_payload(**kwargs) -> dict:
...
-        "parsing_schema_name": config["configurations"]["message_parser"][
-            "parsing_schema_name"
-        ],
-        "message": r["bundle"],
-    }
+
+    # We determine which endpoint to hit by finding the message-parsing
+    # service step and checking the desired call
+    data = {"message": r["bundle"]}
+    for step in config["steps"]:
```
emmastephenson, line 130: This feels a little off to me, because the config is static so theoretically we should be able to check the endpoint value a little more directly than with a for loop. (It wouldn't be much of a performance hit since this would only be hit once during orchestration, but still.) Unfortunately, the way we've configured the dictionary the closest I got was 

`endpoint = [step["endpoint"] for step in config["steps"] if step.get("service") == "message_parser"]`

which is still not ideal.

No action needed here I think, just a note that maybe we should rethink how we're storing the config 🤔

containers/orchestration/app/services.py
```diff
@@ -24,6 +27,11 @@
 
 
 def validation_payload(**kwargs) -> dict:
+    """
```
emmastephenson, line 30: Thanks for adding all these doc strings!

emmastephenson (approved):
Looks great!

containers/orchestration/app/services.py
bamader: Oh yeah this was a lifesaver for parsing out HL7 root templates in particular! I would _not_ have been able to get this functionality ready otherwise 😅

containers/orchestration/app/default_configs/sample-fhir-test-config.json
bamader: This is a great call out. You're right that the validation service only has eCR validation implemented, but I uncovered two bugs / tech debt pieces that make it best to leave the validation step in right now, as I think they would impose some unrelated burden on this PR.

1. The `ingestion_payload` function uses an accessor pattern that prevents it from being the first step in a config workflow. On line 96 of `services.py`, it accesses the `.json()` property of the response. However, this property will only exist if another service has previously been invoked on the API input and spit back out a formatted Response object; trying to use the ingestion service as the first step in a workflow (which could be completely logical for a task involving FHIR data since it doesn't need to be converted) means that the `call_apis` function will have initialized `response` to simply be the `input` to the workflow, rather than an actual response object, because nothing else has returned yet (lines 187 - 206 in `services.py`). Hence, it won't have the `.json()` property and it'll break. The same thing goes for the `ingestion_payload` trying to access the `["response"]` and `["FhirResource"]` properties of the input object. All this means that ingestion can't go first in a workflow because of the way data is passed around and accessed.
2. The natural next step is to start with the validation service to get everything set up correctly in the response bodies for later services. While I was working on that, I noticed that in `validation_payload` of `services.py`, we hard code the `message_type` to `ecr` when returning the dictionary input. This makes sense because the validation service is only set up to actually run things on eCR data, but it still has methods stubbed out for vxu, elr, and FHIR. This means it should theoretically be able to accept a message type for one of these datas and just execute a no-op on its side of things (based on lines 87 - 115 of the validation's `app.main.py`). However, trying to use `input.get("message_type")` breaks the validation service even though methods are already implemented for those types. So we're left in a weird state where validation basically has to be the first service called in a workflow, even if it doesn't do anything for its particular data type, but we also have to hard-code eCR in as the message format, which means that something is weird and/or wrong somewhere else.

This is all fairly gnarly, but none of it breaks this PR or the ability of the orchestration service to handle FHIR data right now. For non-eCR data types, we just have an extra inefficient step that doesn't really do anything before we get to actual processing. In line with our make it work, make it right, make it fast paradigm, this PR feels like a make it work for FHIR, meaning we should open some follow work to debug the interactions I described above to make it right.

Probably a lot longer of an answer than you were looking for, but the orchestration service is turning out to be quite hairy!

containers/orchestration/app/main.py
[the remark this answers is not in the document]
DanPaseltiner: I'm thinking about preventing users from sending a zip file containing something other than eICR. We can handle this in a future ticket. My thought that was that this is an edge case we don't currently handle.

containers/orchestration/app/default_configs/sample-fhir-test-config.json
DanPaseltiner: Thanks for right up @bamader. I'm going to look closely at that that part of the code base to really understand what you are saying. I don't believe the validation service is currently set up to handle FHIR as an input format, so we'll definitely need to address this because we want to ultimately set up a pipeline that starts with FHIR and ends with PHDC.

DanPaseltiner: Right now it looks like for validation we are hard coded to always assume we have ecr data.

emmastephenson: Interesting, thank you for the detailed explanation @bamader! Sounds like we have some thinking to do and further tickets to write here, for both orchestration and validation.

PR #1295, opened 2024-02-08
Create bundled config and steps

### LINKED ISSUE #1287: Update orchestration service workflow config files based on new schema

bamader:
# PULL REQUEST

## Summary
This PR creates workflow configurations in the new format devised during the orchestration service overhaul meeting. Configuration and steps are now bundled together for easier iteration during `call_apis` and easier access of the params for each specific service.

## Related Issue
Fixes #1287 

## Additional Information
JSON dictionaries _do not_ preserve key order, and neither does Python when performing a `json.loads`. That means we can't just use the first level of the JSON dict as the workflow steps, but putting them all into a list will force an ordered serialization. It's a minimal but necessary extra step.

### DIFF
--- a/containers/orchestration/app/default_configs/sample-fhir-test-config-new.json
+++ b/containers/orchestration/app/default_configs/sample-fhir-test-config-new.json
@@ -1,0 +1,45 @@
+{
+  "workflow": [
+    {
+      "service": "validation",
+      "url": "${VALIDATION_URL}",
+      "endpoint": "/validate",
+      "params": {
+        "include_error_types": "error"
+      }
+    },
+    {
+      "service": "ingestion",
+      "url": "${INGESTION_URL}",
+      "endpoint": "/fhir/harmonization/standardization/standardize_names",
+      "params": {
+        "overwrite": true
+      }
+    },
+    {
+      "service": "ingestion",
+      "url": "${INGESTION_URL}",
+      "endpoint": "/fhir/harmonization/standardization/standardize_dob",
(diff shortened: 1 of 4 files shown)


bamader:
NOTE: The integration test failure here is not only unexpected but has nothing to do with any code uploaded here. It's because the connection the database service for `save_to_db` timed out, which is def unrelated to me making three new JSON files and adding a comment. Will not be debugging in this PR.

containers/orchestration/app/default_configs/sample-fhir-test-config-new.json
```diff
@@ -0,0 +1,45 @@
...
+      "params": {
+        "include_error_types": "error"
+      }
+    },
+    {
+      "service": "ingestion",
+      "url": "${INGESTION_URL}",
+      "endpoint": "/fhir/harmonization/standardization/standardize_names",
+      "params": {
+        "overwrite": true
```
DanPaseltiner, line 16: We shouldn't need to specific `overwrite` in our requests to ingestion. One it [defaults](https://cdcgov.github.io/phdi/latest/containers/ingestion.html#tag/fhirharmonization) to `True`. Also this is a param that they API shouldn't expose anyway. I made a mistake doing this when we built ingestion initially.

bamader: Ah gotcha, no problem I'll take it out

DanPaseltiner (approved):

PR #1299, opened 2024-02-08
Write conversion handlers

### LINKED ISSUE #1289: Write request and response handling functions for convert to fhir payloads

bamader:
# PULL REQUEST

## Summary
This PR creates the first set of handler functions for one of the services the orchestration service interacts with. We have a build input and unpack output function for the FHIR converter, kept completely separate from existing payloads to not interfere with compatibility, as well as new unit tests to cover both success and failure cases.

## Related Issue
Fixes #1289


containers/orchestration/app/handlers.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    converter_response = response.json().get("response")
+    print(converter_response)
+    if converter_response.status_code != 200:
+        raise HTTPException(
+            status_code=converter_response.status_code,
+            detail={"error": "Bad Request", "details": "FHIR conversion failed"},
+        )
+    fhir_msg = converter_response.get("FhirResource")
+    return fhir_msg
```
DanPaseltiner, line 60: I'm not sure we want to raise an exception here. I think for a none 200 status code we should log the error and return the status code from this function. Then while we are looping over all of our steps we can exit the loop and formulate a response to the client communicating the error.

containers/orchestration/app/handlers.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    # Template will depend on input data formatting and typing, so try
+    # to figure that out. If we can't, use our default EICR settings
+    # to preserve backwards compatibility
+    try:
+        conversion_settings = _get_fhir_conversion_settings(input_msg)
+        if conversion_settings["input_type"] == "hl7v2":
+            input_msg = standardize_hl7_datetimes(input_msg)
+    except KeyError:
+        conversion_settings = {"input_type": "ecr", "root_template": "EICR"}
```
DanPaseltiner, line 32: Do we need to do this? Won't the client have to tell us what kind of data they are sending given the changes made to the request body made previously?

containers/orchestration/app/handlers.py
```diff
@@ -0,0 +1,60 @@
...
+    :return: A dictionary ready to JSON-serialize as a payload to the
+      FHIR converter.
+    """
+    # Template will depend on input data formatting and typing, so try
+    # to figure that out. If we can't, use our default EICR settings
+    # to preserve backwards compatibility
+    try:
+        conversion_settings = _get_fhir_conversion_settings(input_msg)
+        if conversion_settings["input_type"] == "hl7v2":
+            input_msg = standardize_hl7_datetimes(input_msg)
```
DanPaseltiner, line 30: The FHIR converter already standardizes datetimes on HL7 inputs, so we don't need to here. You can see it [here](https://github.com/CDCgov/phdi/blob/main/containers/fhir-converter/app/service.py#L89)

containers/orchestration/app/handlers.py
```diff
@@ -0,0 +1,60 @@
...
+        conversion_settings = {"input_type": "ecr", "root_template": "EICR"}
+    return {
+        "input_data": input_msg,
+        "input_type": conversion_settings["input_type"],
+        "root_template": conversion_settings["root_template"],
+        "rr_data": rr_data,
+    }
+
+
+def unpack_fhir_converter_response(response: dict) -> dict:
```
DanPaseltiner, line 41: The response won't be a `dict` it will be a `Response` object I believe.

containers/orchestration/app/handlers.py
```diff
@@ -0,0 +1,60 @@
...
+    If the status code of the response the server sent back is OK, return
+    the parsed FHIR bundle from the response body. Otherwise, report what
+    went wrong.
+
+    :param response: The response returned by a POST request to the FHIR
+      converter.
+    :return: The FHIR bundle that the service generated.
+    """
+    converter_response = response.json().get("response")
+    print(converter_response)
```
DanPaseltiner, line 53: Do we need to print here?

containers/orchestration/app/handlers.py
DanPaseltiner: to be more clear I think we should return a tuple of `(status_code, message)`. Logging can even be handled external to this function in one consistent way for all steps while we are looping over them.

DanPaseltiner (changes requested):
Looking good. Exciting to see this starting to come together. Just a couple things I think we should tweak.

containers/orchestration/app/handlers.py
bamader: Yeah we need to do this to identify the root template correctly. You're right that with the new `input_type` parameter the user will be telling us their data stream and format, but they won't know the root template on their own, which is a required parameter for the fhir converter's input model (line 118 of `constants.py`)

containers/orchestration/app/handlers.py
bamader: Ah yep, you're totally right, good catch

containers/orchestration/app/handlers.py
bamader: No I actually think it is a dict. If you look at the `convert` function in the FHIR converter, we return a result that we build on line 97. In the body of the function that builds the returned object, we actually place the fhir converter's result in its own dictionary with the key "response". Check out line 138 in `app/service.py` in the FHIR converter. Unless the calling converter function automatically makes this a response, I think it actually is a dict, which is why in the old payloads and call_apis functions there's an extra level's worth of dictionary keys accessing responses.

containers/orchestration/app/handlers.py
bamader: LOL no, this was debugging kruft

containers/orchestration/app/handlers.py
bamader: Got it, that sounds good to me. It also simplifies error handling logic to be away from the unpacking functions.

containers/orchestration/app/handlers.py
bamader: Unless the code snippet I pointed out is just a KV pair called "response" inside an actual `response` object? I'm having a hard time seeing the distinction in the converter code 😒

DanPaseltiner: The response we have here will be whatever comes back from our POST request. Line 52 here calls the `.json()` method. This is method on a `Response` object not a `dict`.

DanPaseltiner: Checkout [how we make post requests](https://github.com/CDCgov/phdi/blob/848bd48a7f4665a52052c5b6fba3fd0fae1bd268/containers/orchestration/app/services.py#L178) in orchestration.

containers/orchestration/app/handlers.py
DanPaseltiner: What if we change the API to require that we are told explicitly what kind of Hl7v2 message we receiving? Our data types here would be eCR, ELR, VXU, FHIR? This way we have enough info to formulate our request to the FHIR converter without having to look at the incoming data.

DanPaseltiner: We used [this](https://github.com/CDCgov/phdi-azure/blob/main/serverless-functions/ReadSourceData/__init__.py#L24) dictionary in ReadSourceData in the Azure starter kit to get the appropriate template based on input data type. If we are already requiring out client to tell us they have Hl7v2 data it really isn't a big deal if we instead ask for specifically what kind of Hl7v2 message they have.

DanPaseltiner: Actually look like we are already asking the client for the message type, [here](https://github.com/CDCgov/phdi/blob/d1bda22ca0cb177c7747cba3e3bd86854dcf9ff0/containers/orchestration/app/models.py#L18).

bamader: The template mapping is more what I was talking about. We definitely already provide the message_type to the service, so that's not an issue, but I was saying that we needed to figure out the root template that accompanied that message_type since root template is required as an input to the FHIR converter. An HL7 message might, for example, use an ORU_0R1 root template, but it could also use the ADT_A01 template (at least according to some of our test data in the sdk). 

If we want to just stick a template mapper in there though that's totally fine with me (in the vein of the example you linked). We actually do exactly that in the message parser's conversion function in `utils.py` on lines 216-220. I was just trying to code for the possibility that we get HL7 messages or CCDA strings with different roots, but maybe that's just not actually something we need to worry about.

Happy to chat more on Monday about this, too. I feel like a 5 min pairing will straighten this up. But that's enough github for me on a functional friday!

bamader: Per conversation this morning, this is now a standard two-part input of the message data and the orchestration service's request input. This coupled with the template mapping dict really simplifies the function.

containers/orchestration/app/handlers.py
bamader: Signature and handling updated

bamader:
@DanPaseltiner just a heads up this ready for re-review based on our conversation yesterday

robertandremitchell (commented):
looks good to me. I'll defer to @DanPaseltiner if there's any other architectural considerations/cautions he has.

containers/orchestration/app/handlers.py
```diff
@@ -0,0 +1,64 @@
...
+    went wrong.
+
+    :param response: The response returned by a POST request to the FHIR
+      converter.
+    :return: A tuple containing the status code of the response as well as
+      the FHIR bundle that the service generated.
+    """
+    converter_response = response.json().get("response")
+    if converter_response.status_code != 200:
+        return (converter_response.status_code, "Bad Request: FHIR Conversion failed.")
```
DanPaseltiner, line 61: I think instead of returning something generic like `"Bad Request: FHIR Conversion failed."` we should return the response from the FHIR converter so we can have a better understanding of what precisely went wrong. What about something like this?

`f"Request to the FHIR converter failed, received the following response: {converter_response.text} "`

Note we use `.text` here in case the response is not JSON which would be the case for some networking issues.

bamader: Yeah happy to make that change. Looking at the FHIR converter code, it seemed like all we did was use a boolean K-V pair inside the response like `{ 'fhir_conversion_failed': true }` (in fhir-converter/app/service.py) but on second reading this likely just supplements what the FHIR converter already sends back. I'll get this change put in 👍

DanPaseltiner (approved):
🚀 Thanks for working through our approach to these handlers @bamader!

robertandremitchell:
@bamader , I looked at this yesterday since I felt 99.9999% responsible for the merge not going through. I _think_ if you merge with main, these checks should go through, at least when I tested doing so locally. But I didn't want to push that change without seeing if you had any thoughts/concerns about what could be happening here.

PR #1300, opened 2024-02-08
Bump version of FastAPI and typing expressions in all  places to satisfy dependabot

gordonfarrell:
# PULL REQUEST

## Summary
Does what it says on the tin. Please let me know if we were pegging any of these to a particular version for a particular reason. Danke

## Related Issue
Fixes #

## Additional Information

## Checklist


DanPaseltiner (approved):
Thank you for doing this!

emmastephenson:
Thanks for doing this, Gordon! Only thing I'll quickly mention is that we should check that the `/redoc` endpoints still work with this change - I remember there was a breaking change introduced to FastAPI at some point, but don't recall the version number. (Otherwise I think we'll just find out the next time we go to cut a release)

PR #1312, opened 2024-02-12
Add `assignedPerson` to `assignedAuthor`

### LINKED ISSUE #1279: Inside of the `<assignedAuthor>` tag add a child `<assignedPerson>` to wrape `<name>` in
## Action Requested
Inside of `<assignedAuthor>` there needs to be a child element called `<assignedPerson>` that is directly after `<id>` that wraps `<name>` inside of it.

Here is an example of what it should look like:

```xml
        <author>
		<time value="20160715141000"/>
		<assignedAuthor>
			<id root="2.16.840.1.113883.19.5"/>
			<assignedPerson>
				<name>
					<family>STD*MIS NBS NE</family>
				</name>
			</assignedPerson>
		</assignedAuthor>
	</author>
```

Currently, because `<name>` appears without the `<assignedPerson>` tag first, the validator throws the following error:

```
Element '{urn:hl7-org:v3}name': This element is not expected. Expected is one of 
({urn:hl7-org:v3}id, {urn:hl7-org:sdtc}identifiedBy, {urn:hl7-org:v3}code, {urn:hl7-org:v3}addr, 
{urn:hl7-org:v3}telecom, {urn:hl7-org:v3}assignedPerson, {urn:hl7-org:v3}assignedAuthoringDevice, 
{urn:hl7-org:v3}representedOrganization ).
```

This tells us that `<id>` is the tag that comes first, then the following tags are allowed in-between `<id>` and `<assignedPerson>`

## Acceptance Criteria
When the PHDC builder is able to produce XML following the above format, the XML assets have been updated, and the accompanying tests are passing based on the updated input of expected results and what the builder is producing.

## Additional Information
Can use the [validator](https://github.com/robertmitchellv/phdc-validation) I made to test things on your own as we figure out if we'll add something like this to the message parser or not; and if so, in what way.

m-goggins:
# PULL REQUEST

## Summary
This PR adds the `assignedPerson` subelement to the `assignedAuthor` element in order to pass validation.

## Related Issue
Fixes #1279 

## Additional Information

## Checklist

### DIFF
--- a/containers/message-parser/app/phdc/builder.py
+++ b/containers/message-parser/app/phdc/builder.py
@@ -645,6 +645,7 @@ class PHDCBuilder:
         id_element = ET.Element("id")
         id_element.set("root", "2.16.840.1.113883.19.5")
         assigned_author.append(id_element)
+        assigned_person = ET.SubElement(assigned_author, "assignedPerson")
 
         # family name is the example way to add either a project name or source of
         # the data being migrated
@@ -652,7 +653,8 @@ class PHDCBuilder:
         family_element = ET.SubElement(name_element, "family")
         family_element.text = family_name
 
-        assigned_author.append(name_element)
+        assigned_person.append(name_element)
+        assigned_author.append(assigned_person)
 
         author_element.append(assigned_author)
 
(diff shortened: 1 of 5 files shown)


DanPaseltiner (approved):
🚀

PR #1322, opened 2024-02-13
Update `birthTime` for validation & add direct xml comparison in test

### LINKED ISSUE #1278: Change and move `<birthTime>` element
## Action Requested
The `<birthTime>` element is both in the wrong place and not formatted correctly. So we'll need to do two things:
1. Move `<birthTime>` directly after `<administrativeGenderCode>`; and,
2. Add a `value` attribute to `<birthTime>` element
  a. change the datatype of the date to string and ensure there are neither spaces nor dashes, e.g., `"19500101"`

If we try and use the expected ISO standard the validator will throw the following error:

```bash
Element '{urn:hl7-org:v3}birthTime', attribute 'value':  
The value '1950-01-01' is not accepted by the pattern 
'[0-9]{1,8}|([0-9]{9,14}|[0-9]{14,14}\.[0-9]+)([+\-][0-9]{1,4})?'.
```

Here's an example of how it should look and where it should be situated:

```xml
        <administrativeGenderCode code="M" codeSystem="2.16.840.1.113883.12.1"
          codeSystemName="Administrative sex (HL7)" displayName="Male">
          <translation code="M" codeSystem="2.16.840.1.113883.12.1"
            codeSystemName="2.16.840.1.113883.12.1" displayName="Male"></translation>
        </administrativeGenderCode>
        <birthTime value="19500101"></birthTime>
        <maritalStatusCode code="S" codeSystem="2.16.840.1.113883.5.2"
          codeSystemName="MaritalStatus" displayName="Never Married">
          <translation code="S" codeSystem="2.16.840.1.113883.5.2"
            codeSystemName="2.16.840.1.113883.5.2" displayName="Single, never married"></translation>
        </maritalStatusCode>
```
## Acceptance Criteria
When the PHDC builder is able to produce XML following the above format, the XML assets have been updated, and the accompanying tests are passing based on the updated input of expected results and what the builder is producing.

## Additional Information
Can use the [validator](https://github.com/robertmitchellv/phdc-validation/) I made to test things on your own as we figure out if we'll add something like this to the message parser or not; and if so, in what way.

m-goggins:
# PULL REQUEST

## Summary
This PR updates the `birthTime` element so that it is in the correct place and formatted correctly with a `value` for validation. I also implemented a new function in utils.py to parse XML files to make it easier to make comparisons of XML trees in our tests (instead of stringified versions of the XMLs). I used [xmldiff](https://xmldiff.readthedocs.io/en/stable/index.html) because it had documentation that was easy to follow and would allow us to compare files, trees, or text versions, but I'm not strongly attached to it if we go another route for testing.

## Related Issue
Fixes #1278 

## Additional Information

## Checklist

### DIFF
--- a/containers/message-parser/app/phdc/builder.py
+++ b/containers/message-parser/app/phdc/builder.py
@@ -717,6 +717,17 @@ class PHDCBuilder:
             )
             patient_data.append(v)
 
+        if patient.birth_time is not None:
+            e = ET.Element(
+                "birthTime",
+                {
+                    "value": "".join(
+                        [num for num in patient.birth_time if num.isnumeric()]
+                    )
+                },
+            )
+            patient_data.append(e)
+
         if patient.race_code is not None:
             if patient.race_code in race_code_and_mapping:
                 display_name = race_code_and_mapping[patient.race_code]
@@ -750,11 +761,6 @@ class PHDCBuilder:
                     f"Ethnic group code {patient.ethnic_group_code} not "
                     "found in OMB classification."
                 )
-
-        if patient.birth_time is not None:
-            e = ET.Element("birthTime")
-            e.text = patient.birth_time
-            patient_data.append(e)
 
         return patient_data
 
--- a/containers/message-parser/app/utils.py
+++ b/containers/message-parser/app/utils.py
@@ -20,6 +20,7 @@ from app.phdc.models import PHDCInputData
 from app.phdc.models import Telecom
 from fastapi import status
 from frozendict import frozendict
+from lxml import etree as ET
 
 from phdi.cloud.azure import AzureCredentialManager
 from phdi.cloud.core import BaseCredentialManager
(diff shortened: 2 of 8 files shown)


containers/message-parser/app/utils.py
```diff
@@ -327,6 +328,24 @@ def read_file_from_assets(filename: str) -> str:
         return file.read()
 
 
+def parse_file_from_assets(filename: str) -> ET.ElementTree:
```
DanPaseltiner, line 331: Nice!

containers/message-parser/app/phdc/builder.py
```diff
@@ -717,6 +717,17 @@ def _build_patient(self, patient: Patient) -> ET.Element:
             )
             patient_data.append(v)
 
+        if patient.birth_time is not None:
+            e = ET.Element(
+                "birthTime",
+                {
+                    "value": "".join(
+                        [num for num in patient.birth_time if num.isnumeric()]
```
DanPaseltiner, line 724: This seems a little bit brittle to me, but given the constraints on dates in FHIR I think okay for now at least.

DanPaseltiner (approved):
LGTM.

containers/message-parser/app/utils.py
robertmitchellv: great addition!

containers/message-parser/tests/test_phdc.py
```diff
@@ -309,27 +311,33 @@ def test_build_author(family_name, expected_oid, expected_date, expected_author)
...
-    assert ET.tostring(xml_patient_data).decode() == expected_result
+    formatter = formatting.DiffFormatter(
+        normalize=formatting.WS_BOTH, pretty_print=True
+    )
+    diff = xmldiff.diff_trees(xml_patient_data, expected_result, formatter=formatter)
+
+    assert diff == ""
+
+
+def elements_equal(e1, e2):
```
robertmitchellv, line 329: this is great 🚀

robertmitchellv (approved):
i like the new tests for the xml elements!

PR #1324, opened 2024-02-14
Rob/1291 write request and response handling functions for message parser payloads

### LINKED ISSUE #1291: Write request and response handling functions for message parser payloads

robertandremitchell:
# PULL REQUEST

## Summary
This adds the handler functions for message_parser.

I'm not entirely sure I'm handling this right. Right now, at least, the only we could tell the difference between whether we're using PHDC or one of the other message_parser's is to compare the endpoints. Are there arguments we should instead add to the `OrchestrationRequest`? 

I couldn't think of a way without specifying the message_parser's json, but not sure.

Also also, I think I handled the json vs XML aspect of this correctly, but curious if you think this should be handled elsewhere and/or more elegantly.

## Related Issue
Fixes #1291


containers/orchestration/app/handlers.py
```diff
@@ -65,3 +65,81 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
     else:
         fhir_msg = converter_response.get("FhirResource")
         return (converter_response.status_code, fhir_msg)
+
+
+def build_message_parser_request(
```
DanPaseltiner, line 70: I think what we should do here is have a separate functions for the two endpoints on the message parser. We could have a `build_message_parse_message_request` and a `build_convert_fhir_to_phdc_request`. In `call_apis` based on the name of the step we'll be able to chose the correct handler functions.

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +65,81 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+    """
+    # Template will depend on input data formatting and typing
+    if endpoint_type == "/parse_message":
+        return {
+            "message_format": "fhir",
+            "parsing_schema_name": orchestration_request.get("parsing_schema_name"),
+            "credential_manager": "azure",
+        }
+    elif endpoint_type == "/fhir_to_phdc":
+        return {"phdc_report_type": "case_report"}
```
DanPaseltiner, line 99: We need to include the actual data that will be parsed or converted to PHDC in these requests. [For both endpoints](https://cdcgov.github.io/phdi/latest/containers/message-parser.html#operation/fhir_to_phdc_endpoint_fhir_to_phdc_post) this is done by including a `message` key in the dict.

containers/orchestration/app/handlers.py
DanPaseltiner: Additionally, every one of the request builder functions need to have the same signature. They should be of the form `(
    input_msg: str, orchestration_request: OrchestrationRequest
) -> dict`. This will let us set `message` to `input_msg`.

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +65,81 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+
+
+def unpack_message_parser_response(response: Response) -> Tuple[int, str | dict]:
```
DanPaseltiner, line 102: By having dedicated unpacking functions for each of the two relevant parse message endpoints we should be able to simplify this quite a bit.

DanPaseltiner (changes requested):
Thanks for getting this started and uncovering the messiness associated with handling requests and responses from `/parse_message` and `fhir_to_phdc` in the same functions. I think we should break this out into two functions for each endpoint (see comments below). Let me know if you'd like to discuss further.

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +66,114 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
     else:
         fhir_msg = converter_response.get("FhirResource")
         return (converter_response.status_code, fhir_msg)
+
+
+def build_message_parser_message_request(
+    input_msg: str,
+    orchestration_request: OrchestrationRequest,
+) -> dict:
```
bamader, line 74: I think the idea is to put the params for the workflow directly into the function signature. The orchestration service doesn't have an overarching params list it maintains for each service, but with our new workflow files, we do pair those with each step of the workflow, so I think we should just assume the `call_apis` function will hand off the appropriate dict to the function. Here's how I structured mine:

```
def build_valiation_request(
    input_msg: str,
    orchestration_request: OrchestrationRequest,
    workflow_params: dict | None = None,
) -> dict:
```

bamader: This prevents us from having to alter the Orchestration Service's base model before we're ready to commit those changes (we really don't want to add or take away any params from it until we can do so all at once, so we don't disrupt the streamline team)

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +66,114 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+      a string.
+    :param orchestration_request: The request the client initially sent
+      to the orchestration service. This request bundles a number of
+      parameter settings into one dictionary that each handler can
+      accept for consistency.
+    :return: A dictionary ready to JSON-serialize as a payload to the
+      message parser.
+    """
+    # Code idea for future state where we need to know report type
+    report_type = orchestration_request.get("params").get("phdc_report_type")
```
bamader, line 115: With the params moved into the signature, we can delete the comment here and just directly look up phdc report type.

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +66,114 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+    If the status code of the response the server sent back is OK, return
+    the parsed JSON message from the response body. Otherwise, report what
+    went wrong.
+
+    :param response: The response returned by a POST request to the message parser.
+    :return: A tuple containing the status code of the response as well as
+      parsed message created by the service.
+    """
+    try:
+        converter_response = response.json().get("response")
```
bamader, line 137: The reason we need a `.get("response")` accessor in the fhir converter handlers is because of the converter itself. If you look at line 138 of fhir-converter/app/service.py, you'll see the result object is actually encapsulated in another dictionary. None of our other services do this (it's because of the way the command line tool returns 🙄 ), so we don't need the extra `.get("response")` here. Just calling `response.json()` is sufficient.

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +66,114 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+        status_code = converter_response.get("status_code", response.status_code)
+        if status_code != 200:
+            error_message = converter_response.get("text", "Unknown error")
```
bamader, line 140: I think the text of the message will be in the body of the response object proper, rather than the JSON serialization of it, so this should be `error_message = response.text` (which python's `Requests` module assures us will always exist so no need to use get)

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +66,114 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+            return (
+                status_code,
+                f"Message Parser request failed: {error_message}",
+            )
+        else:
+            parsed_message = converter_response.get("FhirResource")
```
bamader, line 146: Does the message parser output actually have a FhirResource key?

containers/orchestration/app/handlers.py
```diff
@@ -65,3 +66,114 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+      parsed message created by the service.
+    """
+    # XML-formatted messages like PHDC
+    try:
+        if response.status_code != 200:
+            # Try to extract error message from JSON, fallback to plain text
+            try:
+                error_message = (
+                    response.json().get("response", {}).get("text", "Unknown error")
+                )
```
bamader, line 170: Same caveats as above, I don't think we need the extra `.get("response")` because of the object structure, and `.text` will be directly on the response object

containers/orchestration/app/models.py
```diff
@@ -44,6 +44,12 @@ class OrchestrationRequest(BaseModel):
             + " Valid types are fatal, errors, warnings, information"
         )
     )
+
+    params: Optional[dict] = Field(
+        description="A dictionary of the endpoint-specific parameters (if necessary).",
+        default=None,
+    )
```
bamader, line 51: We want this to be in the function signature rather than the request, since the params will change with each step of the workflow

bamader (changes requested):
Looking good so far, let's just move that params signature into the handlers rather than the model input, and then double check the "response" accessors once we invoke `.json()` on the response

containers/orchestration/app/handlers.py
robertandremitchell: I'm not sure I follow. without the `.get("response")` I get this:
```json
{
  "response": {
    "status_code": 200,
    "FhirResource": {
      "resourceType": "Bundle",
      "identifier": {
        "value": "a very contrived FHIR bundle"
      },
      "entry": [
        {
          "resource": {
            "resourceType": "Organization",
            "id": "some-org-we-dont-care-about"
          }
        },
        {
          "resource": {
            "resourceType": "Patient",
            "id": "some-uuid",
            "identifier": [
              {
                "value": "123456",
                "type": {
                  "coding": [
                    {
                      "code": "MR",
                      "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
                      "display": "Medical record number"
                    }
                  ]
                },
                "system": "urn...no idea"
              }
            ],
            "name": [
              {
                "family": "doe",
                "given": [
                  "John ",
                  " Danger "
                ],
                "use": "official"
              }
            ],
            "birthDate": "1983-02-01",
            "gender": "female",
            "address": [
              {
                "line": [
                  "123 Fake St",
                  "Unit #F"
                ],
                "BuildingNumber": "123",
                "city": "Faketon",
                "state": "NY",
                "postalCode": "10001-0001",
                "country": "USA",
                "use": "home"
              }
            ],
            "telecom": [
              {
                "use": "home",
                "system": "phone",
                "value": "123-456-7890"
              },
              {
                "value": "johndanger@doe.net",
                "system": "email"
              }
            ]
          },
          "request": {
            "method": "GET",
            "url": "testing for entry with no resource"
          }
        }
      ]
    }
  }
}
```

so we would still need to access "response", unless I'm thinking of it wrongly.

containers/orchestration/app/handlers.py
robertandremitchell: erg, no. I got mixed up here that we need to be looking at a sample message_parser (which has "message" and "parsed_values" as keys. Will update this test.

containers/orchestration/app/handlers.py
robertandremitchell: nvm, I see what you're saying. I've gotten the FHIR and message parser mixed up 🤦🏽

robertandremitchell: updating!

containers/orchestration/app/handlers.py
robertandremitchell: moved

containers/orchestration/app/handlers.py
robertandremitchell: done

containers/orchestration/app/handlers.py
robertandremitchell: updated

containers/orchestration/app/handlers.py
robertandremitchell: updated

containers/orchestration/app/handlers.py
robertandremitchell: fixed

containers/orchestration/app/models.py
robertandremitchell: removed

robertandremitchell:
updated error handling for unpacking after consulting with @DanPaseltiner , for context, @bamader. Curious your thoughts on re-review.

bamader (approved):
LGTM! Thanks for getting all those fixes in! I think the error logic being moved outside of the handler functions is part of the pattern we want. We want call_apis to be responsible for handling those as much as possible, as part of our strategy of percolating the error as high up the call chain as it'll go.

DanPaseltiner (approved):
🚀

PR #1326, opened 2024-02-14
Create validation handlers

### LINKED ISSUE #1288: Write request and response handling functions for validation payloads

bamader:
# PULL REQUEST

## Summary
This PR adds in the handlers for the validation service.

## Related Issue
Fixes #1288 

## Additional Information
The validation service is the one endpoint we have that doesn't send message data back to us. Since our new standardized response signature is going to return a tuple of status code and revised data, I left a TODO about how we should handle this special case in the `call_apis` rewrite. Basically, I think we just do validation before everything else if it's part of the user's workflow, and then if the data is invalid, we short circuit the whole loop. It lets us use a different function signature for the validation output and still preserves the logic of the rest.


containers/orchestration/tests/test_handlers.py
```diff
@@ -83,3 +110,58 @@ def test_unpack_fhir_converter_response():
...
     assert result == (200, sample_bundle)
+
+
+def test_unpack_validation_response():
+    # Case when the service fails before validating
+    # The Mock here represents a Response object generated by FastAPI, which
+    # we can't auto-generate because we don't actually make a post request
+    # when using a handler, and we can't manually give a real Response object
+    # the `.json()` attribute
+    validater_result = Mock()
```
robertandremitchell, line 121: incredibly nitpicky but should be `validator`
I won a spelling bee once

bamader: Ah crap, legit I had validator but I changed it because the `-or` made me think too much of Skeletor 😞

containers/orchestration/app/handlers.py
```diff
@@ -60,8 +81,54 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+    # happened. So, the move in call_apis will be to check if the first step
+    # is validation, and if yes, make a separate validate request. If the
+    # response comes back 200 with no errors, then we move into the rest of
+    # the loop, but if it doesn't (and the workflow inlcuded validation),
+    # then we abort and report the data invalidity to the user.
+    else:
+        if validator_response.get("message_valid") is True:
+            return (
+                validator_response.status_code,
+                validator_response.get("message_valid"),
```
robertandremitchell, line 128: is this where we would also return any non-fatal errors? I'm not sure if we currently do or if that's a consideration we ought to have.

containers/orchestration/app/handlers.py
```diff
@@ -60,8 +81,54 @@ def unpack_fhir_converter_response(response: Response) -> Tuple[int, str | dict]
...
+    # at the top of the function, before getting into the loop, that does
+    # validation service requests and response handling. This is because
+    # the validation service is the only endpoint that _does not_ also return
+    # message data to the caller. The response object contains only whether
+    # the message is valid and what kinds of errors may or may not have
+    # happened. So, the move in call_apis will be to check if the first step
+    # is validation, and if yes, make a separate validate request. If the
+    # response comes back 200 with no errors, then we move into the rest of
+    # the loop, but if it doesn't (and the workflow inlcuded validation),
+    # then we abort and report the data invalidity to the user.
```
DanPaseltiner, line 123: I understand where you are coming from here. I'd like to us to consider how we could handle this complexity inside this function. Part of the goal in the pattern we are using here with the request and response handlers is to keep all service specific logic encapsulated in these functions so that `call_apis` does not need to "know" about these details.

To achieve this we may need to modify our responses from the unpack functions. What if we returned a tuple something  something like this:

`(status_code: int, message: str/dict, continue: bool)`

Having both `status_code` and `continue` would allow us to have a situation where a service returns a `200` but we should not move onto the next step. This is one of the tricky situations we run into with validation. We could also expand our use of message. In the case where we receive a  `200` and `continue is True` then message should be the actual data. In the case where we receive a `200` and we should not continue `message` would contain whatever info was received from the service we called that would explain why we should not continue. Finally we could have a situation where `status_code == 200 and continue is True and message is None`. In this case `call_apis` could know to move onto the next step with the same input data that was used for the current step.

Curious to here what you think. I'm just coming up with this on the fly.

robertandremitchell (commented):
looks good to me. One v minor question/consideration.

DanPaseltiner (changes requested):
Thanks for putting this together and identifying the challenge the validation service not returning the actual data @bamader! I wrote up a possible path forward that I'd like to discuss further before merging. We are not currently under a time crunch and part of our focus here is to address some of the tech debt in the part of the code base so let's take the time required to figure out a solution that should work for us long term.

robertandremitchell (approved):
makes sense to me. I like the change to the Class, a lot more elegant.

containers/orchestration/app/handlers.py
```diff
@@ -12,24 +10,95 @@
...
+
+    @property
+    def status_code(self) -> int:
+        """
+        The status code the service returned to the caller.
+        """
+        return self._status_code
+
+    @status_code.setter
+    def status_code(self, code) -> None:
```
DanPaseltiner, line 42: I think we should have a type hint on `code` and add the param to the doc string.

```suggestion
    def status_code(self, code: int) -> None:
```

containers/orchestration/app/handlers.py
```diff
@@ -12,24 +10,95 @@
...
+        The data the service sent back to the caller. If this is a
+        transformation or standardization service, this value holds
+        the altered and updated data the service computed. If this
+        is the validation or a verification service, this value
+        holds any errors the service flagged in its input.
+        """
+        return self._msg_content
+
+    @msg_content.setter
+    def msg_content(self, content) -> None:
```
DanPaseltiner, line 60: I think we should have a type hint on `content` and add the param to the doc string.

```suggestion
    def msg_content(self, content: str | dict) -> None:
```

containers/orchestration/app/handlers.py
```diff
@@ -12,24 +10,95 @@
...
+    def should_continue(self) -> bool:
+        """
+        Whether the calling function can hand the processed data off
+        to the next service, based on the current service's status
+        and message content validity.
+        """
+        return self._should_continue
+
+    @should_continue.setter
+    def should_continue(self, cont) -> None:
```
DanPaseltiner, line 76: I think we should have a type hint on `cont` and add the param to the doc string.

```suggestion
    def should_continue(self, cont: bool) -> None:
```

DanPaseltiner (approved):
This looks really great @bamader! Once again thanks for working through the design with me. I'm approving so that I'm not blocking, but before merging there are a couple comments below about documentation that should be addressed.

PR #1350, opened 2024-02-21
Fix issue where save to db wouldn't work in ECR viewer

### LINKED ISSUE #1271: Bug fix: writing to ecr-viewer database from orchestration
## Action Requested
Ensure eCRs passed to the orchestration service can be rendered in the ecr-viewer.

## Acceptance Criteria
Upon upload of an eCR ZIP file to the `/process` endpoint, a database record is written.

## Testing
This functionality should be added to integration tests for the orchestration service.

## Background & Context
There seems to be a bug in the service, such that UUIDs can't be extracted from the eCR and used as the database key.

nickbristow:
# PULL REQUEST

## Summary
There was an issue where save to db would just save the response of the previous service causing issues with the ECR_VIEWER. This change makes it so the bundle is saved instead of the response

## Related Issue
Fixes #1271

## Additional Information

## Checklist


containers/orchestration/app/services.py
```diff
@@ -9,6 +9,7 @@
 from fastapi import Response
 from fastapi import WebSocket
 from fastapi.encoders import jsonable_encoder
+from icecream import ic
```
emmastephenson, line 12: We can probably remove the icecream logs now

emmastephenson (commented):
Changes largely LGTM, but we should also run this by the VIPER team - @bamader in particular has been working on recent changes to the orchestration service to make it more iteroperable with different configurations, so would be good to get his eyes on it.

Brandon for your SA this is a temporary change - we're still working to move the `save_to_db` functionality to the eCR Viewer, but in the interim the bundle wasn't saving correctly.

containers/orchestration/app/services.py
nickbristow: i'll miss it

bamader:
I _think_ the ideas in this PR won't mess up anything I'm doing, but I did just open a PR to completely overwrite call_apis to use new and much better service handlers, so we're going to run into merge conflicts between these two. The tests and data themselves weren't actually doing anything at all so those changes shouldn't matter to much, but yeah I think this is just going to cause some conflicts in call apis. Nothing too hard to sift out though.

containers/orchestration/app/default_configs/sample-orchestration-config-new.json
emmastephenson, line 1: Is deleting this file intentional? I thought this was the one Brandon and the rest of VIPER are using.

containers/orchestration/app/services.py
emmastephenson: Please? :) Looks like it's still in the file

containers/orchestration/app/utils.py
```diff
@@ -12,6 +12,7 @@
 from fastapi import UploadFile
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
+from icecream import ic
```
emmastephenson, line 15: Another ic to remove, please

containers/orchestration/app/default_configs/sample-orchestration-config-new.json
emmastephenson: Looked through the rest of the PR, saw that this has been replaced with the old file name - is the goal to just go back to a single unified config?

lina-roth: I believe that we noticed that the old configuration wasn't making it through and when we used the new configuration it worked. And while structurally different, the content was the same, so I think we just decided to use that configuration everywhere. @nickbristow care to comment?

containers/orchestration/app/utils.py
lina-roth: Fixed!

emmastephenson (approved):

PR #1352, opened 2024-02-21
Adding s3 route and environment variables

### LINKED ISSUE #1269: Update ecr viewer to work with S3 instead of postgres

KennethSkylight:
# PULL REQUEST

## Summary
Adding S3 route to enable connections to AWS. CDC Github doesn't allow AWS access IDs or secret keys to be added so you will have to ask me for those values.

Things to note:

- The s3 route will not work unless you add the access IDs and secret keys
- By default, it will still go through postgres. You can change it by giving it "s3" in the NEXT_PUBLIC_SOURCE environment variable

Fixes #1269

### DIFF
--- a/containers/ecr-viewer/src/app/api/s3/route.ts
+++ b/containers/ecr-viewer/src/app/api/s3/route.ts
@@ -1,0 +1,32 @@
+import { NextRequest, NextResponse } from "next/server";
+import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
+import { loadYamlConfig, streamToJson } from "../utils";
+
+// Initialize the S3 client outside of your API route handler to reuse the connection
+const s3Client = new S3Client({ region: process.env.AWS_REGION });
+
+export async function GET(request: NextRequest) {
+  const params = request.nextUrl.searchParams;
+  const ecr_id = params.get("id") ? params.get("id") : null;
+
+  const bucketName = process.env.ECR_BUCKET_NAME;
+  const objectKey = `${ecr_id}.json`; // This could also come from the request, e.g., req.query.key
+
+  try {
+    const command = new GetObjectCommand({
+      Bucket: bucketName,
+      Key: objectKey,
+    });
+
+    const { Body } = await s3Client.send(command);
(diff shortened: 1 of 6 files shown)


containers/ecr-viewer/src/app/api/s3/route.ts
```diff
@@ -0,0 +1,32 @@
+import { NextRequest, NextResponse } from "next/server";
+import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
+import { loadYamlConfig, streamToJson } from "../utils";
+
+// Initialize the S3 client outside of your API route handler to reuse the connection
+const s3Client = new S3Client({ region: process.env.AWS_REGION });
+
+export async function GET(request: NextRequest) {
+  const params = request.nextUrl.searchParams;
+  const ecr_id = params.get("id") ? params.get("id") : null;
```
BobanL, line 10: nit: I think we could leave simply as `params.get("id")`

BobanL (approved):
Looks good, left one nit. If possible, could we add some unit tests for the `api/s3` where we mock the s3 import?

containers/ecr-viewer/src/app/view-data/page.tsx
```diff
@@ -14,13 +14,26 @@ import { clinicalInfoConfig } from "./components/ClinicalInfo";
 import { PathMappings } from "../utils";
 import SideNav, { SectionConfig } from "./components/SideNav";
 import { processSnomedCode } from "./service";
+// string constants to match with possible .env values
+const S3_SOURCE = "s3";
+const POSTGRES_SOURCE = "postgres";
+
+const assignApiPath = () => {
+  console.log(process.env.NEXT_PUBLIC_SOURCE);
```
emmastephenson, line 22: We might want to remove this log statement

emmastephenson (commented):
LGTM! But a +1 to @BobanL's request for testing :) I think @gordonfarrell and @angelathe are currently working on unit tests for their new viewer route, so you may want to sync with them about how to unit test these API routes.

PR #1361, opened 2024-02-22
Use handlers in new call apis

### LINKED ISSUE #1293: Update call_apis function to be cleaner and use new payloads

bamader:
## Summary
This PR rewrites the `call_apis` function to take advantage of our new handler methods. With increasing abstraction, we're able to use our input packing and output unpacking functions to completely eliminate the need to invoke `globals()` while also increasing service modularity and error reporting robustness. This changeset also refactors a couple utils to make the code more readable, and massively updates the unit test for a call apis success to use actual responses for each invoked service in a workflow.

## Related Issue
Fixes #1293 

## Additional Information
We'll have to leave the special case of `save_to_db` in the loop until the eCR team implements those endpoints server-side. But, we'll be able to do a good amount of refactoring as part of kruft elimination once this lands.

### DIFF
--- a/containers/orchestration/app/services.py
+++ b/containers/orchestration/app/services.py
@@ -284,6 +284,7 @@ async def call_apis(
     """
     workflow = config.get("workflow", [])
     current_message = input.get("message")
+    response = current_message
     responses = {}
 
     # For websocket json dumps
@@ -309,7 +310,7 @@ async def call_apis(
                 service_response = response_func(response)
 
                 if websocket:
-                    progress_dict = _send_websocket_dump(
+                    progress_dict = await _send_websocket_dump(
                         endpoint_name,
                         response,
                         service_response,
@@ -330,7 +331,11 @@ async def call_apis(
                         + f"{service_response.msg_content}",
                     )
 
-                current_message = service_response.msg_content
+                # Validation reports only whether we should continue, with
+                # no updated data, so don't send error results to next
+                # service
+                if service != "validation":
+                    current_message = service_response.msg_content
                 responses[service] = response
 
             else:
@@ -338,7 +343,7 @@ async def call_apis(
                 response = save_to_db(url=service_url, payload=db_request)
 
                 if websocket:
-                    progress_dict = _send_websocket_dump(
+                    progress_dict = await _send_websocket_dump(
                         endpoint_name,
                         response,
                         service_response,
@@ -348,11 +353,10 @@ async def call_apis(
 
                 responses[service] = response
 
-            return (response, responses)
-
         except KeyError:
             raise HTTPException(
                 status_code=422,
                 detail="The Building Block you are attempting to call does not exist:"
                 + f" {service}",
             )
+    return (response, responses)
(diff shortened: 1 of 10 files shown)


bamader:
Requesting some reviews from eCR viewer folks to make sure we get this right. Also, the integration test that's failing on github passes without issue for me locally so I'm inclined to ignore that failure.

nickbristow (approved):

robertandremitchell (commented):
getting 5 tests failing locally 

```bash
tests/test_configs_endpoint.py ....                                         [  7%]
tests/test_handlers.py ................                                     [ 38%]
tests/test_process_message_endpoint.py ..........                           [ 57%]
tests/test_services.py ........                                             [ 73%]
tests/test_utils.py ......                                                  [ 84%]
tests/integration/test_db.py .                                              [ 86%]
tests/integration/test_orchestration.py .FFFFF.
```

Doing some digging now.

robertandremitchell:
> getting 5 tests failing locally
> 
> ```shell
> tests/test_configs_endpoint.py ....                                         [  7%]
> tests/test_handlers.py ................                                     [ 38%]
> tests/test_process_message_endpoint.py ..........                           [ 57%]
> tests/test_services.py ........                                             [ 73%]
> tests/test_utils.py ......                                                  [ 84%]
> tests/integration/test_db.py .                                              [ 86%]
> tests/integration/test_orchestration.py .FFFFF.
> ```
> 
> Doing some digging now.

rebuilt the images and now just getting the same error that GitHub Actions is flagging.

containers/orchestration/app/services.py
```diff
@@ -179,54 +229,134 @@ def post_request(url, payload):
...
+                        websocket,
+                    )
 
-            if validate_response(response=response) is False:
-                break
+                responses[service] = response
 
-            responses[endpoint] = response
-        else:
+        except KeyError:
```
DanPaseltiner, line 356: I think we can removed this `try/except` block. First, large blocks like this can be dangerous because they make it hard to know exactly where the exception we are attempting to handle may be raised. Unless I'm mistaken we have a possibility of a `KeyError` on lines 306 and 307. However since we do not currently support users uploading their own configs we can be confident that the configs this team produces will work. Additionally, when we do develop support for user defined configs we'll need a robust validation process and we will only accept configs from them that we know will run instead of waiting to find out at run time. So if the concern here is just that we might have bade configs I think this is reasonable concern but we can address it in a more robust way in the future. If we have this `try/except` for some other reason i apologize I would like to better understand.

DanPaseltiner (commented):
Looks great @bamader, thank you for pulling all of this together! Left one comment on the `try/except` I think is worth discussing.

robertandremitchell (approved):
test is working locally for me, as well. thanks for pairing on this!

containers/orchestration/app/services.py
bamader: Nope it's just for bad configs, but everything you're saying about handling it more robustly in later work makes sense to me. I have no problem scrapping it here, it's kind of complicating the code block anyway.

Issue #1068, opened 2024-01-11
Include PHDC XML validation in at least one unit test of the PHDC builder


robertmitchellv:
When the PHDC builder is able to produce PHDC XML add at least one unit test of XML validation of the output constructed by the PHDC builder.

robertmitchellv:
There are going to be a three new validation tickets to knock out before this can move forward; going to add them the sprint backlog.

PR #1365, opened 2024-02-23
Add an XML validation test based on HL7v3 and SDTC extensions to PHDC tests

robertmitchellv:
# PULL REQUEST

## Summary
We are adding a validation test as its own `test_phdc_validation.py` file and `pip` dependencies related to this test are separated into `dev-requirements.txt`. We're adding conditional logic to our `test.yaml` to look for this file for testing but when building containers via `docker-compose` this tests dependencies and `schema` assets will not be included.

## Related Issue
Fixes:
- #1068 

## Additional Information
I added the required files and file structure of the `xsd` files and pruned as much out of it as possible. This was necessary in some ways because of our pre-commit hooks for prettier and the older `css` assets causing problems. After removed some of the unnecessary files I tested that validation still works and it does. I'll figure out the best way to add this to the container.


containers/message-parser/app/phdc/builder.py
```diff
@@ -50,6 +50,16 @@ def to_xml_string(self) -> bytes:
...
 
+    def to_element_tree(self) -> ET.ElementTree:
+        """
+        Return the PHDC XML document as an ElementTree.
+
+        :return: The PHDC XML document as an ElementTree.
+        """
+        if self.data is None:
+            raise ValueError("The PHDC object must be initialized.")
+        return self.data
```
DanPaseltiner, line 61: Nice!

containers/message-parser/requirements.txt
```diff
@@ -7,4 +7,4 @@ pytest
 frozendict
 testcontainers[compose]==3.7.1
 xmldiff
-     
\ No newline at end of file
+rich
```
DanPaseltiner, line 10: Based on what I can see we have added `rich` purely for testing purposes. This is completely fine, but as things stand now we will include `rich` in the images for this service which would just be bloat. Could we add a `dev_requirements.txt` file in `container/message-parser/` and move any dependencies there that will not be required when the service is running in production? This way on our locals we can install everything, but we keep the images we ship as light weight as possible?

containers/message-parser/tests/test_phdc_validation.py
```diff
@@ -0,0 +1,283 @@
...
+        / "schema"
+        / "extensions"
+        / "SDTC"
+        / "infrastructure"
+        / "cda"
+        / "CDA_SDTC.xsd"
+    )
+
+    # print the resolved xsd schema path to confirm it's correct
+    console.print(f"Loading XSD schema from: {xsd_path}", style="bold yellow")
```
DanPaseltiner, line 41: Is this needed anymore or is this a dev artifact?

containers/message-parser/tests/test_phdc_validation.py
```diff
@@ -0,0 +1,283 @@
...
+    try:
+        # attempt to load and parse the xsd file
+        with open(xsd_path, "rb") as xsd_file:
+            xsd_tree = ET.XMLSchema(ET.parse(xsd_file))
+            # print a confirmation message that the schema is loaded
+            console.print("XSD schema loaded successfully", style="bold green")
+    except Exception as e:
+        # print an error message if the schema fails to load
+        console.print(f"Failed to load XSD schema: {e}", style="bold red")
+        return False
```
DanPaseltiner, line 52: Generally bare exceptions are not ideal. Could we use the specific exception that might be raised? I also wonder if we even need this `try/except` at all. We are using a hardcoded path we know is correct to a file we placed. Is there ever a real chance we could have a failure here?

containers/message-parser/tests/test_phdc_validation.py
```diff
@@ -0,0 +1,283 @@
...
+        table.add_column("Column", style="dim", width=6, justify="right")
+        table.add_column("Message", overflow="fold")
+        for error in xsd_tree.error_log:
+            table.add_row(
+                str(error.line),
+                str(error.column),
+                error.message,
+            )
+        console.print(table)
+        return False
```
DanPaseltiner, line 82: Since we `return` in the if block we can eliminate the `else` block.

```suggestion

      console.print("the XML file is not valid", style="bold red")
      # create the table for the error log
      table = Table(
          title="PHDC Validation", show_header=True, header_style="bold magenta"
      )

      # create the table columns to display the errors
      table.add_column("Line", style="dim", width=6, justify="right")
      table.add_column("Column", style="dim", width=6, justify="right")
      table.add_column("Message", overflow="fold")
      for error in xsd_tree.error_log:
          table.add_row(
              str(error.line),
              str(error.column),
              error.message,
          )
      console.print(table)
      return False
```

containers/message-parser/tests/test_phdc_validation.py
```diff
@@ -0,0 +1,283 @@
...
+                    display_name="Information As of Date",
+                ),
+                value=CodedElement(
+                    xsi_type="TS",
+                    value="20240124",
+                ),
+            ),
+        ],
+    )
+
```
DanPaseltiner, line 273: I think we need to load the PHDC file we use to test the entire process instead of hand rolling some data here. As things change and we implement additional parts of the PHDC we don't want to have to keep maintaining this data.

DanPaseltiner (changes requested):
This is really awesome @robertmitchellv! There are few, but not major, things I think we should tighten up a bit before merging. 

> Additionally, the validation function is in the test script; once I have feedback from others about its construction and how we're doing this I can potentially move it to utils.py in phdc

I think the current structure makes a lot of sense because it does not result in any of this work being included in the message-parser image when we build. If we move the function to `utils.py` it would be included there which would add code the container image that would never run because our tests run outside of the container.

> I added the required files and file structure of the xsd files and pruned as much out of it as possible. This was necessary in some ways because of our pre-commit hooks for prettier and the older css assets causing problems. After removed some of the unnecessary files I tested that validation still works and it does. I'll figure out the best way to add this to the container.

We should NOT add these files to the container as they are not required for the container to run, just for validation of data that the container outputs. The current location of `schemas/` is perfect.

containers/message-parser/requirements.txt
robertmitchellv: @DanPaseltiner that's a great idea; i'll get started on this

containers/message-parser/tests/test_phdc_validation.py
robertmitchellv: dev artifact; will remove

containers/message-parser/tests/test_phdc_validation.py
robertmitchellv: this seemed more necessary when i was using `fzf` as the way to supply paths to the function in the other repo of validation work i was tinker with. i can rework this bit here; there shouldn't be any reason to not know where the paths are especially since they need to maintain their structure since the `xsd`s are all connected and referenced. i'll likely remove the `try`/`except`.

containers/message-parser/tests/test_phdc_validation.py
robertmitchellv: if we load the file then there's no way to test that our builder can produce valid PHDC. this is testing at the point of our dataclasses that from that point on, **if** supplied with the correct data, that what gets built is valid PHDC so if they start failing locally when extending a piece of the builder you know quickly something isn't right.

but if we want to separate that piece for integration and just have a dev norm that the file used in this test will always be a recently converted response from a fhir bundle we've created for PHDC conversion we can do that but it will still need to be maintained as we extend `PHDCBuilder` just like the rest of our test assets.

or did you want to add one of the sample PHDCs as a test here?

robertmitchellv:
@nickclyde i added you to check this piece of the `test.yaml`; this seems to work in this branch, which opens up the possibility to start removing testing only python packages from our container images

https://github.com/CDCgov/phdi/blob/427463fa2713594b1afcfb838d9d9641a230e4eb/.github/workflows/test.yaml#L150

containers/message-parser/tests/test_phdc_validation.py
DanPaseltiner: Oh right that's a good point. Let's leave this as is for now. Eventually though I want to unify all of these different test fixtures we are creating to prevent some from getting stale. My concern is that if we implement a new PHDC section and don't add that data to this fixture then this test is actually checking if that new work produces valid PHDC content. In a separate ticket we should add an integration test that starts with a FHIR bundle and checks the resulting PHDC for validity. Ideally this is the one FHIR bundle we maintain for all end to end testing.

containers/message-parser/requirements.txt
```diff
@@ -7,4 +7,3 @@ pytest
 frozendict
 testcontainers[compose]==3.7.1
 xmldiff
```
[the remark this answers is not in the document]
robertmitchellv: i think this is a great idea! it's a small change and i'm happy to add. the `test.yaml` opens this up as an option for all our python containers now, which is great. i'll probably add a ticket for that in the backlog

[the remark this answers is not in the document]
robertmitchellv: @DanPaseltiner so `frozendict` it turns out **is** needed by the container; it won't spin up without it. but i moved the others; we're good to go now

PR #1366, opened 2024-02-23
set ecr viewer's auth parameter as cookie

### LINKED ISSUE #1354: eCR viewer backend auth api
## Action Requested
As an engineer, I want the user's auth token to be stored as as a cookie so that the client can use it as needed.

## Acceptance Criteria
 - [ ] **Scenario: Create new backend endpoint**
  - Given a new backend endpoint is created that accepts a POST request and an authorization header
  - When the endpoint is called with a valid authorization header
  - Then the endpoint should return a 200 OK response and set an HTTP-only cookie with the name "authorization" with the provided value

## Background & Context
The auth token will be sent originally to the frontend via query parameter like `/view-data?id=1234&auth=efghijk`, in order to secure that authorization we should move that auth to a cookie to be used by the frontend client. 

## Additional Information
A rough example of what this [could look like](https://github.com/CDCgov/phdi/compare/boban/spike-auth#diff-f3dfc6e68cf8958cbe02d5848dd40959d11a66f278aaac35b4a4542d24b643bd).
[NextJs docs](https://nextjs.org/docs/app/api-reference/functions/cookies#cookiessetname-)

BobanL:
# PULL REQUEST

## Summary
Create a middleware that will create an `auth-token` cookie and strip the `auth` query parameter from url.

## Related Issue
Fixes #1354 
Fixes #1355 

## Additional Information

https://github.com/CDCgov/phdi/assets/10108172/dc4d2e12-4605-4f43-8abe-ed80d36aa14d

### DIFF
--- a/containers/ecr-viewer/src/middleware.test.ts
+++ b/containers/ecr-viewer/src/middleware.test.ts
@@ -1,0 +1,23 @@
+/**
+ * @jest-environment node
+ */
+import { middleware } from "@/middleware";
+import { NextRequest, NextResponse } from "next/server";
+import { NextURL } from "next/dist/server/web/next-url";
+
+describe('Middleware', () => {
+  const redirectSpy = jest.spyOn(NextResponse, 'redirect');
+
+  afterEach(() => {
+    redirectSpy.mockReset();
+  });
+  it('should strip the auth query param and set the token', async () => {
+
+    const req = new NextRequest("https://www.example.com/view-data?id=1234&auth=abcd");
+
+    const resp = middleware(req);
+    expect(resp.cookies.get("auth-token")).toEqual({name: "auth-token", path: "/", value: "abcd", httpOnly: true});
+    expect(redirectSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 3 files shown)


containers/ecr-viewer/src/middleware.test.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * @jest-environment node
+ */
```
BobanL, line 3: This is required to load `NextRequest` and `NextResponse`

emmastephenson (approved):
LGTM! A question about the middleware - does Next know automatically to inject it on all incoming requests? I assume so, I'm just curious how exactly it's being called.

BobanL:
> LGTM! A question about the middleware - does Next know automatically to inject it on all incoming requests? I assume so, I'm just curious how exactly it's being called.

Yes thats right! NextJs runs the [middleware before every request](https://nextjs.org/docs/app/api-reference/next-config-js/redirects). [Here is a good resource if you want to read more](https://blog.logrocket.com/using-next-js-middleware-edge-functions/#:~:text=security%2C%20and%20functionality.-,How%20does%20Next.js%20middleware%20work%3F,-Middleware%20functions%20are)

PR #1370, opened 2024-02-23
1356 auth middleware

nickbristow:
# PULL REQUEST

## Summary
Adds middleware to ecr viewer that stops api's from loading in production

## Related Issue
Fixes #

## Additional Information

## Checklist


containers/ecr-viewer/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { NextRequest } from "next/server";
+import { importSPKI, jwtVerify } from "jose";
+export async function middleware(request: NextRequest) {
+  const auth = request.cookies.get("authorization")?.value;
+  const pathname = request.nextUrl.pathname;
+  const isDevelopment = process.env.NODE_ENV === "development" || false;
+  if (
+    !isDevelopment &&
+    pathname.startsWith("/api") &&
+    !pathname.startsWith("/api/auth")
```
BobanL, line 11: You can remove this line (11), since there is no path `/api/auth` anymore

containers/ecr-viewer/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+  ) {
+    if (!auth) {
+      return NextResponse.redirect(new URL("/404", request.url));
```
BobanL, line 14: It would probably be better to just return a response than a redirect. This would be similar to line 19. 

https://nextjs.org/docs/app/building-your-application/routing/middleware#producing-a-response

containers/ecr-viewer/src/app/view-data/page.tsx
```diff
@@ -40,14 +40,29 @@ const ECRViewerPage = () => {
...
+        await fetch("http://localhost:3000/api/auth", {
+          method: "POST",
+          headers: { authorization },
+        });
+        push(`http://localhost:3000/view-data?id=${fhirId}`);
+      };
+      handleAuth();
+    }
+  }, []);
+
```
BobanL, line 57: You won't need this once #1366 is in

containers/ecr-viewer/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  return NextResponse.next();
+}
+
+export const config = {
+  matcher: ["/api/:path*"],
+};
+
+const publickey =
+  "-----BEGIN PUBLIC KEY-----\n" + "PLACEHOLDER\n" + "-----END PUBLIC KEY-----";
```
BobanL, line 30: Can you replace placeholder with a `process.env.NBS_PUB_KEY` or something like that?

containers/ecr-viewer/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import { importSPKI, jwtVerify } from "jose";
+export async function middleware(request: NextRequest) {
```
BobanL, line 4: can you put an empty space between import and this function?

containers/ecr-viewer/src/app/api/auth/route.ts
BobanL, line 1: This file isn't needed. It's part of #1366

containers/ecr-viewer/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.1",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "export NODE_ENV=development && next dev",
     "local-dev": "docker compose up db seed-db -d && docker-compose logs && export DATABASE_URL=postgres://postgres:pw@localhost:5432/ecr_viewer_db && next dev",
```
BobanL, line 8: Will this also need `export NODE_ENV=development`?

BobanL: Can you just replace `next dev` with `npm run dev`?

containers/ecr-viewer/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+  const auth = request.cookies.get("authorization")?.value;
```
BobanL, line 5: in #1366, the name of the cookie is `auth-token`

BobanL (changes requested):
Since #1366 takes care of setting the cookie, some of that code can be removed from this PR.

containers/ecr-viewer/package.json
nickbristow: 👍

containers/ecr-viewer/src/middleware.ts
```diff
@@ -1,15 +1,45 @@
...
-    response = NextResponse.redirect(url);
+    const response = NextResponse.redirect(url);
     response.cookies.set("auth-token", auth, { httpOnly: true });
+    return response;
   }
+  return null;
+}
 
-  return response;
+async function authorize_api(req: NextRequest) {
```
emmastephenson, line 26: I'm kind of surprised this passes existing integration tests - shouldn't the auth trigger? Or does the CI runner automatically run in a "development" NODE_ENV?

BobanL: So I was going to comment this somewhere else, but the default for [NODE_ENV is dev](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production#:~:text=Node.js%20assumes%20it%27s%20always%20running%20in%20a%20development%20environment.%20You%20can%20signal%20Node.js%20that%20you%20are%20running%20in%20production%20by%20setting%20the%20NODE_ENV%3Dproduction%20environment%20variable.). I think we need to update our dockerfile to set that variable to production so it will work when deployed.

BobanL (commented):
There is one extra file and might need to update the dockerfile, but otherwise looks good to me.

containers/ecr-viewer/src/middleware.ts
nickbristow: yeah, default is dev, you have to build it to get it to say production

gordonfarrell (approved):
Looks good!

BobanL (approved):

PR #1373, opened 2024-02-23
Adding XML-specific parsing for Orchestration and FHIR-PHDC XML default config

### LINKED ISSUE #1195: Write a new config file in `default_configs/`to specify the FHIR to PHDC workflow
https://github.com/CDCgov/phdi/blob/main/containers/orchestration/app/default_configs/sample-orchestration-config.json

GIVEN: A new configuration file that calls ingestion, standardization, and message parser services for a FHIR bundle
WHEN: A payload that specifies a PHDC XML request
THEN: The orchestration service will read this new configuration file and convert FHIR to PHDC XML.

robertandremitchell:
# PULL REQUEST

## Summary
This adds code to account for the fhir-to-phdc endpoint where an XML is returned. This updates models.py and main.py while also creating a new default config that uses the XML endpoint.

## Related Issue
Fixes #1195 

## Additional Information
* I created a new config. Would we prefer to just update the other, or is it worth keeping both?
* The test does pass locally, but I am honestly not sure if it's currently written correctly. We do want it to return 
`{"message": "Processing succeeded!", "processed_values": "<phdc xml>"}`? 

I know we had previously discussed the idea of encoding it, but I think for demo purposes we're keeping it just as an XML? So I think it's okay, but curious if we need to rewrite this for the demo to just return the phdc XML?


containers/orchestration/app/main.py
```diff
@@ -258,12 +258,15 @@ async def apply_workflow_to_message(
...
-        }
+        content_type = response.headers.get("content-type", "")
+        if "application/xml" in content_type or "text/xml" in content_type:
+            # Handle XML data as a string
+            api_data = response.text
+        else:
+            # Handle JSON data
+            api_data = response.json()
+
+        return {"message": "Processing succeeded!", "processed_values": api_data}
```
DanPaseltiner, line 269: Hmmm this is a little bit tricky. 

First, for readability and robustness could we make the `if` on line 262 check for `application/json`. If the content is JSON we call `response.json()` and in all other cases we call `response.txt`. This simplifies the condition we evaluate and is more robust because we can gracefully handle response types other than XML and JSON in the event that these ever come up intentionally or otherwise.

The problem with this approach for demo purposes is that while we will see XML come back it will not be nicely formatted. I wonder if the appropriate path forward for now is to break our response body convention temporarily and just return straight XML as `/fhir_to_phdc` does currently. To get Fastapi to allow us to return responses of other types we will need to modify the expected return type on line 184.

robertandremitchell: makes sense. had to change other parts of this section to work with returning just XML.

containers/orchestration/app/default_configs/sample-fhir-test-config-xml.json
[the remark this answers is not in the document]
robertandremitchell: added!

robertandremitchell:
looking right now at why this is erroring on GitHub Actions. it is working for me locally, but curious if y'all are running into the same 500 code on this test.

containers/orchestration/app/models.py
```diff
@@ -113,11 +113,11 @@ class OrchestrationResponse(BaseModel):
     The config for responses from the /extract endpoint.
     """
 
-    message: str = Field(
+    message: Optional[str] = Field(
         description="A message describing the result of a request to "
         "the /process endpoint."
     )
-    processed_values: Dict = Field(
+    processed_values: Union[Dict, str] = Field(
```
DavidMcClatchey, line 120: Dipping my toes in here - should the description below be updated to reflect that `processed_values` can be a string as well?

containers/orchestration/app/main.py
```diff
@@ -257,19 +257,37 @@ async def apply_workflow_to_message(
...
+        case "application/xml" | "text/xml":
+            workflow_content = response.content
+        case "application/json":
+            workflow_content = json.dumps(
+                {
+                    "message": "Processing succeeded!",
+                    "processed_values": response.json(),
+                }
+            )
+        case _:
```
bamader, line 287: Oh god it's proliferating everywhere 😂

bamader (approved):
This work all looks right to me. Thanks for working through the XML output on this one (and double thanks for catching all the little nits, typos, syntax things, and consistency issues as you go--sometimes I feel like I've been working in this codebase so long I can't even see mismatches like that anymore so I'm really glad you're able to catch so many). I agree with you and Dan that at this point we probably do want to briefly break with our response convention and just return XML, and the way you have it set up to do that looks right to me. Unfortunately, I am once again not able to reproduce any github testing failures 😕

DanPaseltiner:
> @DanPaseltiner , I think the last part we discussed this PR has yet to tackle is the update to `docker-compose.yml` to include an `environment` section for ingestion for the smarty constants. I think I will need access to a key/token pair in order to fully test.
> 
> Do we want to tackle that in this PR, or merge this then make a smaller PR to update that? I'm good with whichever.

Let's do this in a separate PR as we'll also have to tinker with repository secrets.

DanPaseltiner (approved):
Nice work!

robertandremitchell:
> the update to `docker-compose.yml` to include an `environment` section for ingestion for the smarty constants. I think I will need access to a key/token pair in order to fully test.

makes sense. adding #1386 to the backlog!

PR #1375, opened 2024-02-26
Cleanup kruft

### LINKED ISSUE #1294: Transition orchestration service from old handlers to new

bamader:
## Summary
This PR cleans up all of the out-dated kruft that's currently living in the orchestration service. It deletes old (and no longer needed) payload functions and switches everything over to the new handlers which have been integrated into the main code. This changeset also ditches one of the model input parameters that was only hanging around because of the validation service, but that parameter has been pushed into the config workflow.

## Related Issue
Fixes #1294 

## Additional Information
Before merging, will need to let eCR viewer / streamline team know about the input signature change.


robertandremitchell (changes requested):
Broadly looks good. 

few thoughts/consideration on cleanup.

1. Do we want to also update/simplify the file paths throughout the testing? I noticed that the message-parser used these simple functions in their `utils.py`, for example: https://github.com/CDCgov/phdi/blob/main/containers/message-parser/app/utils.py#L328-L348
2. Relatedly, do we want to move/create copies of the files we reference to be at the phdi/test/assets level?
3. I noticed we had a minor warning here https://github.com/CDCgov/phdi/blob/main/containers/orchestration/tests/integration/test_db.py#L20 that can be resolved by updating the line to
```python
actual = session.get(PostgresFhirDataModel, str(ecr_id))
```

robertandremitchell (approved):
> @robertandremitchell For your question 1 , I don't feel super strongly either way. I think the idea of some cleanup functions are nice, but unlike in the other services the orchestration tests are pulling from multiple different directories so it'll be less cleanup saving than in say the message-parser, and with us needing to code up multiple directories it might be more trouble than it's worth. For 2, I don't think I'm following--currently all of the non config assets the orchestration service loads for tests already live in `phdi/tests/assets`, just like the other services' tests. Are you saying we should copy/move them into the orchestration service itself? For 3, good catch, I've implemented that now and we should no longer get that warning.

Makes sense re: 1. I don't feel strongly either.
For 2, I think there may be a small handful of tests that reference other building blocks. I agree that given orchestration's different role in all this it's probably okay that they pull from other directory tests, or should be relocated as part of a bigger effort across all the building blocks.

PR #1382, opened 2024-02-26
fix custom filter code smells

BobanL:
# PULL REQUEST

## Summary
Fix code smells that were appearing at the bottom of PRs. 

## Additional Information
No changes were made to logic.

### DIFF
--- a/containers/fhir-converter/CustomFhir/CustomFilters.cs
+++ b/containers/fhir-converter/CustomFhir/CustomFilters.cs
@@ -16,8 +16,9 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
   /// </summary>
   public partial class Filters
   {
-    private static readonly HashSet<string> supportedTags = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "br", "li", "ol", "p", "span", "table", "tbody", "td", "textarea", "th", "thead", "tr", "u", "ul" };
-    private static readonly Dictionary<string, string> replaceTags = new Dictionary<string, string>{
+    private static readonly HashSet<string> supportedTags = new(StringComparer.OrdinalIgnoreCase) { "br", "li", "ol", "p", "span", "table", "tbody", "td", "textarea", "th", "thead", "tr", "u", "ul" };
+    private static readonly Dictionary<string, string> replaceTags = new()
+    {
         {"list", "ul"},
         {"item", "li"}
     };
@@ -27,7 +28,7 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
     {
       if (item is Dictionary<string, object> dict)
       {
-        return new List<Dictionary<string, object>> { dict };
+        return [dict];
       }
       else if (item is IEnumerable<object> collection)
       {
@@ -37,7 +38,7 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
       {
         return collectionTwo.ToList();
       }
-      return new List<Dictionary<string, object>>();
+      return [];
     }
 
     private static Dictionary<string, object>? DrillDown(Dictionary<string, object> item, List<string> list)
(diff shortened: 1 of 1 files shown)


containers/fhir-converter/CustomFhir/CustomFilters.cs
```diff
@@ -16,65 +16,79 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
...
-            return collectionTwo.Select(x => x as Dictionary<string, object>).ToList();
-        }
-        return new List<Dictionary<string, object>>();
+      if (item is Dictionary<string, object> dict)
+      {
+        return [dict];
+      }
+      else if (item is IEnumerable<object> collection)
+      {
+        return collection.Cast<Dictionary<string, object>>().ToList();
```
BobanL, line 35: `.Cast<T>()` will cast each element to `T`. 
https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.cast?view=net-8.0

containers/fhir-converter/CustomFhir/CustomFilters.cs
```diff
@@ -16,65 +16,79 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
...
+      {
         return item;
       }
       string firstElement = list.First(); // Retrieve the first element
       list.Remove(firstElement);
-      var element = item.GetValueOrDefault(firstElement, null);
-      if(element != null && list.Count > 0){
-        return DrillDown(element as Dictionary<string, object>, list);
-      }else if(element !=null && list.Count == 0){
+      if (item.TryGetValue(firstElement, out object? element) && list.Count > 0)
```
BobanL, line 52: Rather than `GetValueOrDefault`, we can use [`TryGetValue`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-8.0) instead. If there is no value, it will return false. If there is a value it will return true and set the value to the second parameter.

containers/fhir-converter/CustomFhir/CustomFilters.cs
```diff
@@ -16,65 +16,79 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
...
-        {
-            return new List<Dictionary<string, object>> { dict };
-        }
-        else if (item is IEnumerable<object> collection)
-        {
-            return collection.Select(x => x as Dictionary<string, object>).ToList();
-        }
-        else if (item is IEnumerable<Dictionary<string, object>> collectionTwo)
-        {
-            return collectionTwo.Select(x => x as Dictionary<string, object>).ToList();
```
BobanL, line 38: It's already of type `Dictionary<string, object>` so we don't need to cast it.

containers/fhir-converter/CustomFhir/CustomFilters.cs
```diff
@@ -16,65 +16,79 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
...
+      }
+      else if (item is IEnumerable<object> collection)
+      {
+        return collection.Cast<Dictionary<string, object>>().ToList();
+      }
+      else if (item is IEnumerable<Dictionary<string, object>> collectionTwo)
+      {
+        return collectionTwo.ToList();
+      }
+      return [];
```
BobanL, line 41: Short hand

containers/fhir-converter/CustomFhir/CustomFilters.cs
```diff
@@ -85,70 +99,72 @@ public static string ConcatenateTds(IDictionary<string, object> data)
 
     private static string WrapHtmlValue(string key, object value)
```
BobanL, line 100: everything below this is linting

containers/fhir-converter/CustomFhir/CustomFilters.cs
```diff
@@ -25,56 +25,70 @@ public partial class Filters
...
+      {
         return null;
       }
     }
     public static string ConcatenateTds(IDictionary<string, object> data)
     {
       var result = new List<string>();
-      var dataDictionary = (data as Dictionary<string, object>);
-      var component = DrillDown(dataDictionary, new List<string> {"text"}) ??
+      var dataDictionary = (Dictionary<string, object>)data;
```
KennethSkylight, line 67: what is the difference...?

KennethSkylight (approved):
LGTM just a curious question

containers/fhir-converter/CustomFhir/CustomFilters.cs
BobanL: [So direct casting (with parentheses), will throw an error if the object is not that type. With the "as" casting, it will just return a null value for data if it is not that type. ](https://stackoverflow.com/questions/132445/direct-casting-vs-as-operator)
It's a little confusing, but since we know the the type of "Data" to be a dictionary we can make that casting without any issue.

PR #1383, opened 2024-02-26
ECR viewer redirect and Cypress e2e test

### LINKED ISSUE #1371: Middleware to fix eCR Viewer for AWS Ingress
AWS ingress routes things to include service name (`/ecr-viewer/view-data` instead of `/view-data`.) We need to add middleware to redirect these routes.

gordonfarrell:
# PULL REQUEST

## Summary
NEW FUNCTIONALITY: Redirects users who navigate to ecr-viewer/view-data etc to just /view-data via next.js config

NEW TESTING LIBRARY FOR E2E TESTS: So, unit testing a next.js config is really hard and they just suggest you use e2e. We added Cypress as a dev dependency due to both Angela and I having familiarity and it being a pretty good option for e2e/journey testing in general if we want to start expanding that part of our testing strategy. Will bring up this addition at stand tomorrow but also let us know if there are any red flags to this approach.

NEW TASK ON GITHUB test.yaml WORKFLOW: Added cypress testing into the automated test suite.

## Related Issue
Fixes #1371

## Additional Information

## Checklist


containers/ecr-viewer/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
emmastephenson, line 1: Do we need this sample fixture?

containers/ecr-viewer/cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
emmastephenson, line 1: Same question here - do we need this file?

emmastephenson (approved):
Overall LGTM! Just some small Cypress files that may need to be deleted. 
Would you mind adding a ticket to the backlog to create more e2e tests for our happy path and bad path flows? Now that we have the basic setup in place, we should expand em!

containers/ecr-viewer/cypress/fixtures/example.json
gordonfarrell: Not sure about this sample fixture, it can probably go. Was a bit hesitant to delete after trying to delete the support files and it got really mad haha.

containers/ecr-viewer/cypress/support/commands.ts
gordonfarrell: This one we do need, Angela and I initially thought it could go as its mostly comment anyway. Nope! Broke everything haha

gordonfarrell: Will double check actually, maybe it only needs the other support file

angelathe: Cypress broke when we removed these files yesterday for some reason 🙄

containers/ecr-viewer/cypress/fixtures/example.json
angelathe: Removed!

containers/ecr-viewer/cypress/support/commands.ts
angelathe: Note: kept this file

PR #1385, opened 2024-02-27
Group repeating question observations by organizer

### LINKED ISSUE #1369: Group component obs by organizer

m-goggins:
# PULL REQUEST

## Summary
This PR adds an `organizer` element to the repeating sections and groups observations from the `component` section of an Observation resource under a single organizer.

## Related Issue
Fixes #1369 

## Additional Information

## Checklist


containers/message-parser/app/phdc/builder.py
```diff
@@ -279,6 +279,38 @@ def build_body(self):
...
+        appropriate section.
+
+        :param section: Section XML element.
+        :section_type: The Section observations are added to.
+        :return: Section XML element with added observations.
+        """
+        if section_type == "social_history":
+            data = self.input_data.social_history_info
+        else:
+            data = self.input_data.clinical_info
```
DanPaseltiner, line 298: Since this function gets called in `_build_clinical_info` and `_build_social_history_info` could we have the selection of the data happen in those methods and then pass to this method by replacing  the `section_types` parameter with a `data` parameter? This method currently needs to do 2 things, select the appropriate data and then add the observations. By removing the responsibility for data selection we improve our separation of concerns, this method no longer needs to know what section it is adding observations to just how to add observations to a given section.

containers/message-parser/app/phdc/models.py
```diff
@@ -166,4 +166,6 @@ class PHDCInputData:
     organization: List[Organization] = None
     clinical_info: List[Observation] = field(default_factory=list)
     social_history_info: List[Observation] = field(default_factory=list)
-    repeating_questions: List[Observation] = field(default_factory=list)
+    repeating_questions: List[Union[Observation, List[Observation]]] = field(
+        default_factory=list
+    )
```
DanPaseltiner, line 171: This mixture of observation objects and lists of observation objects makes me a little nervous. Do we have any examples where we need to handle a single observation resource? In this case is there any reason why we use a list of observations with a length of 1? Having to constantly check type as we iterate through these values seems like a code smell to me. I could definitely be missing something here though.

DanPaseltiner (changes requested):
Thanks for working through this @m-goggins! One question I'd like to discuss and one small change I think we should make before merging.

containers/message-parser/app/phdc/models.py
m-goggins: I think the single observation comes up in the case where we don't have a `component` section but we can definitely change to a list length of 1. I agree that will make it simpler.

containers/message-parser/app/phdc/builder.py
m-goggins: Yep, that's a good suggestion!

containers/message-parser/app/phdc/builder.py
```diff
@@ -279,6 +279,27 @@ def build_body(self):
...
+        :data: The Observations that will be added to the Section element.
+        :return: Section XML element with added observations.
+        """
+        for observation in data:
+            for c in observation:
+                entry = ET.Element("entry", {"typeCode": "COMP"})
+                observation_element = self._build_observation(c)
+                entry.append(observation_element)
+                section.append(entry)
+        return section
```
DanPaseltiner, line 301: 🚀

DanPaseltiner, line 301: 🚀

DanPaseltiner (approved):
Thanks @m-goggins!

PR #1394, opened 2024-02-28
Clean up tests and config model

### LINKED ISSUE #1235: Audit data used for orchestration services tests
Determine if any test data is too fake to provide meaningful tests.

Acceptance Criteria:
Create tickets to address specific test data deficiencies  that are identified.

bamader:
## Summary
This PR makes a second round of cleanup in the orchestration service, implementing a number of changes:
- audits test data to ensure that only realistically formatted and filled data is used for unit tests, unless otherwise noted (e.g. one of the sections of 5 mock side effects)
- removes unnecessary parameters from tests inputs
- changes the config model input to reflect creating orchestration workflows, not message parser schemas
- removes unnecessary assets
- alters request and response examples to reflect new signature

## Related Issue
Fixes #1235 

## Additional Information
The tests and signature for the `save_to_db` function were just modified in another PR by Nick to fix some compatibility issues and loading bugs in the demo UI, so this PR will leave those tests and signatures completely untouched. Once `save_to_db` is implemented as a fully fledged endpoint, we'll revisit them.


containers/orchestration/app/default_configs/test_config.json
```diff
@@ -1,57 +1,54 @@
...
-      "credential_manager": "azure",
-      "url": "http://localhost:8085"
-    }
-  },
-  "steps": [
+  "workflow": [
     {
       "service": "validation",
-      "endpoint": "/validate"
+      "url": "http://localhost:8081",
```
robertandremitchell, line 5: shouldn't URLs be pulled from .env? or because it's the test we want them hardcoded?

containers/orchestration/app/default_configs/test_config.json
```diff
@@ -1,57 +1,54 @@
...
+      "endpoint": "/validate",
+      "params": {
+        "include_error_types": "error"
+      }
     },
     {
-      "service": "fhir_converter",
-      "endpoint": "/convert-to-fhir"
+      "service": "fhir-converter",
+      "url": "http://localhost:8082",
```
DanPaseltiner, line 13: Do we even need the `url` key in the config anymore? We are setting urls for each service as environment variables and accessing via service name. Am I missing something? At the very least we have these URLs hardcoded for local dev, but they will change with every deploy to every individual cloud environment.

DanPaseltiner (approved):
Thanks @bamader! It's really great to see all the refactoring and cleaning coming to together to really level up our code quality here. One comment below that I _think_ might be an easy tweak before merging, but it doesn't have to be blocking. Additionally, in the spirit of the this PR, but out of its scope, there are two other cleaning tasks I think we should take on to help with organization and minimizing the size of our container images.

1. We should introduce a `dev_requirements.txt` file and move all dependencies required for testing and development in there and out of `requirements.txt`. This will ensure that we only install the dependencies required to run the service in the container images without additional bloat.

2. Similar to 1 we should introduce a `tests/assets/` directory and move all data for testing in there. This will separate data that used to document example requests and responses from data used for integration/unit testing and prevent us from including the test data in the container images as additional bloat.

I can work on tickets for these two things and @robertmitchellv has just done similar work in the message parser.

robertandremitchell (approved):
minor quibble but lgtm!

containers/orchestration/app/default_configs/test_config.json
bamader: Yep they can completely disappear now, great catch!

containers/orchestration/app/default_configs/test_config.json
bamader: Nope, you and Rob are totally right. This is outdated and has been deleted.

bamader:
@DanPaseltiner I created a `tests/assets` folder and moved our one file for orchestration testing in there, but I didn't address splitting up of the requirements file so that one still needs a ticket

Issue #1149, opened 2024-01-25
Create machine-readable CSV for condition code mappings


emmastephenson:
## Action Requested
Create a machine-readable table that the FHIR converter can use to stamp condition codes onto FHIR resources. This table should take the core table published with RCTC codes ([spreadsheet](https://ecr.aimsplatform.org/cms/resources/blocks/reference-rctc-3-22.xlsx)) and consolidate it into a single sheet, removing the extraneous header information. 

## Acceptance Criteria
A single csv exists with mappings available from all 6 mappings (Diagnosis_Problem, Organism_Substance, Lab_Order_Test_Name, Lab_Obs_Test_Name, Suspected_Disorder, and Medications).

The table should have the following format:

| Type | Name | OID | Code System | Code System OID | Status | Condition Name | Condition Code | Condition Code System |
| ----- | ------ | --- | -------------- | ----------------- | ------- | ---------------- | --------------- | ------------------------ |
| Lab Observation | COVID_19 (Tests for SARS_CoV_2 IgG Antibody) | 2.16.840.1.113762.1.4.1146.1152 | LOINC | 2.16.840.1.113883.6.1 | Published | Disease caused by severe acute respiratory syndrome coronavirus 2 (disorder) | 840539006 | SNOMEDCT

## Testing
We'll need to verify that the table is readable by the FHIR converter.

## Background & Context
This table is the source of truth we'll need to associate specific labs and clinical information with a condition in an eCR.

PR #1398, opened 2024-02-29
Modify script that converts RCTC xlsx to readable csv

angelathe:
# PULL REQUEST

## Summary
Modifies script to also convert and output Expansion List from RCTC xlsx

## Related Issue
Adds additional fix to #1149 

## Additional Information

## Checklist


containers/fhir-converter/rctc-to-json.py
```diff
@@ -6,16 +6,24 @@
 
 # arg list: 1) input file location/name 2) output file location/name
 wb = openpyxl.load_workbook(sys.argv[1])
-jsonData = {}
+jsonData_grouping = {}
+jsonData_expansion = {}
+top_rows_buffer = 17
```
nickbristow, line 11: just a question, what are these hard coded numbers?

angelathe: @nickbristow In the core table with the RCTC codes (ex. [Link](https://ecr.aimsplatform.org/cms/resources/blocks/reference-rctc-3-22.xlsx)) there are a couple misc. rows before the actual table data begins. So `top_rows_buffer` indicates that there are 17 rows before the Grouping Data table, and `tables_buffer` represents the number of rows between the Grouping Data table and the Expansion Data table. 

Values seem to be consistent (so far) between table updates, but lmk if you think we should either 1) write a comment to clarify this, or 2) find the beginning rows of the tables without using hard-coded vals!

angelathe: fyi, added comments explaining what these hard-coded vals are

JNygaard-Skylight (commented):
My only suggestion is to look into using the [`argparse` library](https://docs.python.org/3.10/library/argparse.html). It will make validating the inputs easier, and will add documentation to the command line. But everything as is looks good to me, so I'll leave it up to you if you want to look into adding `argparse`.

JNygaard-Skylight (approved):
LGTM!

PR #1421, opened 2024-03-06
[RELEASE] - 1.2.4 (Use basePath in production for ecr-viewer)

### LINKED ISSUE #1272: Create AWS ecr-viewer release
## Action Requested
Write Terraform to publish the DIBBs pipeline, eCR Viewer, and S3 bucket to an AWS subscription.

## Acceptance Criteria
It's possible to hit the `/process` endpoint with a ZIP file of eCR data, then view that data in the eCR Viewer.

Unauthenticated requests to `/view-data` should be rejected.

## Testing
Lots of manual testing on AWS will be required.

## Background & Context
NBS needs the eCR viewer working more or less end-to-end in AWS to be able to develop locally against it. (The "less" is that we won't need a function app to trigger the pipeline; they'll just hit the `process` endpoint manually.)

## Additional Information
There's a good chance this should be split into smaller tickets! But a mitigating factor is that Nick C. has been working to publish the pipeline to AWS for some time now, so the main resources we'll need to create should simply be:
* Adding the eCR viewer Helm chart into the EKS instance
* Creating the S3 bucket
* ~Ensuring all the env vars are in place for authentication~ (blocked by auth spike)
* Ensuring that networking works for the eCR Viewer (we may need to include a middleware redirect for AWS - see #1193 for details)

nickclyde:
# PULL REQUEST

## Summary
Use basePath in production for ecr-viewer

## Related Issue
Fixes #1272


KennethSkylight (approved):
how many releases will it take

Issue #1406, opened 2024-03-04
Add `xsi:type` to `<value>` child element of `<observation>` in PHDC


robertmitchellv:
## Action Requested
When creating observations it is required that the child element `<value>` inside of `<observation>` have an `xsi:type` declared in order for that data to be valid. It is possible to remove the `<value>` element and for the XML to be considered valid but if the tag is present it needs the `xsi:type` listed.

Unfortunately, the interface spec does not really flesh out what is and isn't acceptable for these `xsi:type` datatypes, and I've only observed the following in the case report as examples:
- `CE` for coded elements, which include a code, display name, and code system
- `ST` for string data, to represent textual information with no attributes and often with that weird `<![CDATA[string]]>` between the opening and closing `<value>` tags
- `TS` for time-stamped data, which we've talked about; set in a `value=` attribute of the `<value>` element
- `II` which seems like instance identifiers? used for unique identification.

We will need some post-process to inspect the data that has been passed through to determine what code should be included **prior** to it being input for our dataclasses. Ideally it will handle the above `xsi:type`s.

## Acceptance Criteria
When values that contain coded elements `CE`, string `ST`, or timestamps `TS` are correctly included in `value` with the appropriate `xsi:type`. Additionally if there is no heuristic to follow to include a value with a correct `xsi:type` that `<value>` is **not** added to the PHDC output since it will fail validation. This will have to be the line in the sand for this ticket in order to push us forward in terms of CI validation since we need more context in order to continue to develop this piece. Additionally, tests will need to be adjusted and they must pass locally and in CI.

Issue #1069, opened 2024-01-11
When the PHDC endpoint `/fhir-to-phdc` is finished, run the returned PHDC XML output through an integration test via GitHub Actions


robertmitchellv:
When the PHDC endpoint `/fhir-to-phdc` is able to produce PHDC XML output, add an integration test via GitHub Actions validating its output ensuring the PHDC XML is well formed and valid.

PR #1423, opened 2024-03-06
Add PHDC XML validation to integration tests

robertmitchellv:
# PULL REQUEST

## Summary
Adding PHDC XML validation to integration tests. This will take a prepared FHIR bundle and run it through the PHDC Case Report parser to the PHDC Builder to produce PHDC Case Report XML that is checked against the [HL7 CDA schema](https://github.com/HL7/cda-core-2.0), which includes the SDTC extensions.

Additionally we're also adding the ability to set the correct `xsi:type` for `<value>` child elements of `<observation>`, which affects validation.

## Related Issue
Fixes:
- #1069 
- #1406 

## Additional Information
XML validation is pain 🤕


robertmitchellv:
@DanielPaseltiner i think i'm going to address the new FHIR bundle after this lands if that sounds ok? unless you have strong feelings then i can add it to this work

robertmitchellv:
@DanPaseltiner @m-goggins i added some changes after Dan approved because i was in the middle of moving some things around to make it easier to read assets from the different places in the tests so want to make sure you both have a chance to weigh in before i merge

m-goggins (approved):

Issue #1344, opened 2024-02-20
Save data from eCR Viewer to S3 bucket


emmastephenson:
## Action Requested
Update the new `save-data` endpoint of the eCR viewer to work with either a locally-running Postgres instance or an S3 bucket.

## Acceptance Criteria
eCRs sent through the orchestration service are saved to an S3 bucket by the eCR Viewer. Those eCRs can then be viewed when given the proper UUID.

In short: eCR viewer works with S3 end-to-end.

## Testing
Manual testing with S3 will be needed.

## Designs
n/a - this is a backend-only ticket.

## Background & Context
As part of the effort to make the orchestration service more dynamic (and less specific to the eCR viewer) we're moving the functionality to save FHIR bundles to the viewer in #1331. This is continued follow-up work to that effort.

PR #1435, opened 2024-03-07
Add API endpoint on eCR Viewer to save FHIR bundle

### LINKED ISSUE #1331: Add route to save FHIR bundles in eCR Viewer container
## Action Requested
Create an API endpoint on the eCR Viewer container that will save a FHIR bundle to data storage.

## Acceptance Criteria
With a request to a new endpoint on the eCR viewer (`save-fhir-data`?) and a FHIR bundle as a parameter, the FHIR bundle is saved to the local postgres database of the eCR Viewer.

## Testing
Unit testing will be required for the new routes. 

## Background & Context
Right now, the orchestration service saves data directly to a database (rather than calling out to another service.) This poses some problems:
* We have to maintain database connections in two different containers (orchestration and the eCR Viewer)
* It's harder to make flexible calls within the orchestration service, because this pattern doesn't conform to the others (this is causing problems for VIPER)

Adding the functionality to the eCR viewer instead will cut down on our maintenance needs in the long run.

## Additional Information

* This is blocking [#1270](https://app.zenhub.com/workspaces/streamline-ecr-6480bf2ee530095ab41ebbe9/issues/gh/cdcgov/phdi/1270)
* I think this new route should live within the [`api` folder](https://github.com/CDCgov/phdi/tree/main/containers/ecr-viewer/src/app/api), as it'll be a backend-only route (we shouldn't have a frontend for saving data)
* There will be a second ticket to update the orchestration service to point at the new endpoint

angelathe:
# PULL REQUEST

## Summary
Create an API endpoint on the eCR Viewer container that will save a FHIR bundle to Postgres. Request body takes in a `fhirBundle` and a `saveSource` -- see docstring for more info.

## Related Issue
Fixes #1331 

## Additional Information: background on why we didn't include tests
- Tried writing tests with Jest, but came into a host of issues (Related Link here: ([Async Server Components])(https://nextjs.org/docs/app/building-your-application/testing#async-server-components)
- Next, tried E2E testing, where I wrote tests in Cypress that stubbed the API response. However, the tests were just asserting directly on our own mocked response, and we couldn't (yet) figure out a way to test the outgoing db request.

## Checklist
- [ ] Make a POST request to http://localhost:3000/api/save-fhir-data 
- [ ] Example request body below:
`{
    "fhirBundle": {
        "resourceType": "Bundle",
        "type": "batch",
        "entry": [
        {
            "fullUrl": "urn:uuid:12345",
            "resource": {
                "resourceType": "Composition",
                "id": "12345"
            }
        }
        ]
    },
    "saveSource": "postgres"
}`


containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (fhirBundle && ecrId) {
+      const { ParameterizedQuery: PQ } = pgPromise;
+      const addFhir = new PQ({
+        text: "INSERT INTO fhir VALUES ($1, $2) RETURNING ecr_id",
+        values: [ecrId, fhirBundle],
+      });
+
+      try {
+        const saveECR = await database.one(addFhir);
+        console.log("Created Resource with: ", saveECR);
```
JNygaard-Skylight, line 52: You should remove this `console.log`

JNygaard-Skylight (approved):
Looks good to me!

nickbristow (changes requested):
because this will also save to s3, we should change the route from `save-fhir-data` to something like `save-fhir-data-db` or something like that, im bad at names

angelathe:
> because this will also save to s3, we should change the route from `save-fhir-data` to something like `save-fhir-data-db` or something like that, im bad at names

@nickbristow My impression was that the next ticket (#1344) updates this endpoint to also include s3 (rather than making a new endpoint)? If so, not sure if `db` would make sense to have in the route name

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
angelathe: done!

angelathe:
> > because this will also save to s3, we should change the route from `save-fhir-data` to something like `save-fhir-data-db` or something like that, im bad at names
> 
> @nickbristow My impression was that the next ticket (#1344) updates this endpoint to also include s3 (rather than making a new endpoint)? If so, not sure if `db` would make sense to have in the route name

Note: Added `saveSource` property in request body that indicates where the FHIR data should be saved (`postgres` or `s3`). Docstrings are also updated!

nickbristow (approved):

PR #1442, opened 2024-03-11
Replace "N/A" with "No data" in EcrViewer

### LINKED ISSUE #1343: Change "N/A" to "No data" in eCR Viewer
## Action Requested
Change "N/A" everywhere in the eCR Viewer to be "No data" with different styling (gray font color and italicized)

## Acceptance Criteria
Wherever data isn't available, the field value reads as "No data" with new styling.

## User Value
It's not clear to the user what N/A really means (not applicable? not available?). Replacing it with more plain language and making the styling different from the rest of the text better indicates to the user that this data just didn't come in with the original eCR

## Testing
(Sarah does not know this info)

## Designs
[Figma link](https://www.figma.com/proto/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?page-id=871%3A2283&type=design&node-id=1024-136062&viewport=1050%2C947%2C0.13&t=sZFPN9VIDj6qdeTO-1&scaling=min-zoom&starting-point-node-id=1606%3A27619&mode=design)

## Design Review
things to look for:
- using the right shade of gray for the text color (should be base gray, hex: 71767A)
- the text is italicized
- capitalization matches designs

BobanL:
# PULL REQUEST

## Summary
Replace "N/A" with "No data"

## Related Issue
Fixes #1343 

## Additional Information
Some refactoring was done to make things simpler like just leaving "Value" as undefined for "N/A" options. 

## Demo
Now:
image
Before:
image

### DIFF
--- a/containers/ecr-viewer/src/app/utils.tsx
+++ b/containers/ecr-viewer/src/app/utils.tsx
@@ -175,8 +175,8 @@ const formatDateTime = (dateTime: string) => {
 };
 
 export const formatDate = (date: string) => {
-  if (!date || date === null) {
-    return "N/A";
+  if (!date) {
+    return undefined;
   }
 
   const options: Intl.DateTimeFormatOptions = {
@@ -293,10 +293,9 @@ const formatTable = (
     columns.forEach(function (column, index) {
       let isFirstCell = index === 0;
 
-      let rowCellData;
-      evaluate(entry, mappings[column.infoPath])[0]
-        ? (rowCellData = evaluate(entry, mappings[column.infoPath])[0])
-        : (rowCellData = "N/A");
+      let rowCellData = evaluate(entry, mappings[column.infoPath])[0] ?? (
+        <span className={"text-italic text-base"}>No data</span>
+      );
 
       let rowCell = isFirstCell ? (
         <th key={`row-header-${index}`} scope="row" className="text-top">
@@ -321,7 +320,7 @@ const formatTable = (
       <tbody>{tableRows}</tbody>
     </>
   );
-  const table = (
+  return (
     <Table
       bordered={false}
       fullWidth={true}
@@ -332,8 +331,6 @@ const formatTable = (
       {tableContent}
     </Table>
   );
-
-  return table;
 };
 
 const extractTravelHistory = (
@@ -358,8 +355,8 @@ const extractTravelHistory = (
   )[0];
   if (startDate || endDate || location || purposeOfTravel) {
     return `Dates: ${startDate} - ${endDate}
-       Location(s): ${location ?? "N/A"}
-       Purpose of Travel: ${purposeOfTravel ?? "N/A"}
+       Location(s): ${location ?? "No data"}
+       Purpose of Travel: ${purposeOfTravel ?? "No data"}
        `;
   }
   return undefined;
@@ -606,9 +603,7 @@ export const returnProblemsTable = (
   ];
 
   problemsArray.forEach((entry) => {
-    entry.onsetDateTime
-      ? (entry.onsetDateTime = formatDate(entry.onsetDateTime))
-      : (entry.onsetDateTime = "N/A");
+    entry.onsetDateTime = formatDate(entry.onsetDateTime);
   });
 
   problemsArray.sort(function (a, b) {
@@ -635,9 +630,7 @@ export const returnImmunizations = (
   ];
 
   immunizationsArray.forEach((entry) => {
-    entry.occurrenceDateTime
-      ? (entry.occurrenceDateTime = formatDate(entry.occurrenceDateTime))
-      : (entry.occurrenceDateTime = "N/A");
+    entry.occurrenceDateTime = formatDate(entry.occurrenceDateTime);
   });
 
   immunizationsArray.sort(function (a, b) {
@@ -667,9 +660,7 @@ export const returnProceduresTable = (
   ];
 
   proceduresArray.forEach((entry) => {
-    entry.performedDateTime
-      ? (entry.performedDateTime = formatDate(entry.performedDateTime))
-      : (entry.performedDateTime = "N/A");
+    entry.performedDateTime = formatDate(entry.performedDateTime);
   });
 
   proceduresArray.sort((a, b) => {
@@ -759,7 +750,6 @@ const evaluateData = (data: DisplayData[]) => {
   data.forEach((item) => {
     if (!item.value || (Array.isArray(item.value) && item.value.length === 0)) {
       unavailableData.push(item);
-      item.value = "N/A";
     } else {
       availableData.push(item);
     }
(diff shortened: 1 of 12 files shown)


containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -7,7 +7,7 @@ import classNames from "classnames";
 
 export interface DisplayData {
   title: string;
-  value: string | React.JSX.Element | undefined;
+  value?: string | React.JSX.Element;
```
BobanL, line 10: This means the same thing

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -175,8 +175,8 @@ const formatDateTime = (dateTime: string) => {
 };
 
 export const formatDate = (date: string) => {
-  if (!date || date === null) {
-    return "N/A";
+  if (!date) {
+    return undefined;
```
BobanL, line 179: Return undefined instead of "N/A" to make it more clear that it's not a date.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -293,10 +293,9 @@ const formatTable = (
     columns.forEach(function (column, index) {
       let isFirstCell = index === 0;
 
-      let rowCellData;
-      evaluate(entry, mappings[column.infoPath])[0]
-        ? (rowCellData = evaluate(entry, mappings[column.infoPath])[0])
-        : (rowCellData = "N/A");
+      let rowCellData = evaluate(entry, mappings[column.infoPath])[0] ?? (
+        <span className={"text-italic text-base"}>No data</span>
+      );
```
BobanL, line 298: If evaluate has no value, add formatting to the cell data.

sarahtress:
@BobanL this looks good to me, thanks!

nickbristow (changes requested):
Can you add some JSDoc commenting to some of the major functions you've changed?

containers/ecr-viewer/src/app/view-data/components/tests/ActiveProblems.test.tsx
```diff
@@ -281,7 +282,7 @@ describe("Active Problems Table", () => {
       },
     ];
     container = render(
-      returnProblemsTable(activeProblemsData, fhirPathMappings),
+      returnProblemsTable(activeProblemsData, fhirPathMappings)!,
```
BobanL, line 285: [The `!` is a non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). This fixes the problem where the IDE was flagging that `render` only accepts `JSX.Elements` or `string` and not undefined.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -20,7 +27,7 @@ export interface ColumnInfoInput {
 }
 
 export const formatPatientName = (
-  fhirBundle: Bundle | undefined,
+  fhirBundle: Bundle,
```
BobanL, line 30: Removed the option to send `undefined` since it should never be called with that value.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -267,15 +273,23 @@ const formatVitals = (
...
+ * @param {FhirResource[]} resources - An array of FHIR Resources representing the data entries.
+ * @param {PathMappings} mappings - An object containing the FHIR path mappings.
+ * @param {ColumnInfoInput[]} columns - An array of objects representing column information.
+ *                                      The order of columns in the array determines the order of appearance.
+ * @param {string} caption - The caption for the table.
+ * @returns {React.JSX.Element} - A formatted table React element.
+ */
 const formatTable = (
-  resources: React.JSX.Element[],
+  resources: FhirResource[],
```
BobanL, line 286: This was incorrectly typed to `React.JSX.Element[]` and should be a `FhirResource[]` instead.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -267,59 +273,46 @@ const formatVitals = (
...
-    headers.push(header);
-  });
-
-  let tableRows: React.JSX.Element[] = [];
-  resources.forEach((entry, index) => {
-    let rowCells: React.JSX.Element[] = [];
-    columns.forEach(function (column, index) {
-      let isFirstCell = index === 0;
+): React.JSX.Element => {
+  let headers = columns.map((column, index) => (
```
BobanL, line 291: Replace `.forEach` with [`.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) for clarity since we need the return from every value in `columns`.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -267,59 +273,46 @@ const formatVitals = (
...
+    let rowCells = columns.map((column, index) => {
       let rowCellData = evaluate(entry, mappings[column.infoPath])[0] ?? (
         <span className={"text-italic text-base"}>No data</span>
       );
-
-      let rowCell = isFirstCell ? (
-        <th key={`row-header-${index}`} scope="row" className="text-top">
-          {rowCellData}
-        </th>
-      ) : (
```
BobanL, line 305: Remove the `isFirstCell` logic because for [accessibility on a table with only one header only the first row or column needs to have the scope. ](https://www.w3.org/WAI/tutorials/tables/one-header/#table-with-ambiguous-data)

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -606,19 +613,25 @@ export const returnProblemsTable = (
...
 
-  problemsArray.sort(function (a, b) {
-    return (
-      new Date(b.onsetDateTime).getTime() - new Date(a.onsetDateTime).getTime()
-    );
-  });
+  problemsArray.sort(
+    (a, b) =>
+      new Date(b.onsetDateTime ?? "").getTime() -
+      new Date(a.onsetDateTime ?? "").getTime(),
```
BobanL, line 619: The weird comma at the end is added by linter.

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -606,19 +613,25 @@ export const returnProblemsTable = (
...
   });
 
-  problemsArray.sort(function (a, b) {
-    return (
-      new Date(b.onsetDateTime).getTime() - new Date(a.onsetDateTime).getTime()
-    );
-  });
+  problemsArray.sort(
+    (a, b) =>
+      new Date(b.onsetDateTime ?? "").getTime() -
```
BobanL, line 618: Default to empty string since `new Date()` only accepts string and `Condition.onsetDateTime` can be string or undefined.

PR #1456, opened 2024-03-11
Save FHIR data to S3 in ecr-viewer

### LINKED ISSUE #1344: Save data from eCR Viewer to S3 bucket

angelathe:
# PULL REQUEST

## Summary
- Add function in `save-fhir-data` API route to save to s3
- Adds tests that mock s3
- Add `setup-local-env` script in `package.json` to automate the setup of `.env.local`

## Related Issue
Fixes #1344 

## Additional Information
These tests aren't true integration tests (doesn't spin up an s3 instance, etc.,).

## Checklist
- [ ] Run http://localhost:3000/api/save-fhir-data with example body below, and check that file was correctly saved to s3 bucket.

Example request body:
```
{
    "fhirBundle": {
        "resourceType": "Bundle",
        "type": "batch",
        "entry": [
        {
            "fullUrl": "urn:uuid:12345",
            "resource": {
                "resourceType": "Composition",
                "id": "12345"
            }
        }
        ]
    },
    "saveSource": "s3"
}
```


containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -50,6 +59,16 @@ export async function POST(request: NextRequest) {
     );
   }
 
+  if (!fhirBundle || !ecrId) {
+    return new NextResponse(
+      JSON.stringify({
```
BobanL, line 64: I think we can just use [`NextResponse.json` since we return a JSON response](https://nextjs.org/docs/app/api-reference/functions/next-response#json). This can be applied throughout the file.
```ts
return NextResponse.json(
{ message:
          "Invalid request body. Body must include a FHIR bundle with an ID.",
      }},
{ status:400 }`
```

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -50,6 +59,16 @@ export async function POST(request: NextRequest) {
...
+        message:
+          "Invalid request body. Body must include a FHIR bundle with an ID.",
```
BobanL, line 66: This property name `message` is inconsistent with line 57. We use `error` for the message in that error.

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -60,42 +79,141 @@ export async function POST(request: NextRequest) {
...
+      {
+        message:
+          "Invalid source. Please provide a valid source (postgres or s3)",
+      },
+      { status: 400 },
+    );
+  }
+}
+
+/**
```
BobanL, line 97: Can we move all of these other functions into a different file?

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -60,42 +79,141 @@ export async function POST(request: NextRequest) {
...
+            "resource": {
+                "resourceType": "Composition",
+                "id": "12345"
+            }
+        }
+        ]
+    }
+ * const ecrId = 'some-unique-ecr-id';
+ *
+ * await saveToS3(fhirBundle, ecrId);
```
BobanL, line 123: I'm ok to have a conversation, but I don't know how valuable these examples are. The params make it clear what types should be passed in.

BobanL (changes requested):
I have some thoughts about the jsdocs, and the NextResponse not being consistently created throughout the file.

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
angelathe: Done!

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
angelathe: Done!

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
angelathe: Done!

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -1,6 +1,15 @@
...
+  S3Client,
+  PutObjectCommand,
+  PutObjectCommandOutput,
+} from "@aws-sdk/client-s3";
+import { Bundle } from "fhir/r4";
 
+const S3_SOURCE = "s3";
+const POSTGRES_SOURCE = "postgres";
+const s3Client = new S3Client({ region: process.env.AWS_REGION });
 /**
```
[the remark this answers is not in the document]
angelathe: Ok updated! Removed the example call, but left in the example request body to clarify what the fhirBundle requires (must have ecr id).

PR #1464, opened 2024-03-12
Orchestration sends data to ecr viewer instead of directly to s3 or postgres

### LINKED ISSUE #1332: Update orchestration service to save data via eCR Viewer
## Action Requested
Update the orchestration service so that it saves data through the new endpoint on the ecr-viewer container, rather than directly through the orchestration service.

## Acceptance Criteria
The orchestration service doesn't know anything about databases itself, and instead just calls the ecr viewer's `save fhir data` endpoint.

## Testing
Some integration tests will need to be removed/changed (since the database won't live in the orchestration service anymore, just within the ecr viewer).

## Background & Context
This is part two for #1331. The orchestration service shouldn't know anything about data storage of FHIR bundles, just delegate it to the eCR Viewer.

## Additional Information
We'll also likely need to update the docker-compose and local dev script for the orchestration service to stop referencing the local postgres DB.

KennethSkylight:
# PULL REQUEST

## Summary
ECR Viewer has the endpoint to save fhir bundle data to s3 or postgres. Orchestration is now changed so it sends the bundle data to ecr viewer. 

All boto3 and s3 references are taken out as well as direct postgres inserts

## Related Issue
Fixes #1332


BobanL (approved):
👍

Issue #1473, opened 2024-03-13
For exposure data, ensure the array from an Observation resouce's component field is used for repeating questions PHDC


robertmitchellv:
## Action Requested
For the Generic Repeating Questions section of the PHDC Case Report, we need to ensure that an Observation that has a `category.coding.code` of `"EXPOS"` will not use the `code` for the `<observation>` sections of Generic Repeating Questions, but instead will use the array inside of `"component"`. The array should have a list of both `"code"` and `"valueCodeableConcept"` pairs that will constitute the `<code>` and `<value>` elements within `<entry>`s.

## Acceptance Criteria
When building and running the message parser container locally it parses the message for the generic repeating questions as described above. That all tests downstream of this change are updated and passing both locally and via GitHub Actions.

Issue #1323, opened 2024-02-14
Create a FHIR Bundle that produces a PHDC ready for ingest into NBS


DanPaseltiner:
We want to produce a PHDC that we can hand off to the GDIT team for testing. This PHDC should have all of the required sections and should use code systems that NBS can interpret.

Acceptance Criteria:
GIVEN this FHIR bundle
WHEN it is converted to PHDC
THEN the result PHDC is ready to be successfully ingested into NBS

PR #1485, opened 2024-03-15
Rework repeating questions section, add finalized FHIR bundle for PHDC conversion, update tests and assets

robertmitchellv:
# PULL REQUEST

### Rework repeating questions section:
Making sure that the repeating questions section is only pulling observation based data from the `"component"` array within the Observation resource. Also making sure that an individual Observation resource's `"component"` array is organized under the same `<organizer>` element. 

### Finalized FHIR bundle for PHDC converstion:
Rather than have two FHIR bundles that contain differing information that are leveraged in different parts of the tests, we're now using **only** the FHIR bundle that, when sent to the Message Parser to convert to PHDC, its response passes XML validation.

### Update tests and assets
As mentioned above, now that we have a bundle that can return valid XML, all the tests and assets have been updated to point to these files. This means that the XML in our `message-parser/tests/` tests are testing against valid PHDC XML.

## Related Issue
Fixes:
- #1473 
- #1323 

## Additional Information
This was a painful journey


m-goggins (approved):
🚀 done with XML (for this week)🥳

PR #1489, opened 2024-03-15
update issues template to use proper name for the Streamline ECR team.

BobanL:
# PULL REQUEST

## Summary
Update the issues template to say "Streamline eCR" team.


emmastephenson (approved):
LGTM! Would you mind changing the file name, too? Thanks for catching this.

PR #1546, opened 2024-03-28
Address SQL Injection Issue in Record Linkage `where` clause generation

### LINKED ISSUE #1529: Address SQL Injection Issues in RL

m-goggins:
# PULL REQUEST

## Summary
This PR addresses a SQL injection issue that Eric noticed when testing RL for IDWA. I modified the `_generate_where_clause` method to return both a `where_query` and `query_params` that can be safely passed in to `_generate_block_query` and implemented in `dal.select_results`. I added Eric's test called `test_link_record_with_quote_name_against_mpi_none_record` as well.

## Related Issue
Fixes #1529 

## Additional Information

## Checklist


phdi/linkage/dal.py
```diff
@@ -302,7 +302,7 @@ def bulk_insert_dict(
         return return_results
 
     def select_results(
-        self, select_statement: select, include_col_header: bool = True
+        self, select_statement: select, include_col_header: bool = True, **kwargs
```
ericbuckley, line 305: Given that the other arguments are typed, what do you think about making this one as well?

```python
def select_results(
    self, select_statement: select, include_col_header: bool = True, query_params: dict = None
) -> List[list]:
```

phdi/linkage/mpi.py
```diff
@@ -255,27 +259,34 @@ def _generate_where_criteria(self, block_criteria: dict, table_name: str) -> lis
         :return: A list of where criteria used to append to the end of a query.
```
ericbuckley, line 259: I think the return docstring needs to be updated now that the function signature changed.

phdi/linkage/mpi.py
```diff
@@ -255,27 +259,34 @@ def _generate_where_criteria(self, block_criteria: dict, table_name: str) -> lis
...
+                        f"LEFT({table_name}.{key},4) = :{placeholder}"
                     )
                 elif criteria_transform == "last4":
-                    where_criteria.append(
-                        f"RIGHT({table_name}.{key},4) = '{criteria_value}'"
+                    where_placeholders.append(
+                        f"RIGHT({table_name}.{key},4) = :{placeholder}"
                     )
-        return where_criteria
+            where_params[placeholder] = criteria_value.strip()
```
ericbuckley, line 281: The existing code doesn't call `strip()` on the value, any reason for that change here?

bamader (approved):
Really nice work. Stupid query params and placeholdering everything is annoying but it's definitely the right more. Only question is do we want the `select_results` function in `dal.py` to have an explicit `query_params` in its signature to avoid the `kwargs`? I don't feel strongly about it but we have tried to eliminate kwargs where we can.

phdi/linkage/mpi.py
```diff
@@ -284,27 +295,29 @@ def _generate_block_query(
...
 
             cte_query_table = table_info["table"]
-            query_criteria = self._generate_where_criteria(
+            where_clause, where_params = self._generate_where_clause(
                 table_info["criteria"], table_key
             )
+            new_query_params = {**new_query_params, **where_params}
 
-            if query_criteria is not None and len(query_criteria) > 0:
+            if where_clause is not None and len(where_params) > 0:
```
ericbuckley, line 316: minor thought, I think in the current iteration of `_generate_where_clause`, the second condition will make sure we don't enter this block when the query isn't defined however two thoughts...

1. I think you want to check for the where clause to equal the empty string or just do `if where_clause and ...`, I've I'm reading line 285 correctly, that function will return an empty string for the where_clause when no criteria are defined.
2. In the current implementation of `_generate_where_clause()`, it always has where_params, but will that always be the case?  It might be more adaptable to future changes if you just check for the first condition.  Having a where clause with 0 `where_params` could be an acceptable use case in the future.

ericbuckley (commented):
Changes look great @m-goggins, thanks for fixing this bug.

phdi/linkage/mpi.py
m-goggins: Nope! I was just doing some local testing and forgot to remove it. In theory, data reaching this point should be standardized so I'm going to remove.

phdi/linkage/mpi.py
m-goggins: Good catch!

phdi/linkage/mpi.py
m-goggins: I like your suggestion of changing to check for an empty string and have implemented that change!

I think the where clause will always have parameters because of the conditional nature of SQL where statements, right?

ericbuckley: Yes, very likely and this is really minor so feel free to skip.  It is possible though that the method would add a condition that is not parameterized and the value is hard coded.

DanPaseltiner (approved):
LGTM. Nice work @m-goggins. Thanks for catching this bug and the review here @ericbuckley. 🚀

PR #1552, opened 2024-03-29
fix fhir converter for kentucky data

BobanL:
# PULL REQUEST

## Summary
- fix reference evaluation in fhir converter

## Additional Information
- Previously threw an error for "value" not being a valid key. 
image

### DIFF
--- a/containers/fhir-converter/app/service.py
+++ b/containers/fhir-converter/app/service.py
@@ -18,7 +18,7 @@ def add_data_source_to_bundle(bundle: dict, data_source: str) -> dict:
     :param bundle: The FHIR bundle to add minimum provenance to.
     :param data_source: The data source of the FHIR bundle.
     :return: The FHIR bundle with the a Meta.source entry for each
-      FHIR resource in the bunle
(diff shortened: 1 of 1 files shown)


containers/fhir-converter/app/service.py
```diff
@@ -50,9 +50,9 @@ def resolve_references(input_data: str):
...
     refs = ecr.xpath("//hl7:reference", namespaces=ns)
     for ref in refs:
-        ref_id = ref.attrib["value"][1:]
-        value = " ".join(ecr.xpath("//*[@ID='" + ref_id + "']/text()"))
-        ref.text = value
+        if "value" in ref.attrib:
+            ref_id = ref.attrib["value"][1:]
+            ref.text = " ".join(ecr.xpath("//*[@ID='" + ref_id + "']/text()"))
 
     return etree.tostring(ecr).decode()
```
DanPaseltiner, line 57: Can we properly document this function please? We are currently missing a return type and doc string.

angelathe (approved):
looks good to me!

PR #1553, opened 2024-03-29
Move RL to Container & Remove `phdi` dependency

### LINKED ISSUE #1582: Containerize RL

m-goggins:
# PULL REQUEST

## Summary
This moves all of the record linkage code from the SDK into the container, removing `phdi` as a dependency of the container. 

## Related Issue
Fixes #1582 

## Additional Information

## Checklist


containers/record-linkage/requirements.txt
```diff
@@ -2,9 +2,14 @@ typing_extensions==4.8.0
...
 httpx
 pathlib
 pyway
 psycopg2-binary==2.9.9
-tabulate
\ No newline at end of file
+tabulate
+fhirpathpy
+matplotlib>=3.8.3
+pandas>2.0.0
```
DanPaseltiner, line 12: Are these two actually needed here in the context of only what this service can do?

containers/record-linkage/app/linkage/new_tables.ddl
DanPaseltiner, line 1: Can we consider using the migrations directory that already exists in this service? Seems like a code smell to be defining MPI schema in two different places.

m-goggins: I don't think it can live in `migrations` because of the specific way that pyway interacts with that folder, i.e., expects files named in a certain way, but I moved it into `tests/assets` which is where is probably should have been all along since it's only used in tests.

containers/record-linkage/requirements.txt
m-goggins: Yes, we use pandas for computing log-odds and matplotlib is used in one function to profile the log odds. Replacing pandas is out of scope for this PR but could be another ticket. Off the top of my head I'm not sure what the matplotlib alternatives are but we could also look into replacing it in the future.

containers/record-linkage/app/linkage/new_tables.ddl
DanPaseltiner: Sorry I'm suggesting that we removed `new_tables.ddl` and use the migrations we already have defined in our tests.

containers/record-linkage/requirements.txt
DanPaseltiner: > we use pandas for computing log-odds and matplotlib is used in one function to profile the log odds

Both of these sounds like things that are not exposed via the API. Now that this code is inside the RL container how a potential use it other than via the API? The API does not allow for computing log odds or profiling them, this work needs to be done already and the results provided to this service as configuration.

containers/record-linkage/app/linkage/new_tables.ddl
m-goggins: Oh yep, that's a better idea! Coffee is just kicking in 😆

containers/record-linkage/requirements.txt
m-goggins: I was mistaken; we do use pandas in one function that is exposed via the API. `aggregate_given_names_for_linkage`. Agree about removing Matplotlib and all of the profiling functions that use pandas.

DanPaseltiner: Great. Doesn't need to be part of this PR, but I think we should look at wether this can be done without pandas in the future. Seems pretty heavy weight to pull in such a large dependency to be used in one place.

DanPaseltiner (approved):
Thanks for your work on this!

PR #1535, opened 2024-03-25
Update base python images in all Dockerfiles

nickclyde:
# PULL REQUEST

## Summary
[Snyk reports](https://app.snyk.io/org/prime-phdi/project/90c2de40-a8f1-46e2-adc8-6fc810c24f11) that there are a few fixable vulnerabilities in older python base images, particularly orchestration since it uses 3.11:

Screenshot 2024-03-25 at 3 25 01 PM

Assuming all integration tests pass, we should be good to just bump the base image version.

PR #1587, opened 2024-04-02
Revert "Update base python images in all Dockerfiles (#1535)"

BobanL:
This reverts commit ddf8ecedef8ae98173e53f62948a496e1490fdad.

# PULL REQUEST

## Summary
- Revert to python 3.10
- Set integration test timeout to 30 minutes so we know if this happens again in the future.
- Remove standardization

## Related Issue
Fixes #

## Additional Information

## Checklist


containers/message-parser/dev-requirements.txt
```diff
@@ -1,3 +1,3 @@
 pytest
 rich     
-testcontainers[compose]==3.7.1
+testcontainers[compose]==4.1.1
```
m-goggins, line 3: I think this should still be `3.7.1`...or at least it was on 3/29. For reference, I am looking at the code at [this point in time](https://github.com/CDCgov/phdi/tree/2414772fe0e53e8a76d57fb2dc3c395ab8a13bf0/containers)

containers/fhir-converter/requirements.txt
```diff
@@ -32,12 +32,12 @@ six==1.16.0
 sniffio==1.2.0
 starlette==0.35.0
 syrupy==4.6.1
-testcontainers[compose]==3.7.1
+testcontainers[compose]==4.1.1
```
m-goggins, line 35: I think this should still be 3.7.1...or at least it was on 3/29. For reference, I am looking at the code at [this point in time](https://github.com/CDCgov/phdi/tree/2414772fe0e53e8a76d57fb2dc3c395ab8a13bf0/containers)

containers/validation/requirements.txt
```diff
@@ -6,4 +6,4 @@ uvicorn
 httpx
 pytest
 jsonschema
-testcontainers[compose]==3.7.1
\ No newline at end of file
+testcontainers[compose]==4.1.1
```
m-goggins, line 9: Ditto above

m-goggins (commented):
Thank you for putting this together quickly! I'm not sure about the reversion of the `testcontainers[compose]` but wanted to double check to see if there's a reason to move to 4.1.1

containers/message-parser/dev-requirements.txt
BobanL: I'm not sure why it was failing, but in [previous jobs](https://github.com/CDCgov/phdi/actions/runs/8542647648/job/23405162229#step:6:438) the integration test for FHIR converter and validator would randomly fail. In orchestration, its set to 4.1.1 so I just used the same version here and it seems to have fixed whatever problem it was having.

robertmitchellv:
i'm not able to see the security warning but looking on [docker hub](https://hub.docker.com/_/python/tags?page=1&name=3.10) i'm seeing that there are significant issues with the `debian` versions that could be mitigated by using the `alpine` base instead (based on the vulnerabilities scan). @BobanL would you like me to work through adding that to this PR to see if it resolves our main reason for switching? @DanielPaseltiner what are your thoughts on potentially using `alpine` over `debian`?

DanPaseltiner:
> i'm not able to see the security warning but looking on [docker hub](https://hub.docker.com/_/python/tags?page=1&name=3.10) i'm seeing that there are significant issues with the `debian` versions that could be mitigated by using the `alpine` base instead (based on the vulnerabilities scan). @BobanL would you like me to work through adding that to this PR to see if it resolves our main reason for switching? @DanielPaseltiner what are your thoughts on potentially using `alpine` over `debian`?

I don't think we should do anything in this PR beyond reverting to the exact state when things were last working. As separate, but very important, work we should explore addressing these security issues and migrating to the latest version of Python. It is important stay as current as possible because the longer we let these kind of updates go the harder it gets to upgrade in the future.

containers/message-parser/dev-requirements.txt
m-goggins: Hmm I wonder if there are other differences that we haven't accounted for since it was working a couple of days ago on that version?

robertmitchellv: is the consensus to leave at `4.1.1` or to move it to `3.7.1` at this point? we're all green at this point on tests for this.

containers/fhir-converter/Dockerfile
```diff
@@ -31,8 +31,8 @@ RUN curl https://pyenv.run | bash
 RUN export PYENV_ROOT="$HOME/.pyenv"
 RUN export PATH="$PYENV_ROOT/bin:$PATH"
 RUN eval "$(pyenv init -)"
-RUN /root/.pyenv/bin/pyenv install 3.12.2
-RUN /root/.pyenv/bin/pyenv global 3.12.2
+RUN /root/.pyenv/bin/pyenv install 3.10.13
+RUN /root/.pyenv/bin/pyenv global 3.10.13
```
bamader, line 35: In the github workflows files we're installing 3.10.12; should those be standardized?

containers/ingestion/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.12-slim
+FROM python:3.10-slim
```
bamader, line 1: Confirming we want to use 3.10-slim vs 3.10.12, which has no listed security vulnerabilities on docker hub?

containers/message-parser/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.12-slim
+FROM python:3.10-slim
```
bamader, line 1: ditto

containers/message-parser/dev-requirements.txt
bamader: In the spirit of total reversion I would probably lobby for trying this at 3.7.1. If 3/29 was our last point of full stability I'd say let's get back fully to that before re introducing changes.

containers/orchestration/Dockerfile
```diff
@@ -5,7 +5,7 @@ ARG MESSAGE-PARSER_URL
 ARG SMARTY_AUTH_ID
 ARG SMARTY_AUTH_TOKEN
 
-FROM python:3.12-slim
+FROM python:3.11.5-slim
```
bamader, line 8: Another difference? Feels like we should try to standardize our versions of python across services. This is like the third or fourth different version I've seen.

containers/validation/aks_deployment.md
```diff
@@ -34,7 +34,7 @@ aks-nodepool1-31718369-0   Ready    agent   6m44s   v1.12.8
 In the `Dockerfile`, take note of the port that the app is exposed to. In this example, we will use 8080
 
 ``` 
-FROM python:3.12-slim
+FROM python:3.11-slim
```
bamader, line 37: versioning

bamader (commented):
In the spirit of consistency and intentionality I'd just ask another dev to take a look at the versioning differences I called out in different dockerfiles. Unless there's a compelling reason to have different services use multitudes of different python versions (which there could be for package compatibility), it seems like we should unify those. Python 3.10.12 is stable and has no known vulnerabilities on Docker Hub, at the least.

containers/ingestion/Dockerfile
robertmitchellv: technically there are vulnerabilities for both of these `debian` based images but it doesn't seem like snyk finds them to be worthy of failing the test over. at least based on the workflow here; i still need to learn more about how snyk works

bamader (approved):

robertmitchellv (approved):
in the words of a dear friend of mine:

> :shipit: 
>       - Nick Clyde

PR #1596, opened 2024-04-04
Build new metrics into dashboard

### LINKED ISSUE #1569: Implement a dashboard for requests over time
As an engineer monitoring the DIBBs platform I would like to be able to see requests to the orchestration service over time separated by successes (200 status codes) and failures (4xx and 5xx status codes) over time.

Acceptance Criteria:
GIVEN the Grafana
WHEN a DIBBs engineer goes to this new dashboard
THEN they can view requests to the orchestration service overtime separated by successes and failures.

bamader:
# Add Some Cool New Metrics

## Summary
This PR builds out a prototype MVP dashboard using our open telemetry instrumentation visualized by prometheus and grafana. All changes to the dashboard have been persisted to the mounted storage volume we use for our graph configs, so other than the actual data shown, these layouts will persist for us. Significantly, no new metrics instrumentation code needed to be added to create any of these charts. Outside of the metric tracking requests at the endpoint proper, all the object size and duration metrics  are generated natively by a combination of OpenTelemetry (via automatic instrumentation, so it'll work on any deployment) and PromQL (so on other stacks, the queries that create these would need adjusted for non-Prometheus shops).

## Related Issue
Fixes #1569 

## Screenshot of generated dashboard, if you don't want to start up the service yourself
Screenshot 2024-04-04 at 4 25 15 PM


DanPaseltiner (approved):
This is freaking awesome! 🚀

containers/orchestration/grafana/dashboards/process_message.json
```diff
@@ -91,7 +91,32 @@
...
+          },
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "{exported_job=\"dibbs-orchestration\", instance=\"otel-collector:8889\", job=\"aggregated-otel-collection\", status_code=\"400\"}"
+            },
+            "properties": [
+              {
+                "id": "displayName",
+                "value": "Status Code 400"
```
DanPaseltiner, line 115: Ooops I missed this before. We really want to capture all `4xx` and `5xx` status codes. Is there a way to use some pattern matching, maybe regex, to avoid having to hard code each status code we want to track? As an example a request validation issues will result in a `422` and we would probably want to see those on the tracked as well as just regular old `400` codes.

robertandremitchell (approved):
(meant to post earlier)

looks good! I was able to successfully log in and see the new metrics. sent some test 200/400 data without issue.

containers/orchestration/grafana/dashboards/process_message.json
bamader: Ah, so what you're looking at here is actually just a display name filter. The charts will always automatically retrieve data for all status codes that Orchestration has returned. These rules here are just for renaming the long really gross display string that comes up with a nice short name. I might be able to use a regex to create a rule for all display names though, I'll look into it. But even if I can't, actual data on every status code would already be returned, it would just show up with a longer string with the prometheus job name in the legend!

PR #1597, opened 2024-04-04
Create new "dibbs" package

nickclyde:
# PULL REQUEST

## Summary
This is the first step in our effort to remove the "phdi" package as a dependency from all of our containers. To do that, we have created a new "dibbs" package, which contains only our fastapi BaseService. We also have created a base docker image, based on python:3.10-slim, which installs this package. Finally, to test, we swapped out the "phdi" dependency with the new "dibbs" package in tefca-viewer. We also changed the tefca-viewer's Dockerfile to be based on the new dibbs base image.

A few other minor changes:
- Upgrade all `actions/setup-python` calls to v5
- Remove all `ALERTS_URL`
- Remove all `FHIR_CONNECTOR_URL`, which doesn't seem to be used anywhere

## Related Issue
Fixes:
- #1598


robertmitchellv (approved):
:shipit:

m-goggins (approved):
🚀

.github/workflows/buildContainers.yaml
DanPaseltiner, line 1: 🚀

.github/workflows/createNewRelease.yaml
DanPaseltiner, line 1: Love to see all these actions updates.

containers/dibbs/.gitignore
DanPaseltiner, line 1: Should this live here or in the `.gitignore` at root level? I'm not opposed to this if there is a compelling reason, but want to be careful of the proliferation of `.gitignore` files.

containers/dibbs/Dockerfile
DanPaseltiner, line 1: Could we throw a `pip` update in here to ensure we are always using the most current version?

containers/dibbs/dibbs/base_service.py
```diff
@@ -0,0 +1,107 @@
...
+        "name": "Creative Commons Zero v1.0 Universal",
+        "url": "https://creativecommons.org/publicdomain/zero/1.0/",
+    },
+    "MIT": {"name": "The MIT License", "url": "https://mit-license.org/"},
+}
+
+DIBBS_CONTACT = {
+    "name": "CDC Public Health Data Infrastructure",
+    "url": "https://cdcgov.github.io/phdi-site/",
+    "email": "dmibuildingblocks@cdc.gov",
```
DanPaseltiner, line 22: I think we have a new official email now. I will see if I can track it down in slack and get follow up here.

containers/dibbs/dibbs/base_service.py
```diff
@@ -0,0 +1,107 @@
...
+
+        @self.app.middleware("http")
+        async def rewrite_path(request, call_next):
+            if request.url.path.startswith(self.service_path):
+                request.scope["path"] = request.scope["path"].replace(
+                    self.service_path, ""
+                )
+                if request.scope["path"] == "":
+                    request.scope["path"] = "/"
+            return await call_next(request)
```
DanPaseltiner, line 85: Could we add a doc string and type hints here?

containers/dibbs/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+requires = ["setuptools>=40.6.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "dibbs"
+version = "0.1.0"
+description = "A short description of your project"
```
DanPaseltiner, line 8: Could add our own description? just 1-2 sentences is fine.

containers/dibbs/requirements.txt
```diff
@@ -0,0 +1,4 @@
+pydantic==1.10.9
```
DanPaseltiner, line 1: Why are we locking the pydantic version here? I think if there are specific containers that need a specific pydantic version that should be handled at the individual container level. Ideally the base image is always pushing us to stay current.

containers/tefca-viewer/dev-requirements.txt
```diff
@@ -0,0 +1,2 @@
+pytest
+httpx
```
DanPaseltiner, line 2: Is `httpx` truly only needed for dev? I though `pytest` somehow relied on it. Maybe I'm wrong or not remembering something correctly.

DanPaseltiner (changes requested):
Thank you @robertmitchellv and @nickclyde for putting this together so quickly! 🚀 😲 I have left some comments on mostly relatively minor things I think we should consider. Before merging I'd like us to think through our release process. I see the advantages of building one common DIBBs base image, but there are some challenges. I think a shared understanding of what the future intended release process will look like should help us think through the relative pros and cons here.

containers/dibbs/dibbs/base_service.py
DanPaseltiner: Found it!

```suggestion
    "email": "dibbs@cdc.gov",
```

containers/dibbs/.gitignore
nickclyde: Yeah good call, I'll move to the root `.gitignore` 👍🏻

containers/dibbs/requirements.txt
nickclyde: All of the containers are currently locked to this version. I think there was a specific reason for that, but I don't remember what it was. I agree we should try to update, but maybe we should save it for another PR?

containers/tefca-viewer/dev-requirements.txt
nickclyde: Yeah it's needed for some tests, but pytest is also only needed for dev! No need to bundle it into our containers

DanPaseltiner (approved):

PR #1607, opened 2024-04-05
Active problems with no display name should be excluded

### LINKED ISSUE #1606: Active Problem is showing a row when there is no active problem on the ecr
http://localhost:3000/view-data?id=1.2.840.114350.1.13.297.3.7.8.688883.567479

lina-roth:
# PULL REQUEST

## Summary
- If there is no display name on an active problem, remove it from the list. 
- If there are no remaining active problems after that, return undefined
- Add test for this specific case

http://localhost:3000/view-data?id=1.2.840.114350.1.13.297.3.7.8.688883.567479

### Pre Change ECR Viewer
Screen Shot 2024-04-05 at 2 12 13 PM

### Related issue ecr html
Screen Shot 2024-04-05 at 3 16 49 PM

### After change
Screen Shot 2024-04-05 at 4 07 59 PM

## Related Issue
Fixes #1606 

## Additional Information

## Checklist


containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -479,12 +479,20 @@ export const returnProblemsTable = (
...
-    };
+    if (entry.code?.coding?.[0].display !== undefined) {
+      entry.onsetDateTime = formatDate(entry.onsetDateTime);
+      entry.onsetAge = {
+        value: calculatePatientAge(fhirBundle, mappings, entry.onsetDateTime),
+      };
+    } else {
+      problemsArray.splice(problemsArray.indexOf(entry), 1);
+    }
   });
```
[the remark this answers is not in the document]
lina-roth: @BobanL So are you saying that I should just filter the array right away and then check if the array is empty and if so return undefined. otherwise proceed with lines 483 - 485

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -510,6 +517,10 @@ export const returnProblemsTable = (
     };
   });
 
+  if (problemsArray.length === 0) {
+    return undefined;
+  }
+
```
BobanL, line 523: I think this is a duplicate!

lina-roth: @BobanL 

**My thinking was -** 

**If I don't check to see if it empty first,** 
```
 if (problemsArray.length === 0) {
    return undefined;
  }

```
**this might throw an NPE.** 

```
  problemsArray = problemsArray.filter(
    (entry) => entry.code?.coding?.[0].display,
  );
```

I guess I could combine the checks into some sort of while loop

lina-roth: But I guess with `?.` it is null safe, right?

lina-roth: I just tested it. You are right, we only need to do the check once. Updating now

BobanL (approved):

PR #1609, opened 2024-04-05
Update Record Linkage to use the new dibbs package

### LINKED ISSUE #1608: Update Record Linkage to use the new dibbs package / base container
Update Record Linkage to use the new dibbs package / base container introduced in #1597

nickclyde:
# PULL REQUEST

## Summary
Update Record Linkage to use the new dibbs package introduced by #1597  (followup to #1553)

## Related Issue
Fixes #1608

### DIFF
--- a/containers/record-linkage/app/utils.py
+++ b/containers/record-linkage/app/utils.py
@@ -1,16 +1,9 @@
 import json
 import logging
 import pathlib
-import random
 import subprocess
-from functools import cache
-from typing import Any
-from typing import Callable
-from typing import List
 from typing import Literal
-from typing import Union
 
-import fhirpathpy
 from app.config import get_settings
 from app.linkage.dal import DataAccessLayer
 from sqlalchemy import text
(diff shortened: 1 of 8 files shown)


containers/record-linkage/app/utils.py
```diff
@@ -122,38 +119,6 @@ def run_migrations():
...
-# TODO: Move this to the dibbs SDK once created
-class StatusResponse(BaseModel):
-    """
-    The schema for the response from the health check endpoint.
-    """
-
-    status: Literal["OK"]
-
-
 selection_criteria_types = Literal["first", "last", "random", "all"]
```
m-goggins, line 122: Do you want to move the other functions originally from `phdi` (the ones with `TODO`s) into `dibbs` as well?

nickclyde: Are they used by any of the other services? If so - we should move them into dibbs, otherwise they should stay in RL only!

m-goggins: Yeah, they are used by other services and are originally from `phdi/fhir/utils.py`. I added comments/TODOs for each of the functions in this file that I thought should end up in `dibbs` and where they came from in `phdi`.

m-goggins (approved):
Thanks for moving these!

PR #1632, opened 2024-04-09
Centralize testing utils

bamader:
# Making our RL tests (at least a little bit) less brittle

## Summary
This PR introduces a couple of refactoring cleanup changes designed to make the record linkage tests more robust in the face of our DB and CI/CD woes. There are a couple of main changes here:
- removing duplicate copies of `_clean_up`, `set_env_variables`, and `pop_env_variables`; previously, there were three different versions of most of these functions with either minor variations or no variations at all. These have now all been centrally located in the `app.utils` file for sharing across all tests. In particular, with the `_clean_up` function, we've now made it more flexible for testing purposes because the `dal` parameter is optional.
- wrapping migrations, DB setup, and env variable teardown in the testing suite in an `autouse` `pytest.fixture`. Instead of manually defining and calling these functions with every test, we can create an automation point that tells pytest to run setup and teardown functions for us around every test in a scoped file. Importantly, this gives us the benefit of a `try/finally` flow in that pytest is now responsible for both creating and removing the environment variables used in the testing suite (i.e. if we encounter some issue down the line with the pyway migrations command, all tables should still get removed from the testing MPI so we better achieve idempotency).

## Related Issue
No zenhub issue, but makes some strides on cleaning up some of the most brittle testing offenders, and I said I'd make a fast follow to address the duplications that Robert found

## Additional Information
There is one more area of duplication that I didn't touch here, the doubling up config sets in `app/config.py` and `app/linkage/utils.py`. While I'm 80ish percent sure we could delete either one of these and wire all config through the other, I know cache performance on the DB had a massive performance impact when seeding in LAC's environment, so I'm hesitant to mess with anything that involves LRU cache settings since I don't know which part of the cache being defined was part of that speed up.

### DIFF
--- a/containers/record-linkage/app/linkage/mpi.py
+++ b/containers/record-linkage/app/linkage/mpi.py
@@ -5,7 +5,7 @@ import uuid
 from functools import cache
 from typing import Dict
 from typing import List
-from typing import Literal
+from typing import Union
 
 from app.linkage.core import BaseMPIConnectorClient
 from app.linkage.dal import DataAccessLayer
@@ -603,7 +603,7 @@ class DIBBsMPIConnectorClient(BaseMPIConnectorClient):
                 )
 
     @cache
-    def _get_external_source_id(self, external_source_name: str) -> Literal[str, None]:
+    def _get_external_source_id(self, external_source_name: str) -> Union[str, None]:
         """
         Gets the external source id for the external source name provided.
         :param external_source_name: The external source name.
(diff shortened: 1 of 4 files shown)


robertmitchellv (approved):
thanks for working through how we thread between the duplicate environment variable setting, config setting, and tear down setting. this is a lot cleaner.

i also built these containers locally: ran `docker compose up --build -d` and all the tests passed locally as well; `200` on health check via insomnia as well

🚀

PR #1633, opened 2024-04-09
Pydoc linting and new pydocs

nickbristow:
# PULL REQUEST

## Summary
Turns on pydoc linting in ruff and adds new pydocs where needed

## Related Issue
Fixes #

## Additional Information

## Checklist


m-goggins:
Are you wanting to change all of the docstrings to this format? All of the functions with existing docstrings are formatted differently

bamader:
We actually have the format we do so that Sphinx can auto-generate and auto-populate our documentation and user guides--is this format compatible with Sphinx / how will it interact with auto-generation? Since we made a deliberate choice for our existing documentation format I'm hesitant to replace it.

robertandremitchell (commented):
same note as others, that the rest of the docstrings in `phdi` folllow the format for the purposes of auto-formatting:

"""
Description
:param variable1: description
:param variable2: description

optional:
:return: description
"""

DanPaseltiner:
I agree we should think through if we want to change the doc string format, but I'd like to raise a couple points.

1. We are actually using `pdoc` for docs on the SDK not Sphinx, as you can see [here](https://cdcgov.github.io/phdi/latest/sdk/phdi.html)
2. Long term we are planning to deprecate `phdi` so we will not need to worry about publishing to pypi for hosting nice docs.
3. Changing our doc string format is a pain, but if this format allows for easy linting I think the change may be worth it.

Perhaps we should discuss at eng sync?

nickbristow:
@m-goggins @bamader @robertandremitchell @DanPaseltiner 
we aren't trying to change the format. That is just what we found. 

Currently the linter only checks that they exist as noted by [this guy](https://github.com/astral-sh/ruff/issues/6606#issuecomment-1949348108). We can update the new ones we added to be spinx style.

m-goggins (approved):
Nice! Thank you for changing the format to match our existing docstrings.

PR #1635, opened 2024-04-10
Remove redundant Python Linting

JNygaard-Skylight:
# PULL REQUEST

## Summary
Remove Python linting and formatting as a step in the pipeline. This already happens in `pre-commit` and that also gets ran as part of the pipeline.

## Related Issue
Fixes #

## Additional Information

## Checklist

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -16,30 +16,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-env:
-  TEST_RUNNER_PYTHON_VERSION: 3.10.12
-
 jobs:
-  python:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Setup python ${{env.TEST_RUNNER_PYTHON_VERSION}}
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{env.TEST_RUNNER_PYTHON_VERSION}}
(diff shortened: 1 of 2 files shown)


DanPaseltiner (changes requested):
Since it is possible to push changes without running pre-commit this would effectively allow for bypassing linting which I don't think is what we want. What is the problem we are trying to solve here?

JNygaard-Skylight:
> Since it is possible to push changes without running pre-commit this would effectively allow for bypassing linting which I don't think is what we want. What is the problem we are trying to solve here?

Sure you can skip pre-commit when you push changes, but doesn't pre-commit get ran as a a pipeline action, and isn't it required?

No problems, just thought I could remove a few lines of code.

DanPaseltiner:
> > Since it is possible to push changes without running pre-commit this would effectively allow for bypassing linting which I don't think is what we want. What is the problem we are trying to solve here?
> 
> Sure you can skip pre-commit when you push changes, but doesn't pre-commit get ran as a a pipeline action, and isn't it required?
> 
> No problems, just thought I could remove a few lines of code.

I completely forgot that the pre-commit runs automatically when changes are merged into a PR. Trying to confirm that pre-commit passing is required to merge. Basically I just want to ensure that the code has to pass linting before being allowed to merge.

DanPaseltiner:
So it doesn't appear to me that a successful run of pre-commit is required for merging. Am I missing something in the configs for settings for the repo? I guess the case I'm concerned about is when `ruff` fails and can't correct the issue.

PR #1651, opened 2024-04-15
record linkage performance enhancments

ericbuckley:
# PULL REQUEST

## Summary
3 optimizations to the record linkage SDK that, collectively, has the potential for a 2x performance boost.
- fix 1: remove schema initialize calls from `dal::get_table_by_column_*` functions
- fix 2: remove deep copy call when retrieving foreign key info from the block query
- fix 3: aggregate given names in postgres

### Testing Process
A testing framework was developed in the [IDWA repo](https://github.com/CDCgov/IDWA/tree/main/dedupe/record_linkage_testing) for running repeatable performance tests.  Please take a look at the code, along with some accompanying [videos on finding the optimizations](https://loom.com/share/folder/b582449e01184047b519b9efc938bfa4) for more context.

### Notes on Fix 1
The MPI is the only module that uses instances of the DataAccessLayer (dal.py).  During the MPI initialization process, it invokes the [initizlize_schema function](https://github.com/CDCgov/phdi/pull/1651/files) almost immediately after instantiating a DAL object.  Thus, it isn't necessary to initiate it every time a [block query is built](https://github.com/CDCgov/phdi/pull/1651/files).  In my tests, there were some outlier requests that required as much as ~100ms to create the SQLAlchemy objects created in the initialize_schema function.

### Notes on Fix 2
It appears that the [MPI _generate_block_query function](https://github.com/CDCgov/phdi/blob/main/phdi/linkage/mpi.py#L325-L328) is making a copy of the CTE query table, just so we can pop off the first foreign key column and use that info to augment the query.  The CTE query table is an SQLAlchemy object, and while in most cases it can quickly create a copy, in my tests there were some outlier requests that required as much as ~100ms to make the copy.  Since we only need to reference the data for its values, we can just retrieve it off the original query and not copy it first.

### Notes on Fix 3
Using [pandas to aggregate the given names](https://github.com/CDCgov/phdi/blob/main/phdi/linkage/link.py#L1565) into a first name adds overhead, particularly in initializing the DataFrame and running the in-memory query operations on it.  Since we are already running a postgres query to retrieve the results from the datastore, we can save ourselves a step and just aggregate directly in postgres.  Performing the aggregation on postgres adds a negligible amount of latency, as while the query is every so slightly longer to compute, the amount of data transferred back to the API service is less, thus it doesn't appear to take longer.

While the previous `link::aggregate_give_names_for_linkage` was removed because the aggregation has been added to the MPI block query, a new link.py method was still necessary to convert the array of given names into a space delimited first name value that link.py needs.

### Other
- a minor change was made to core.py to get its abstract _generate_block_query signature to match that of the MPI implementation.  This isn't a bug, nor did it have an effect on performance, this is just a recommended change by python linters.

## Checklist


containers/record-linkage/app/linkage/mpi.py
```diff
@@ -308,8 +309,7 @@ def _generate_block_query(
                         .cte(f"{table_key}_cte")
                     )
                 else:
-                    fk_query_table = copy.deepcopy(cte_query_table)
-                    fk_info = fk_query_table.foreign_keys.pop()
+                    fk_info = next(iter(cte_query_table.foreign_keys))
```
DanPaseltiner, line 312: Nice!

containers/record-linkage/app/linkage/mpi.py
```diff
@@ -457,6 +461,19 @@ def _get_base_query(self) -> Select:
...
+                "birthdate",
+                self.dal.PATIENT_TABLE.c.sex,
+                id_sub_query.c.mrn,
+                self.dal.NAME_TABLE.c.last_name,
+                self.dal.NAME_TABLE.c.name_id,
+                "address",
+                "zip",
+                self.dal.ADDRESS_TABLE.c.city,
+                self.dal.ADDRESS_TABLE.c.state,
+            )
```
DanPaseltiner, line 476: Since we are introducing a `GROUP BY` on the base query that is run 2x per linkage request I'm concerned about how this will scale when the DB has a large number of records in it e.g. > 1M. In order for this to scale acceptably I _think_ we need this aggregation to be applied after the joins on the CTEs for blocking conditions so that we are aggregating after filtering down to just the set of records that belong in a block. It's possible the Postgres query optimizer is smart enough to be doing this already, but I think we might want to do some profiling to be sure. The CTEs I'm referring to are created in `_generate_block_query` and joined onto this base query on line 330. It's possible I'm not thinking something through here correctly and in general I like the idea of making the DB wrangle the given names for us.

DanPaseltiner (changes requested):
This is really great @ericbuckley! I'm blocking merging because I'd like to make sure we think through the potential implications of this change to the base query so that we don't trade one performance issue for another (see comment below). However in principle I like the approach. Thank you for taking the time not only to do this performance testing, but also to code up solutions and open this PR!

containers/record-linkage/app/linkage/mpi.py
ericbuckley: The tests I ran and shared in the loom videos, we're against a seeded database.  Far from 1M records, but it had about ~10,000 person records and ~37,000 patient records (so possibly just 1% of what you are talking about).

Let me rerun the tests and get the explain/analyze output for both the before and after of the block queries.  If you want, I can also increase the seed size to something much larger.

DanPaseltiner: If we could take a look a the explain/analyze output that's probably sufficient. I didn't realize you had loaded the MPI with that many records. This is probably totally fine. We had a significant performance issue a few months ago related to any aggregation being applied improperly so I might be being overly cautious here. Thanks!

ericbuckley: @DanPaseltiner I dug into this a bit deeper and have 3 takeaways

1) There is a bug in my PR, in that I need to add a `DISTINCT` clause to the CTE queries, as we're no longer using pandas to [remove duplicates from the blocking query](https://github.com/CDCgov/phdi/blob/main/phdi/linkage/link.py#L1595).  I'll push a commit shortly to address that.

2) As you'll see in the EXPLAIN diff below (most of the explain was removed, so I can highlight the differences), the GROUP BY clause adds little overhead to the query, as by the time it runs we've already reduced the set of rows to a small amount.

**optimized explain**
```
GroupAggregate  (cost=3251.51..3251.68 rows=5 width=165)
  Group Key: patient.patient_id, identifier.patient_identifier, name.name_id, address.line_1, address.zip_code, address.city, address.state
  ->  Sort  (cost=3251.51..3251.52 rows=5 width=147)
        Sort Key: patient.patient_id, identifier.patient_identifier, name.name_id, address.line_1, address.zip_code, address.city, address.state
        ->  Nested Loop Left Join  (cost=2302.98..3251.45 rows=5 width=147)
            ...
```
**original explain**
```
Nested Loop Left Join  (cost=2302.98..3251.45 rows=5 width=147)
...
```

3) The Hash Join and Seq Scan (full explain analyze of the original query below) on the patient table seems costly.  Now it could be that because there is so little data in the patient table (<50k rows), postgres just decides to look at everything, but that could be a problem in production.  I think it would be a good idea to scale the database seed up to something like ~100k person records and ~400k patient records, and rerun some of this analysis to see if the sequential scan is still being used on large tables.
```
Nested Loop Left Join  (cost=2302.98..3251.45 rows=5 width=147) (actual time=19.891..19.977 rows=4 loops=1)
  ->  Nested Loop Left Join  (cost=2302.27..3249.51 rows=2 width=107) (actual time=19.853..19.930 rows=1 loops=1)
        ->  Nested Loop Left Join  (cost=2301.98..3248.69 rows=2 width=64) (actual time=19.832..19.908 rows=1 loops=1)
              ->  Hash Join  (cost=2301.56..3246.90 rows=2 width=42) (actual time=19.755..19.829 rows=1 loops=1)
                    Hash Cond: (patient.patient_id = name_2.patient_id)
                    ->  Hash Join  (cost=434.99..1379.67 rows=249 width=58) (actual time=0.878..10.285 rows=279 loops=1)
                          Hash Cond: (patient.patient_id = name_1.patient_id)
                          ->  Seq Scan on patient  (cost=0.00..846.60 rows=37360 width=42) (actual time=0.059..4.441 rows=37360 loops=1)
                          ->  Hash  (cost=431.88..431.88 rows=249 width=16) (actual time=0.311..0.314 rows=279 loops=1)
                                Buckets: 1024  Batches: 1  Memory Usage: 22kB
                                ->  HashAggregate  (cost=426.90..429.39 rows=249 width=16) (actual time=0.211..0.247 rows=279 loops=1)
                                      Group Key: name_1.patient_id
                                      Batches: 1  Memory Usage: 45kB
                                      ->  Bitmap Heap Scan on name name_1  (cost=6.22..426.27 rows=249 width=16) (actual time=0.042..0.135 rows=279 loops=1)
                                            Recheck Cond: ("left"((last_name)::text, 4) = 'Scha'::text)
                                            Heap Blocks: exact=66
                                            ->  Bitmap Index Scan on name_last_name_index  (cost=0.00..6.16 rows=249 width=0) (actual time=0.033..0.033 rows=279 loops=1)
                                                  Index Cond: ("left"((last_name)::text, 4) = 'Scha'::text)
                    ->  Hash  (cost=1862.53..1862.53 rows=323 width=16) (actual time=9.473..9.476 rows=254 loops=1)
                          Buckets: 1024  Batches: 1  Memory Usage: 20kB
                          ->  Unique  (cost=1857.69..1859.30 rows=323 width=16) (actual time=9.254..9.383 rows=254 loops=1)
                                ->  Sort  (cost=1857.69..1858.50 rows=323 width=16) (actual time=9.252..9.277 rows=308 loops=1)
                                      Sort Key: name_2.patient_id
                                      Sort Method: quicksort  Memory: 39kB
                                      ->  Hash Join  (cost=629.75..1844.23 rows=323 width=16) (actual time=0.299..9.138 rows=308 loops=1)
                                            Hash Cond: (name_2.name_id = given_name_1.name_id)
                                            ->  Seq Scan on name name_2  (cost=0.00..971.50 rows=47950 width=32) (actual time=0.007..3.729 rows=47918 loops=1)
                                            ->  Hash  (cost=625.71..625.71 rows=323 width=16) (actual time=0.164..0.165 rows=308 loops=1)
                                                  Buckets: 1024  Batches: 1  Memory Usage: 23kB
                                                  ->  Bitmap Heap Scan on given_name given_name_1  (cost=6.80..625.71 rows=323 width=16) (actual time=0.030..0.098 rows=308 loops=1)
                                                        Recheck Cond: ("left"((given_name)::text, 4) = 'Bern'::text)
                                                        Heap Blocks: exact=30
                                                        ->  Bitmap Index Scan on given_name_given_name_index  (cost=0.00..6.71 rows=323 width=0) (actual time=0.025..0.025 rows=308 loops=1)
                                                              Index Cond: ("left"((given_name)::text, 4) = 'Bern'::text)
              ->  Index Scan using identifier_patient_id_index on identifier  (cost=0.42..0.89 rows=1 width=38) (actual time=0.072..0.074 rows=1 loops=1)
                    Index Cond: (patient_id = patient.patient_id)
                    Filter: ((type_code)::text = 'MR'::text)
                    Rows Removed by Filter: 4
        ->  Index Scan using address_patient_id_index on address  (cost=0.29..0.40 rows=1 width=59) (actual time=0.017..0.018 rows=1 loops=1)
              Index Cond: (patient_id = patient.patient_id)
  ->  Nested Loop Left Join  (cost=0.71..0.95 rows=2 width=56) (actual time=0.035..0.042 rows=4 loops=1)
        ->  Index Scan using name_patient_id_index on name  (cost=0.29..0.38 rows=1 width=42) (actual time=0.018..0.019 rows=2 loops=1)
              Index Cond: (patient_id = patient.patient_id)
        ->  Index Scan using given_name_name_id_index on given_name  (cost=0.42..0.55 rows=2 width=30) (actual time=0.009..0.009 rows=2 loops=2)
              Index Cond: (name_id = name.name_id)
Planning Time: 2.188 ms
Execution Time: 20.384 ms
```

Here is a loom video with more details of the analysis I did this evening.
https://www.loom.com/share/ac637562f28742ae8133f6efb0b3cfe0?sid=4879d391-3bef-450b-ae66-49197c5ff5db

bamader (approved):
Absolutely wonderful, Eric. This is so awesome to see that we're making further performance optimizations on linkage. I agree with your call to see if we can scale up the DB to see if that sequential hash and join is still being called, but what you have here now that the `DISTINCT` is in there and I can look at the performance results is all excellent. I'd say once we can check out that scale up and see if that addresses Dan's concern (which @DanPaseltiner I don't think is overly cautious, this is a very fragile beast of a system and one where the queries have massively swung performance in the past, so I wouldn't want to overcommit too quickly on a change either), we'll be good to go.

containers/record-linkage/app/linkage/mpi.py
ericbuckley: I ran some tests today with a much larger database (100,000 person rows and 367,000 patient rows).  The API timings performed very similar to a database at 1/10th that size (avg 100ms for current RL API and 50ms for the version proposed in this PR).

@bamader @DanPaseltiner do either of you know if the latency issues LAC observed were on all API requests, or was it pertaining to a subset of those requests that had large blocking query matches (eg trying to match on John Doe)?

DanPaseltiner: Thank you for doing this additional testing @ericbuckley. Looks good to me and I think we should merge. Eliminating pandas is another task that has been on the back burner for a while so that's another value add for this PR.

To answer your question, the issue we are referring to was related to each blocking query so every request was hit by it twice regardless of the size of the blocks. IIRC for some reason we were running a `GROUP BY` on the entire patient table for every block before applying the CTEs to filter down to a limited number of records. Obviously, this became very expensive as data was added to the MPI. We rewrote the query to eliminate the unnecessary aggregation and performance improved dramatically. It was really a mistake on our part in first place, but that experience has made me cautious of introducing aggregations on the patient table. The testing and profiling you have done demonstrates that we should be totally fine in this regard. I just wanted us to confirm before moving ahead given this past experience. Thank you again for your work here.

DanPaseltiner (approved):
🚀

DanPaseltiner:
@ericbuckley are we good to merge here?

ericbuckley:
> @ericbuckley are we good to merge here?

yep, I'll merge in just a minute.

PR #1654, opened 2024-04-16
Adminstered medications

### LINKED ISSUE #1107: Add Administered Medications to eCR Viewer
## Action Requested
Add the "Administered Medications" field to the eCR Viewer.
>eCR Viewer
>> Clinical Info
>>> Treatment Details
>>>> Administered Medications

## Acceptance Criteria
Administered Medications information is displayed for eCRs that have the field.

## Testing
Unit testing will be required.

## Background & Context
This field may be a list of sub-fields.

## Designs
[Figma Designs](https://www.figma.com/file/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?type=design&node-id=1024%3A136062&mode=design&t=lClwL7CnvrviunUF-1)
[Figma Clickable Prototype](https://www.figma.com/proto/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?page-id=871%3A2283&type=design&node-id=1606-27619&viewport=1011%2C558%2C0.05&t=sGUJKIN0MXVFYIfm-1&scaling=min-zoom&starting-point-node-id=1606%3A27619&mode=design)

## Additional Information
We'll need to test to see whether or not this field is already being converted to FHIR, or if we'll need to complete both FHIR conversion and adding to the eCR Viewer.

Administered Medications may be a list, so if it seems difficult to incorporate all available information into the design, please chat with Sarah.

KennethSkylight:
# PULL REQUEST

## Summary
Adds the Administered Medications Table as part of the Treatment Plan section under Clinical Info. It largely follows a similar pattern to procedures. It also assumes that the table has medication name and medication start date in the table, which most do. 

## Related Issue
Fixes #1107 

## Screenshot
image

## How to test

`cd containers/ecr-viewer`
`npm run local-dev`
go to `http://localhost:3000/view-data?id=df33f075-83a1-4d28-ba63-05bcc3339c21`
Observe the administered medications table in the clinical info section


containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -587,6 +587,56 @@ export const returnPendingResultsTable = (
   }
 };
 
+export const returnAdminMedTable = (
```
nickbristow, line 590: Don't we already have generic create table jsx functions?

containers/ecr-viewer/src/app/utils.tsx
```diff
@@ -779,12 +829,13 @@ export const evaluateClinicalData = (
...
-      <>
-        <div className={"data-title margin-bottom-1"}>Plan of Treatment</div>
-        {pendingResults}
-      </>
-    );
+    planOfTreatmentElement = <>{pendingResults}</>;
+  }
+
+  const adminMedResults = returnAdminMedTable(fhirBundle, mappings);
+  let adminMedElement: React.JSX.Element | undefined = undefined;
```
nickbristow, line 836: you could write this as a big ternary?
```let adminMedElement: React.JSX.Element | undefined = adminMedResults ? (<>{adminMedResults</>) : undefined;```

nickbristow (changes requested):

containers/ecr-viewer/src/app/utils.tsx
KennethSkylight: Done

containers/ecr-viewer/src/app/utils.tsx
KennethSkylight: It's different because the data is coming in as a string of html rather than an actual FHIR Resource. The generic one uses defined FHIR resources. The reason why AdministeredMedications does not use the FHIR definition is because it is returned as a string and the formats can be quite different

nickbristow (approved):

KennethSkylight:
@sarahtress I accidentally merged this before design review. But any comments you have here I can implement

sarahtress:
> @sarahtress I accidentally merged this before design review. But any comments you have here I can implement

no worries! this looks good to me anyways

PR #1656, opened 2024-04-16
Implement `/refine` endpoint

### LINKED ISSUE #1561: implement /refine
Now the that we have a function that can produced a reduced eICR given config let's implement `/minify-ecr` and put this new function to use.

Acceptance Criteria:

GIVEN an eICR
WHEN the eICR is POSTed to `/minify-ecr` with appropriate config
THEN a reduced version of the eICR is returned to the client

m-goggins:
# PULL REQUEST

## Summary
In an effort to work in vertical slices, this PR implements a placeholder for `/refine` that currently returns a given eCR but will eventually return a refined eCR once the function to refine eCRs is completed.

## Related Issue
Fixes #1561 

## Additional Information

## Checklist


containers/refiner/app/models.py
```diff
@@ -0,0 +1,31 @@
...
+    The schema for requests to the /refine endpoint.
+    """
+
+    include_headers: Optional[bool] = Field(
+        description="Include headers in the output."
+    )
+    fields_to_include: Optional[List[str]] = Field(
+        description="Fields to include in the refinement."
+    )
+    message: str = Field(description="The XML to refine.")
```
DanPaseltiner, line 19: I think we should probably not have a JSON formatted request body here as this will make things more complicated to implement on the client side in Rhapsody. Could we have an XML body which is just the eICR and move `include_headers` and `fields_to_include` to be query parameters? This would mean that the URL for a request might look something like `/refine-ecr?include_headers=true&fields_to_inlcude=my_field1,my_filed2` and the body of the request would simply be the XML document. We should also require the `Content-Type application/xml` header, which could be done in a separate PR.

containers/dibbs/dibbs/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Reads a file from the assets directory.
+
+    :param filename: The name of the file to read.
+    :return: A string containing the contents of the file.
+    """
+    with open(
+        (pathlib.Path(__file__).parent.parent / "assets" / filename), "r"
+    ) as file:
+        return file.read()
```
DanPaseltiner, line 25: These are probably good candidates to move to `dibbs` so we don't repeat them in every service. We can address in another PR.

containers/refiner/app/models.py
```diff
@@ -0,0 +1,31 @@
...
+class RefinerResponse(BaseModel):
+    """
+    The schema for responses from the /refine endpoint.
+    """
+
+    status: str = Field(
+        description="A message describing the result of a request to "
+        "the /refine endpoint."
+    )
+    refined_message: str = Field(description="The refined XML.")
```
DanPaseltiner, line 31: I think the response can just be the refined XML document. There isn't really a need to include the status code as it will be accessible to the client in the response from the service separate from the body. In fairness I think we (probably me) started including status codes in body responses early on when I/we didn't have as good an understanding of building these kind of endpoints.

containers/refiner/app/main.py
```diff
@@ -19,3 +22,18 @@ async def health_check():
     properly.
     """
     return {"status": "OK"}
+
+
+@app.post("/refine-ecr")
```
DanPaseltiner, line 27: What do you think about just making this `/refine` and calling the service `ecr-refiner` then requests would go to `ecr-refiner/refine`.

m-goggins: I had thought about that but I was wondering if we wanted to structure it like the message parser that can parse a variety of message types. I guess an eCR wouldn't have different sections than any other FHIR bundle/XML and we could just allow different sections to include in the query params like you suggested above. It is more intuitive to me if the service is generic and the endpoint is more specific, but I don't feel too strongly and am okay to change if you feel more strongly about it.

containers/refiner/app/models.py
m-goggins: Ah I had forgotten that we had talked about that during backlog refinement. I'll make those changes!

containers/refiner/app/main.py
DanPaseltiner: That's a good point and there could very well be format specific logic that is required if we need to support  XML, HL7v2, and FHIR/JSON. Then what if we make this the `/ecr` endpoint so the full path would be `refiner/ecr`? I think the redundancy in `refiner/refine-ecr` is throwing me off, maybe just a me thing though and this is certainly not a hill I want to die on.

If we are setting ourselves up for a more general model what do you think about calling the service the "message refiner"? My concern with just "refiner" is that out of context it is hard to have any idea of what the service does.

m-goggins: I like `message-refiner/ecr`!

robertandremitchell (commented):
looks good!

containers/refiner/app/models.py
```diff
@@ -0,0 +1,18 @@
...
+
+    message: str = Field(description="The XML to refine.")
+
+
+class RefinerResponse(BaseModel):
+    """
+    The schema for responses from the /refine endpoint.
+    """
+
+    refined_message: str = Field(description="The refined XML.")
```
DanPaseltiner, line 18: Again, I think the response can just be the XML

DanPaseltiner (approved):
Thanks for working through the API design with me. 🚀

Issue #1684, opened 2024-04-24
Write function to add headers to refined eCRs


m-goggins:
**Background**: The refined messages do not include headers, such as `realmCode`

**Acceptance Criteria**:

- [ ] all headers are included in the refined messages (up to the first `component`)
- [ ] test assets are updated to include headers

PR #1691, opened 2024-04-25
Add function to refine eCR messages

### LINKED ISSUE #1555: Write a function to minify eCRs
Following the work in the [this ticket](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/1534) let's create a function that when given an eICR and a config returns a reduced eICR.

Acceptance Criteria

GIVEN an eICR and a reduction config
WHEN this new function is called
THEN a reduced eICR is returned containing only the sections specified by the config

m-goggins:
# PULL REQUEST

## Summary
Given an optional string of sections to include (e.g., "social history, labs", we can now use the `refine` function to return an eCR with only the specified sections. The `refine` function uses XPaths to decide which sections of the eCR should be included. 

There is still work to do to ensure that the "headers" of an eCR are included in the output (e.g., the assigned author section etc.) but I created a separate ticket for that work (#1684).

## Related Issue
Fixes #1555 

## Additional Information
There might be a better way to store the test assets as a single file and to deal with the namespaces but I wanted to get feedback on the `refine` function itself rather than let those details hold up the PR. Once this PR has merged, it will be much easier for multiple people to work on the refiner work simultaneously.

## Checklist


containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,92 @@ async def refine_ecr(
...
+        return raw_message
+    else:
+        raw_message = ET.fromstring(raw_message)
+
+        # Validate sections to include
+        sections = validate_sections_to_include(sections_to_include)
+
+        # Set up XPath expression
+        namespaces = {"hl7": "urn:hl7-org:v3"}
+        sections_xpath_expression = "or".join(
```
DanPaseltiner, line 105: This `or` is pretty slick!

containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,92 @@ async def refine_ecr(
...
+    :return: The refined message.
+    """
+
+    if sections_to_include is None:
+        return raw_message
+    else:
+        raw_message = ET.fromstring(raw_message)
+
+        # Validate sections to include
+        sections = validate_sections_to_include(sections_to_include)
```
DanPaseltiner, line 101: I wonder if the call to `validate_sections_to_include` should not be nested within `refine`, but instead called in `refine_ecr` before `refine`. First, less nesting is ofter easier to reason about, but more importantly by validating the sections early and independently we have an opportunity to return a `422` response to the client if they have given us sections we don't know how to handle.

containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,92 @@ async def refine_ecr(
...
+def refine(raw_message: bytes, sections_to_include: str | None = None) -> bytes:
+    """
+    Refines an incoming XML message based on the sections to include.
+
+    :param raw_message: The XML input.
+    :param sections_to_include: The sections to include in the refined message.
+    :return: The refined message.
+    """
+
+    if sections_to_include is None:
```
DanPaseltiner, line 95: We are checking whether `sections_to_include` is null here and on line 41. I think we can probably do this just once.

containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,92 @@ async def refine_ecr(
     data = await refiner_input.body()
 
     if sections_to_include:
-        # TODO: add logic to filter the XML message based on the sections to include
-        refined_message = Response(content=data, media_type="application/xml")
+        refined_data = refine(data, sections_to_include)
```
DanPaseltiner, line 42: If we make this line `data = refine(data, sections_to_include)`. Then we can else below and simply `return Response(content=data, media_type="application/xml")`

DanPaseltiner (changes requested):
Thanks for getting this going @m-goggins. Really exciting to see eCR filtering actually happening in code after so much talk. There are a few things I think we can improve on, but generally I really like this approach.

containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,92 @@ async def refine_ecr(
...
+        "history of present illness": "10164-2",
+        "history of immunization narrative": "11369-6",
+        "medications administered": "29549-3",
+        "plan of care note": "18776-5",
+        "problem list - reported": "11450-4",
+        "reason for visit": "29299-5",
+        "relevant diagnostic tests/laboratory data narrative": "30954-2",
+        "social history narrative": "29762-2",
+        "history of hospitalizations+outpatient visits narrative": "46240-8",
+    }
```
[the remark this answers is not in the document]
m-goggins: Yeah, expecting LOINCs is more robust. I wasn't sure if that was a reasonable expectation but I like the idea of providing the documentation to ensure we get LOINCs

containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,86 @@ async def refine_ecr(
...
+
+    section_loincs = None
+    if sections_to_include:
+        sections = sections_to_include.split(",")
+        section_loincs = []
+        for section in sections:
+            if section not in section_LOINCs:
+                raise ValueError(
+                    f"{section} is invalid. Please provide a valid section."
+                )
```
DanPaseltiner, line 78: I don't think we should raise an exception here because that will kill things and ultimately result in the client receiving a `500 internal server error` response. I think this function should return `section_loincs` as well as something like `error_message`. Then in `refine_ecr` if `error_message != ""` we can return a 422 and include the message we currently creating.

DanPaseltiner: If you'd like to handle this in another PR that works for me. I don't want to keeping blocking this PR for something that can be addressed separately.

DanPaseltiner (approved):
Thanks for the additional refinements on the refiner (haha). Left one more comment about validation of section codes. We could address in another PR. If you would like to do this @m-goggins could you please add a ticket to the backlog so we don't loose sight of it?

containers/message-refiner/app/main.py
```diff
@@ -38,9 +39,95 @@ async def refine_ecr(
...
+def refine(raw_message: bytes, sections_to_include: str) -> str:
+    """
+    Refines an incoming XML message based on the sections to include.
+
+    :param raw_message: The XML input.
+    :param sections_to_include: The sections to include in the refined message.
+    :return: The refined message.
+    """
+
+    raw_message = ET.fromstring(raw_message)
```
DanPaseltiner, line 104: Sorry to keep coming back with other things. We should probably have some error handling here in case we can't successfully parse to etree. I wonder if this should get pulled out into it's own function so that we can easily return a `400` to the client if they send us a message we can't parse.

m-goggins: That's a good point! I'm going to add it as a ticket to the backlog

PR #1704, opened 2024-04-30
remove fhir_bundles from seed scripts

BobanL:
# PULL REQUEST

## Summary
- Refactor how the database is seeded. 
  - Currently it looks at the `fhir_data` folder  and the `seed-db` job connects to the DB and inserts data
  - With this change a `data.sql` file is created that will be part of the [db init](https://hub.docker.com/_/postgres#:~:text=and%20POSTGRES_DB.-,Initialization%20scripts,-If%20you%20would).
- Remove `seed-db` job in favor of db init as mentioned above.
- `convert-seed-data` now runs a docker compose that turns on the `fhir-converter` and saves the fhir data in a file called `data.sql`

## Additional Information
- [Docker compose versions are obsolete](https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element-obsolete)

### Using this new pattern
  1. If you need to regenerate seed data then run `npm run convert-seed-data`.
  2. Delete you Postgresql volume in docker
  3. Run DB service like normal. The new data will be loaded into the DB once initialized.


gordonfarrell (approved):
This looks really great Boban, excellent changes! Thanks

PR #1707, opened 2024-04-30
Update algorithms in containerized linkage service

bamader:
# Switch to prod-tested algos

## Summary
So it turns out that even though we updated our algorithms in the evaluation notebook with all these sweet new values, our actual linkage service was still using old and outdated algorithm parameters. This PR integrates the values we determined in our spark evaluation into our DIBBs defaults, finally aligning our production testing with our service offering. As part of this update, a number of tests needed updating to reflect new parameter values, and our new utils' ability to use column names rather than column indices.

## Related Issue
No linked issue, caught as part of a RL resource collation


m-goggins (approved):
🚀 Maybe just let the IDWA team know in their channel?

bamader:
Yeah good call on letting IDWA know, once I get this integration test working I'll give them a ping. Glad we caught this!

Issue #1627, opened 2024-04-08
Address Snyk Scans


emmastephenson:
## Background
We recently re-enabled Snyk scanning for the `phdi` project. Several vulnerabilities came up, which we need to address before going live with our containers in any pilot jurisdiction.

## Action Requested
Go through the problems listed by Snyk and address, likely by updating requirements as needed.

## Acceptance Criteria (Requirements)
* Snyk security scans for our containers are clean of any major or severe issues.
* A plan is developed to continue addressing Snyk vulnerabilities as they come up

## Additional Information
Our STLT partners have indicated they would run security scans on our containers before putting them in production. To get ahead of any issues, we should address known security issues now, before we work more closely with STLTs.

If you need Snyk access slack Gordon

PR #1709, opened 2024-04-30
update python versions to 3.11-slim

BobanL:
# PULL REQUEST

## Summary
Upgrade base image from `python-3.11.5-slim` to be the latest `python-3.11-slim` as it reduces vulnerabilities generally. 

image
image

## Related Issue
Part of #1627


gordonfarrell (approved):
Nice, thanks for doing this!

PR #1751, opened 2024-05-01
Fixing issue with message-parser upload-schema

### LINKED ISSUE #1743: Tech debt: fix message-parser secondary schema bug
Currently, if you `PUT` a schema to the message-parser endpoint, that will succeed but insert NULL values for `secondary_schema` if a `secondary_schema` is not specified, which will cause that schema to fail.

We will need to:
1. Identify in the code where the NULLs are inserted.
2. Remove that aspect of the code, or otherwise pop those `secondary_schema` keys.
3. Confirm that when a loaded schema can be loaded.

This came up w/r/t to the eRSD schema work. Rather than load the schema to the `message-parser` as a permanent schema, it was decided to house it with the vocab-mapper work, as it is still a WIP and this will allow us to continue to modify the config without requiring a new version of message-parser to be released each time.

This is also generally a must-need to fulfill the promise of the message-parser in allowing anyone to load a schema that may better suit their particular needs.

robertandremitchell:
# PULL REQUEST

## Summary
This fixes an issue where the new secondary schema code would loop through even when `secondary_schema` is None. This adds an exception to prevent this from occurring in the `utils.py` and also fixes what I believe to have been the initial fix that existed in `main.py`.

This can be tested with a `PUT` to `localhost:8080/schemas/<any relevant schema>` then `localhost:8080/parse_message` with the `parsing_schema` set to the new schema.

## Related Issue
Fixes #1743 

## Additional Information
Th fix to utils.py the immediate issue of it erroring, though, I'm not quite sure I understand the core issue. When I `GET` and check for the schema `localhost:8080/schemas/ersd.json`, in this case, it was inserting a bunch of `null` secondary schemas. Example below.

I believe this occurs because of...something here not working correctly:
https://github.com/CDCgov/phdi/blob/main/containers/message-parser/app/main.py#L241-L245

I _think_ the intention of the code is to pop these empty secondary_schemas, but I'm not sure I understand where or why they're added in the first place. @bamader , I know you recently worked on this, so I'm curious if you have ideas for places we might try to remove them? Or rewrite this section to correctly pop them?  Or if it'd be a bad idea to remove them?

The uploaded schemas will still work, they're just slightly bigger with the null secondary schema items. So, we could get away with just the smaller change in utils, but the additional function in utils.py does appear to successfully pop the further nested secondary schemas.

```json
{
	"message": "Schema found!",
	"parsing_schema": {
		"value_set_type": {
			"fhir_path": "entry.resource.where(resourceType='ValueSet' and url.contains('http://ersd.aimsplatform.org/fhir/ValueSet/')",
			"data_type": "string",
			"nullable": false,
			"secondary_schema": {
				"id": {
					"fhir_path": "id",
					"data_type": "string",
					"nullable": false,
					"secondary_schema": null
				},
				"clinical_service_type": {
					"fhir_path": "title",
					"data_type": "string",
					"nullable": false,
					"secondary_schema": null
				}
			}
		},
....
``` 

## Checklist

### DIFF
--- a/containers/message-parser/app/main.py
+++ b/containers/message-parser/app/main.py
@@ -239,7 +239,6 @@ async def upload_schema(
         }
 
     # Convert Pydantic models to dicts so they can be serialized to JSON.
-    # Convert Pydantic models to dicts so they can be serialized to JSON.
     schema_dict = input.dict()
     clean_schema(schema_dict["parsing_schema"])  # remove secondary_schemas
 
--- a/containers/message-parser/app/main.py
+++ b/containers/message-parser/app/main.py
@@ -17,6 +17,7 @@ from app.models import ParseMessageResponse
 from app.models import ParsingSchemaModel
 from app.models import PutSchemaResponse
 from app.phdc.builder import PHDCBuilder
+from app.utils import clean_schema
 from app.utils import convert_to_fhir
 from app.utils import extract_and_apply_parsers
 from app.utils import freeze_parsing_schema
@@ -238,14 +239,12 @@ async def upload_schema(
         }
 
     # Convert Pydantic models to dicts so they can be serialized to JSON.
-    for field in input.parsing_schema:
-        field_dict = input.parsing_schema[field].dict()
-        if "secondary_schema" in field_dict and field_dict["secondary_schema"] is None:
-            del field_dict["secondary_schema"]
-        input.parsing_schema[field] = field_dict
+    # Convert Pydantic models to dicts so they can be serialized to JSON.
+    schema_dict = input.dict()
+    clean_schema(schema_dict["parsing_schema"])  # remove secondary_schemas
 
     with open(file_path, "w") as file:
-        json.dump(input.parsing_schema, file, indent=4)
+        json.dump(schema_dict["parsing_schema"], file, indent=4)
 
     if schema_exists:
         return {"message": "Schema updated successfully!"}
--- a/containers/message-parser/app/utils.py
+++ b/containers/message-parser/app/utils.py
@@ -112,7 +112,10 @@ def get_parsers(extraction_schema: frozendict) -> frozendict:
             ].items():
                 # Base case: secondary field is located on this resource
                 if not secondary_field_definition["fhir_path"].startswith("Bundle"):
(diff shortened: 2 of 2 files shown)


containers/message-parser/app/main.py
```diff
@@ -238,14 +239,11 @@ async def upload_schema(
         }
 
     # Convert Pydantic models to dicts so they can be serialized to JSON.
-    for field in input.parsing_schema:
-        field_dict = input.parsing_schema[field].dict()
-        if "secondary_schema" in field_dict and field_dict["secondary_schema"] is None:
-            del field_dict["secondary_schema"]
-        input.parsing_schema[field] = field_dict
```
bamader, line 245: I have no memory at all of working on this part of the code 🤔 . Upon further investigation that is because my work on refactoring did not in fact touch any of the endpoints for querying or putting configs. I think I agree with your assumption that this code is intended to remove empty or unnecessary secondary field portions, but the only way I can see they would even get there in the first place is if the user submitted a config and just didn't understand they didn't need to do that. I almost wonder if instead of cleaning after the fact, we have some kind of validation at the input level that rejects any schemas uploaded with blank secondary schema dicts? @DanielPaseltiner curious to hear your thoughts on that.

robertandremitchell: So they do populate...somewhere in the PUT or in the GET that I couldn't directly find, because the schemas I was uploading were inserting the NULLs. 

I tend to agree though that if someone uploads a schema with NULL'd secondary_schema, we should probably throw an error or warning.

DanPaseltiner: I think Pydantic actually ends up introducing these empty dicts that cause problems. So we are really cleaning up the mess that one of our tools creates, not necessarily one that the client sends us. @robertandremitchell discovered this when attempting upload schemas that don't define a `secondary_schema`.

DanPaseltiner (approved):
LGTM. Thanks @robertandremitchell

containers/message-parser/app/main.py
robertandremitchell: ahhh okay, that would explain it. Then yeah, the new utils function should cover this well. thank you!

PR #1755, opened 2024-05-02
Create vocab mapper service stubs

### LINKED ISSUE #1665: Create a stubbed version of the service
Let's stub out the terminology service to lay the foundation for its development and allow us to generate proposed API documentation for the streamline team review. We need to stub out two endpoints:

1. `/ersd/insert-condition-extensions`
    - This endpoint with have a FHIR bundle POSTed to it. Extensions will be added to each resource pertinent to given conditions. We should check with the streamline team about how these conditions will be provided. Are the conditions specified somewhere in the FHIR bundle? Should they be specified as query params?
 
2. `/ersd/clinical-services/{condition} 
    - Given a condition this endpoint will return the value sets pertinent to the condition. Should we return these value sets as FHIR use `valueset` resources? Should we offer a `service_type` query param to allow clients to specify which clinical services they would like value sets for?

*I'm proposing appending `ersd/` to these endpoint paths to allow for other types of terminology uses cases potentially on different datasets to be added to this service in the future. Happy to discuss further.

Acceptance Criteria:
A stubbed service with these endpoints in addition to the health check is available on `main`.

bamader:
# Initialize terminology service

## Summary
This PR sets up the infrastructure and container boilerplate for our terminology service. We follow the same pattern as the message parser, message refiner, and orchestration services, in terms of managing docker files, directory structure, unit and integration tests, and pydantic model inputs. 

Of particular note for team members, this changeset introduces two endpoints that will be used for condition value set purposes (both querying/retrieving and extending). Let's make sure this API structure works for all devs working on this code. I'm able to build and run this container locally (not that it does a whole lot yet) no problem and get responses from the healthcheck, so if the endpoints and model inputs look good, we can move forward with incorporating actual logic into the endpoints.

## Related Issue
Fixes #1665 

## Additional Information
- [x ] If this code affects the other scrum team, have they been notified? (In Slack, as reviewers, etc.)

### DIFF
--- a/containers/vocab-mapper/app/main.py
+++ b/containers/vocab-mapper/app/main.py
@@ -1,0 +1,49 @@
+from pathlib import Path
+
+from dibbs.base_service import BaseService
+from fastapi import Response
+
+from app.models import InsertConditionInput
+from app.models import ValueSetForConditionInput
+
+# Instantiate FastAPI via DIBBs' BaseService class
+app = BaseService(
+    service_name="Vocabulary Mapper",
+    service_path="/vocab-mapper",
+    description_path=Path(__file__).parent.parent / "description.md",
+).start()
+
+
+@app.get("/")
+async def health_check():
+    """
+    Check service status. If an HTTP 200 status code is returned along with
+    '{"status": "OK"}' then the mapper service is available and running
+    properly.
+    """
+    return {"status": "OK"}
+
+
+@app.post("/ersd/insert-condition-extensions/")
+async def insert_condition_extensions(input: InsertConditionInput, response: Response):
+    """
+    Extends the resources of a supplied FHIR bundle with extension tags
+    related to one or more supplied conditions. For each condition in the
+    given list of conditions, each resource in the bundle is appended with
+    an extension structure indicating which snowmed condition code the
+    resource is linked to.
+    """
+    # TODO: This method is a stub.
+    return {"extended_bundle": input.bundle}
+
+
+@app.post("/ersd/get-value-sets/")
+async def get_value_sets_for_condition(
+    input: ValueSetForConditionInput, response: Response
+):
+    """
+    For a given condition, queries and returns the value set of clinical
+    services associated with that condition.
+    """
+    # TODO: This method is a stub.
+    return {"value_set": []}
--- a/containers/vocab-mapper/app/models.py
+++ b/containers/vocab-mapper/app/models.py
@@ -1,0 +1,53 @@
+from typing import List
+
+from pydantic import BaseModel
+from pydantic import Field
+from pydantic import root_validator
+
+
+class InsertConditionInput(BaseModel):
+    """
+    The schema for requests to the /insert-condition-extensions endpoint.
+    """
+
+    bundle: dict = Field(
+        description="The FHIR bundle to modify. Each resource in the bundle related "
+        "to one or more of the conditions in the other supplied parameter will have "
+        "an extension added to the resource noting the snowmed code relating to the "
+        "associated condition(s)."
+    )
+    conditions: List[str] = Field(
+        description="The list of snowmed codes to insert as extensions into any "
+        "associated resources in the supplied FHIR bundle."
+    )
+
+    @root_validator
+    def require_non_empty_condition_list(cls, values):
+        """
+        Ensures that the supplied conditions list parameter is both a list
+        and has at least one element by which to extend the supplied FHIR
+        bundle.
+
+        :param cls: The InsertConditionInput class.
+        :param values: The condition list supplied by the caller.
+        :raises ValueError: Errors when supplied condition list contains no
+          elements, since this can't be used to extend a bundle.
+        :return: The endpoint's values, if the condition list is valid.
+        """
+        if len(values.get("conditions")) == 0:
+            raise ValueError(
+                "Supplied list of snowmed conditions must contain "
+                "one or more elements; given list was empty."
+            )
+        return values
+
+
+class ValueSetForConditionInput(BaseModel):
+    """
+    The schema for requests to the /get-value-sets endpoint.
+    """
+
+    condition: str = Field(
+        description="The snowmed code for a condition of interest, whose value "
+        "sets will be retrieved from the service."
+    )
--- a/containers/vocab-mapper/description.md
+++ b/containers/vocab-mapper/description.md
@@ -1,0 +1,50 @@
+## Getting Started with the DIBBs Vocabulary Mapper Service
+
+### Introduction
+
+The DIBBs vocabulary mapper (vocab-mapper) service offers a REST API devoted to querying and enriching SNOWMED condition code analysis. This service stores condition codes and their associated value sets so that they can be queried by users, as well as inserted into supplied FHIR bundles as tagged extensions for future path parsing.
+
+### Running the eCR Refiner
+
+The vocab mapper can be run using Docker (or any other OCI container runtime e.g., Podman), or directly from the Python source code.
+
+#### Running with Docker (Recommended)
+
+To run the voca mapper with Docker, follow these steps.
+
+1. Confirm that you have Docker installed by running `docker -v`. If you do not see a response similar to what is shown below, follow [these instructions](https://docs.docker.com/get-docker/) to install Docker.
+
+```
+❯ docker -v
+Docker version 20.10.21, build baeda1f
+```
+
+2. Download a copy of the Docker image from the PHDI repository by running `docker pull ghcr.io/cdcgov/phdi/vocab-mapper:latest`.
+3. Run the service with ` docker run -p 8080:8080 vocab-mapper:latest`.
+
+Congratulations, the vocab mapper should now be running on `localhost:8080`!
+
+#### Running from Python Source Code
+
+We recommend running the vocab mapper from a container, but if that is not feasible for a given use-case, it may also be run directly from Python using the steps below.
+
+1. Ensure that both Git and Python 3.10 or higher are installed.
+2. Clone the PHDI repository with `git clone https://github.com/CDCgov/phdi`.
+3. Navigate to `/phdi/containers/vocab-mapper/`.
+4. Make a fresh virtual environment with `python -m venv .venv`.
+5. Activate the virtual environment with `source .venv/bin/activate` (MacOS and Linux), `venv\Scripts\activate` (Windows Command Prompt), or `.venv\Scripts\Activate.ps1` (Windows Power Shell).
+6. Install all of the Python dependencies for the vocab mapper with `pip install -r requirements.txt` into your virtual environment.
+7. Run the vocab mapper on `localhost:8080` with `python -m uvicorn app.main:app --host 0.0.0.0 --port 8080`.
+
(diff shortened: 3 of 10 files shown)


containers/vocab-mapper/app/models.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    bundle: dict = Field(
+        description="The FHIR bundle to modify. Each resource in the bundle related "
+        "to one or more of the conditions in the other supplied parameter will have "
+        "an extension added to the resource noting the snowmed code relating to the "
+        "associated condition(s)."
+    )
+    conditions: List[str] = Field(
+        description="The list of snowmed codes to insert as extensions into any "
```
robertandremitchell, line 20: `snomed`. I had to catch this one a few times and had Mandela effect'd myself into believing it had to be snowmed.

containers/vocab-mapper/description.md
```diff
@@ -0,0 +1,50 @@
...
+
+The DIBBs vocabulary mapper (vocab-mapper) service offers a REST API devoted to querying and enriching SNOWMED condition code analysis. This service stores condition codes and their associated value sets so that they can be queried by users, as well as inserted into supplied FHIR bundles as tagged extensions for future path parsing.
+
+### Running the eCR Refiner
+
+The vocab mapper can be run using Docker (or any other OCI container runtime e.g., Podman), or directly from the Python source code.
+
+#### Running with Docker (Recommended)
+
+To run the voca mapper with Docker, follow these steps.
```
robertandremitchell, line 13: vocab mapper. More so flagging so that when we had to replace all these references down the line with the name of the new service we don't miss this one.

containers/vocab-mapper/tests/integration/test_vocab_mapper.py
```diff
@@ -0,0 +1,10 @@
+import httpx
+import pytest
+
+PARSER_URL = "http://0.0.0.0:8080"
+
+
+@pytest.mark.integration
```
robertandremitchell, line 7: test didn't work for me until I added an empty `assets` folder because it's referenced in the `Dockerfile`. so we might just want to include the empty folder in the PR?

```
(.venv) rob@Roberts-MacBook-Pro vocab-mapper % pytest tests/integration -vv -s
=========================================================================== test session starts ============================================================================
platform darwin -- Python 3.10.14, pytest-7.1.2, pluggy-1.0.0 -- /Users/rob/phdi/containers/orchestration/.venv/bin/python3.10
cachedir: .pytest_cache
rootdir: /Users/rob/phdi, configfile: pytest.ini
plugins: asyncio-0.23.5, syrupy-4.6.1, env-1.1.3, Faker-18.13.0, anyio-3.7.1, mock-3.14.0, requests-mock-1.12.1
asyncio: mode=strict
collected 1 item                                                                                                                                                           

tests/integration/test_vocab_mapper.py::test_health_check Setting up tests...
WARN[0000] /Users/rob/phdi/containers/vocab-mapper/tests/integration/docker-compose.yaml: `version` is obsolete 
[+] Building 0.1s (10/11)                                                                                                                              docker:desktop-linux
 => [vocab-mapper-service internal] load build definition from Dockerfile                                                                                              0.0s
 => => transferring dockerfile: 325B                                                                                                                                   0.0s
 => [vocab-mapper-service internal] load metadata for ghcr.io/cdcgov/phdi/dibbs:latest                                                                                 0.1s
 => [vocab-mapper-service internal] load .dockerignore                                                                                                                 0.0s
 => => transferring context: 2B                                                                                                                                        0.0s
 => [vocab-mapper-service 1/7] FROM ghcr.io/cdcgov/phdi/dibbs:latest@sha256:a28c6414744ad190263bd668f55a2c2d2a9d8b1f34667636bd5d8356cdb6ff0e                           0.0s
 => [vocab-mapper-service internal] load build context                                                                                                                 0.0s
 => => transferring context: 399B                                                                                                                                      0.0s
 => CACHED [vocab-mapper-service 2/7] WORKDIR /code                                                                                                                    0.0s
 => CACHED [vocab-mapper-service 3/7] COPY ./requirements.txt /code/requirements.txt                                                                                   0.0s
 => CACHED [vocab-mapper-service 4/7] RUN pip install -r requirements.txt                                                                                              0.0s
 => CACHED [vocab-mapper-service 5/7] COPY ./app /code/app                                                                                                             0.0s
 => ERROR [vocab-mapper-service 6/7] COPY ./assets /code/assets                                                                                                        0.0s
------
 > [vocab-mapper-service 6/7] COPY ./assets /code/assets:
------
failed to solve: failed to compute cache key: failed to calculate checksum of ref 0231bb62-3b2d-452d-b045-2c61119d126e::wxtz6xfcsh6t0n0vejbldq4p3: "/assets": not found
Waiting to be ready...
```

robertandremitchell (changes requested):
minor formatting things, but overall looks good!

containers/vocab-mapper/app/main.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@app.get("/")
+async def health_check():
+    """
+    Check service status. If an HTTP 200 status code is returned along with
+    '{"status": "OK"}' then the mapper service is available and running
+    properly.
+    """
+    return {"status": "OK"}
```
DanPaseltiner, line 24: I believe this comes for free from the `BaseService` class so I don't think we need to redefine it here.

containers/vocab-mapper/app/main.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@app.post("/ersd/insert-condition-extensions/")
```
DanPaseltiner, line 27: I'm wondering if it is actually useful to include `ersd` in these endpoint paths? I know I included them in the ticket, but in principle we could add value sets that come from another source to the schema of this DB and serve them with this endpoint.

containers/vocab-mapper/app/main.py
```diff
@@ -0,0 +1,49 @@
...
+@app.post("/ersd/get-value-sets/")
+async def get_value_sets_for_condition(
+    input: ValueSetForConditionInput, response: Response
+):
+    """
+    For a given condition, queries and returns the value set of clinical
+    services associated with that condition.
+    """
+    # TODO: This method is a stub.
+    return {"value_set": []}
```
DanPaseltiner, line 49: We'll need to talk to Streamline about what will work best for them on `/ersd/insert-condition-extensions/`, but I really think this end point should be a simple `GET` where the condition code is provided as a query param. Generally speaking `POST` is used to send data to the server while `GET` is used to retrieve data. @m-goggins has implemented this kind of pattern on the eCR refiner if you'd like an example.

containers/vocab-mapper/app/models.py
DanPaseltiner: SNOMED-CT should be capitalized as it is an acronym that stands for Systematized Medical Nomenclature for Medicine–Clinical Terminology. Often it is shortened to just SNOMED, but I think should probably capitalized in either case.

containers/vocab-mapper/app/main.py
```diff
@@ -0,0 +1,49 @@
...
+
+@app.post("/ersd/insert-condition-extensions/")
+async def insert_condition_extensions(input: InsertConditionInput, response: Response):
+    """
+    Extends the resources of a supplied FHIR bundle with extension tags
+    related to one or more supplied conditions. For each condition in the
+    given list of conditions, each resource in the bundle is appended with
+    an extension structure indicating which snowmed condition code the
+    resource is linked to.
+    """
```
DanPaseltiner, line 35: We are missing params and returns in this doc string as well as a return type on the function. Ditto for the endpoint below.

containers/vocab-mapper/app/models.py
bamader: Easy fix, can do!

containers/vocab-mapper/description.md
bamader: Ah great catch. Stupid keypress 😒

containers/vocab-mapper/tests/integration/test_vocab_mapper.py
bamader: Ugh yeah same. I actually created that folder on my local, but it looks like github didn't add it to the commit because the directory is empty and it can only do filepaths. Tbh might put a blank placeholder file in there for the moment until we have use to put assets in.

containers/vocab-mapper/app/main.py
bamader: Huh, we explicitly define it in all of the services I mentioned which also use the BaseService to startup. But looking at the code for BaseService, it does look like we define one. Should I leave to just follow the pattern, or should I clean up those other health check endpoints?

containers/vocab-mapper/app/main.py
bamader: I personally thought we didn't need `ersd` but my clinical knowledge is nowhere near as developed as yours so I just tend to defer to you on things related to PH domain expertise. But yeah I don't actually think the `ersd` flag is doing anything for us here other than lengthening the endpoint.

containers/vocab-mapper/app/main.py
bamader: I copied this format from our message parser docstrings, which also don't have those types 😳 . Should I go back and add typing there too?

bamader: Our orchestration service does indeed have params/returns for its endpoint, so I suspect the message parser one is an oversight. Adding them in here.

containers/vocab-mapper/tests/integration/test_vocab_mapper.py
bamader: Got all tests to be okay using some placeholder directories and files. This is a known issue with git (which they call a feature, debatable haha) so this seems like an easy workaround until we're ready to overwrite.

robertandremitchell (approved):
looks good!

containers/vocab-mapper/app/main.py
DanPaseltiner: We can leave for now, although I think we introduced health check on the base service and the never removed the redundant definitions everywhere else (record linkage is an exception). Since we are using SQLite which is just a file we won't need to check for a DB connection because our build process and testing will guarantee the db file is included and that we can connect to it. If we migrate to PG or some other RDBMS in the future then I completely agree.

containers/vocab-mapper/app/main.py
```diff
@@ -0,0 +1,62 @@
...
+
+    :param condition_code: A query param supplied as a string representing a
+      single SNOMED condition code.
+    :return: An HTTP Response containing the value sets of the queried code.
+    """
+    if condition_code is None or condition_code == "":
+        return Response(
+            content="Supplied condition code must be a non-empty string",
+            status_code=422,
+        )
```
DanPaseltiner, line 59: Looks good for now. In the future we'll probably want to validate  the codes against the DB to make sure they are all meaningful. Definitely a separate ticket though.

DanPaseltiner (approved):
Thanks for addressing all the feedback. Let's get this merged and see what other's have to say about this as our proposed API contract.

Issue #1744, opened 2024-05-01
Create indices for the the ersd.db tables


robertandremitchell:
To enhance the speed of the future API service in being able to call conditions, find relevant value sets, and load any and all relevant clinical services, those tables will need to have an index. As found in the other record-linkage work, creating the index will allow for speedier and less memory-intensive joins.

Prior work on `record-linkage` should be able to serve as a basis for this code in adding it to the relevant tables in the DB.

robertandremitchell:
Adding indices in sqlite should be relatively straightforward. Primary thing we need to do is determine indices we want. 

Based on the eRD: https://drive.google.com/file/d/1nxm0HpiJsONizM5NzvSIH2dbUCcDif6f/view, I think we'll want:

* conditions: index for id and value_set_id
* value_sets: index for id and clinical_service_type_id
* value_set_type: index for id
* clinical_services: index for id and value_set_id


We should just be able to write a short function that dynamically adds an index:
```
CREATE INDEX IF NOT EXISTS f"idx_{table_name}_{column_name}" ON {table_name}({column_name})
```
where the input would be the table_name we are adding it to and the column name for which the index should be added.

Alternatively, we could add it to the create_table function, although that already has like 4 arguments, so I think adding them post-table creation maybe makes more sense? Especially because some tables require multiple indices, so it could get a little wonky and over-tuned to add them to create_table.

robertandremitchell:
completed as part of work on #1767

PR #1767, opened 2024-05-03
Refining SQL database creation in vocab-mapper

### LINKED ISSUE #1745: Refactor sql code in seed-database.py scripting
This will involve:

1. Moving the sql queries out of seed-database.py into a new sql file.
2. Adding code in the creation step to enforce foreign key. This will allow us to also remove the helper function that checks the list of lists for uniqueness and move this check to sqlite, ideally.
3. Migrating the DDL for creating the tables in the DB: decouple table creation and data loading such that we run migrations to put the DB in the desired state - creation of tables, constraints, indexes, etc. then load data.

robertandremitchell:
# PULL REQUEST

## Summary
Based on prior feedback, this amends the SQL code to more closely follow the structure used in record-linkage to create tables, add indices, and overall refine/simplify the SQL code for scalability.

## Related Issue
Fixes #1745, #1744

## Additional Information
* I did not create separate .sql scripts for deleting and inserting tables. I could see the argument for adding either more explicitly, but the syntax for each is handled pretty capably within python to dynamically build their queries.

Whereas, the logic for table creation and index setting does feel right we want that very explicit. I can see the argument either way, so curious what others think.

* I switched to executemany for our INSERT after reading up on that being a tad more performant.

* Conversely, I also did a lot of reading and could not find a definitive read on whether we ought to create indexes first or load data first. I tried both and did not see any performance diff.

## Checklist


containers/vocab-mapper/seed-scripts/migrations/V01_02__indexes.sql
DanPaseltiner, line 1: We can combined into a single file here. They are broken on record linkage because we started without indexing and needed to add them to an existing database.

DanPaseltiner (approved):
One minor suggestion below. Otherwise LGTM! Have a great weekend thanks for your work on this @robertandremitchell.

DanPaseltiner:
Oh one additional request. Could we rename the PR as this doesn't include a "queries" strictly speaking. We have database migrations in the form of DDL (Data Definition Language). I'd like to make sure that the git log remains interpretable especially when we have upcoming tickets that will be about truly querying the DB.

containers/vocab-mapper/seed-scripts/migrations/V01_02__indexes.sql
robertandremitchell: ah, makes sense. sounds good, updated!

PR #1773, opened 2024-05-03
Deploy and Octopus workflows

alismx:
# PULL REQUEST

## Summary
- Adds a new deploy workflow that gets a tag, builds and pushes docker images and calls octopus for release and deployments.
- Some GH action version updates

## Related Issue
- resolves CDCgov/dibbs-cloud#46 (for now...)

## Additional Information
- Due to some unknowns in our remote environments downstream, I've made this manually triggered at the moment so that we don't have to watch failed jobs on merge or worry about it before we're ready.
- Trying to deploy `Staging` or `Production` should cause the job to exit, these environments are not yet available.

## Checklist


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,75 @@
+---
+name: Deploy
+run-name: Deploy ${{github.event.inputs.branch }} to ${{ github.event.inputs.environment }} by @${{ github.actor }}
```
rin-skylight, line 3: Non-blocking: inconsistent spacing for the variable.

.github/workflows/octopus.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+jobs:
+  octopus-release:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Required for authentication
+    outputs:
+      release_number: ${{ steps.create_release.outputs.release_number }}
+    steps:
+      - name: Login to Octopus 🐙
```
rin-skylight, line 41: <squee>! The emojis! Love this!

rin-skylight (approved):
Thank you for all your work on this!

.github/workflows/deploy.yaml
alismx: Fixed :heavy_check_mark:

.github/workflows/octopus.yaml
alismx: :octopus:

alismx:
> @alismx this looks awesome! I'd like to apologize for the delayed review here, catching up after being OOO at a conference for most of last week.
> 
> I have one question. Maybe this is something I have missed, but I think we have eliminated the steps when we make a release that updates our API docs. The current process starts [on line 168](https://github.com/CDCgov/phdi/blob/main/.github/workflows/createNewRelease.yaml#L168) in `createNewRelease.yaml`.
> 
> We are slowly deprecating the Python SDK called `phdi` so updating those docs is no longer important, but the swagger docs for the APIs of all of the building blocks is are definitely something we want to maintain and continue to auto update.

@DanPaseltiner Thanks for taking a look! I understand what you're saying. I'll get that swagger doc process added back in.

alismx:
Discussed with @DanPaseltiner , going to leave the current release workflow in place along with not fixing the broken doc generation steps as it's also broken on `main`. For now, if this workflow is triggered, docs will go to the `alis/tmp-docs` branch. The new workflow will be manual until more infrastructure is in place to require productionizing this process.

If I could get some reviews, I'd appreciate it! @DanPaseltiner @rin-skylight @marycrawford

PR #1783, opened 2024-05-06
Render Diagnostic Reports in TEFCA Viewer

### LINKED ISSUE #1713: Add rendering for DiagnosticReport resources

m-goggins:
# PULL REQUEST

## Summary
Renders the diagnostic reports of a returned bundle. The [Connectathon specifications](https://docs.google.com/spreadsheets/d/1m4YTTMYBNMTwXWPG1EkpkY7iWIUHT0Ul4MXU2DTGP94/edit?pli=1#gid=1877733207) for diagnostic Reports are:

- Code (diagnosticReport.Code) 
- Result (diagnosticReport.result|Observation)

## Related Issue
Fixes #1713 

## Additional Information
The Connectathon requirements for Diagnostic Reports display the same information as the Observation results. We can display the DRs for now and get more info from users at the connectathon, but I am not sure what added value they provide to the user right now.

image

## Checklist


containers/tefca-viewer/src/app/query/components/ObservationTable.tsx
```diff
@@ -32,7 +32,7 @@ const ObservationTable: React.FC<ObservationTableProps> = ({
       <tbody>
         {observations.map((obs) => (
           <tr key={obs.id}>
-            <td>{obs?.effectiveDateTime}</td>
```
DanPaseltiner, line 35: I wasn't quite sure which date field to grab when I wrote this piece. Does `issued` seem more present to you in the connectathon data? Should we create a function for this so we can check multiple places with an order of preference like we do in `formatValue`? I don't think any additional work is required before merging, just looking to understand the motivation for this change so we can plan accordingly.

DanPaseltiner (approved):
LGTM. One non-blocking question below. Thanks @m-goggins!

containers/tefca-viewer/src/app/query/components/ObservationTable.tsx
m-goggins: Yeah, I saw `issued` more but the sample size is obviously small. I'm not sure what the difference is between `effectiveTimeDate` and `issued` other than `issued` appears to have a timestamp as well.

m-goggins: I like the idea of adding a datetime/date formatting function. I'll add a separate ticket for it to the backlog.

PR #1795, opened 2024-05-06
Getting Clinical Services from SNOMED

### LINKED ISSUE #1756: Write a query to look at condition code then join back to clinical services
We need to be able to have a query that reads in the `ersd.db` tables to do the following:

* Query the `conditions` table based on the SNOMED condition `id`s provided by an input
* * We will also want to provide keywords or some other way for folks to query for the `clinical_service_type` they are most interested in. This would likely be an optional parameter
* Join on `value_set_id` to `value_sets`
* Join on `clinical_service_type_id` to `value_set_type` (Maybe. This table only has the description of the clinical service types, so it could be optional)
* Join on `value_set_id` to `clinical_services`
* Return `code`s and `display`s for all the `clinical_services`, grouped by `clinical_service_type`
* * If the optional parameter for `clinical_service_type` is provided, then limit return to the relevant `clinical_service_type`

robertandremitchell:
# PULL REQUEST

## Summary
This implements the helper function that will be needed to take SNOMED condition code provided by a user and return them in a structure (defined here: https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/1788) that will allow us to use the API endpoint for the ecr-refiner.

## Related Issue
Fixes #1756 

## Additional Information

* Is this the most performant way to do the JOINs and SQL query?
* Are there other sanitization steps we should add to avoid SQL injection?
* Should the clinical service type parsing be done in SQL, or on the dictionary? It works in the dictionary without issue, but if we wanted to write a more complex sql_query, we could potentially add that.

## Checklist


containers/vocab-mapper/app/utils.py
```diff
@@ -0,0 +1,111 @@
...
+            for type in clinical_service_dict.keys()
+            if type not in clinical_services
+        ]
+        # Remove the types
+        for type in remove_list:
+            clinical_service_dict.pop(type, None)
+    return clinical_service_dict
+
+
+pprint.pprint(get_clinical_service_dict("276197005, junk"))
```
akintner, line 111: If this is a relic from the mob session earlier, you'll probably want to remove eventually.

robertandremitchell: ha, yeah, good catch, thanks! 😅  
removed!

containers/vocab-mapper/app/utils.py
```diff
@@ -0,0 +1,111 @@
...
+        c.id = ?
+    GROUP BY
+        vs.clinical_service_type_id, cs.code_system
+    """
+
+    # Connect to the SQLite database, execute sql query, then close
+    conn = sqlite3.connect("seed-scripts/ersd.db")
+    cursor = conn.cursor()
+    try:
+        cursor.execute(sql_query, snomed_id)
```
DanPaseltiner, line 79: I believe `sqlite3` automatically handles sanitization when using `execute` like this. Therefore, I think we can rename and simply `sanitize_inputs_to_list` to just handle transforming a comma delimited string into a `list`.

containers/vocab-mapper/app/utils.py
```diff
@@ -0,0 +1,111 @@
...
+    :return: A nested dictionary with clinical service type as the key with
+    the relevant codes and code systems as objects within.
+    """
+    # sanitize snomeds - confirm only 1 snomed provided
+    snomed_id = sanitize_inputs_to_list(snomed_id)
+    if len(snomed_id) != 1:
+        return {
+            "error": f"{len(snomed_id)} SNOMED codes provided. "
+            + "Provide only one SNOMED code."
+        }
```
DanPaseltiner, line 55: Could we flatten things out a bit here to have each function do ideally only one thing and make the input of  `get_clinical_service_dict` be a list of snomed codes? Meaning we would call `sanitize_inputs_to_list` and then pass the return value into this function as a param?

containers/vocab-mapper/app/utils.py
```diff
@@ -0,0 +1,111 @@
...
+        clinical_services = sanitize_inputs_to_list(clinical_services)
+        # Create a list of types to remove
+        remove_list = [
+            type
+            for type in clinical_service_dict.keys()
+            if type not in clinical_services
+        ]
+        # Remove the types
+        for type in remove_list:
+            clinical_service_dict.pop(type, None)
```
DanPaseltiner, line 110: Could we pull each of these steps out into their own functions that get called with the output of this function? I guess that will also require renaming this function. We are just doing a lot of different things in here and from testing and reusability standpoints breaking things up into smaller units that are decouple will likely benefit us.

DanPaseltiner (changes requested):
Super cool to see this working! Thank you @robertandremitchell and Vocabulizer team for your work here. A few thoughts below on how I think we can refactor things a bit.

containers/vocab-mapper/app/utils.py
bamader: Big +1 to this, I like keeping the input signature as small as possible. The function that submits the query to the DB shouldn't have to worry about the typing, imo.

containers/vocab-mapper/app/utils.py
DanPaseltiner: I'm happy to have a single format function that is called with the data in whatever form it comes back from the query. Really I just don't want formatting to be coupled to the query. I think we should

1. Prepare list of conditions codes
2. Pass list of conditions codes into a function to query the DB
3. Take the data that comes back from the DB and format as needed

3 functions that can be composed as opposed to called within each other. 

If we need to change how to format this data we should not have to touch the logic for querying the DB.

containers/vocab-mapper/app/utils.py
robertandremitchell: makes sense. new push has this split out a bit more.

containers/vocab-mapper/app/utils.py
```diff
@@ -0,0 +1,111 @@
...
+    :return: A list free of excess whitespace
+    """
+    if isinstance(value, (int, float)):
+        return [str(value)]
+    elif isinstance(value, str):
+        common_delimiters = [",", "|", ";"]
+        for delimiter in common_delimiters:
+            if delimiter in value:
+                return [val.strip() for val in value.split(delimiter) if val.strip()]
+        return [value.strip()]  # No delimiter found, return the single value
```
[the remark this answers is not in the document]
robertandremitchell: I used to be a big into using `re` for a lot of string stuff I've needed to do. `re.sub` was my bestie

containers/vocab-mapper/app/utils.py
```diff
@@ -0,0 +1,111 @@
...
+    clinical service code, and clinical service system from the eRSD database.
+
+    It will then parse that information into a dictionary for use in the
+    /get-value-sets API endpoint.
+
+    There is an optional parameter to return select clinical service type(s)
+    specified as either a string or a list.
+
+    :param snomed_id: SNOMED code to check
+    :param clinical_services: List of clinical service types to keep
```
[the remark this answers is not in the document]
robertandremitchell: yeah, this is a good call out, I had a tough time trying to find the right way to frame. It may make sense to rename, but not sure what I'd change to. I changed up this description a little to try to make it clearer that by default all 6 will be kept.

containers/vocab-mapper/app/utils.py
robertandremitchell: I've decoupled a bit more in the latest pushes, which has put it more in the layout I think Dan describes. Curious your thoughts; I could definitely see linking these more, particularly because the formatting from list of tuples to nested dict is something we will presumably always do, so I could see that becoming a helper function.

containers/vocab-mapper/app/utils.py
robertandremitchell: makes sense. I redid some of the cleaning here

containers/vocab-mapper/app/utils.py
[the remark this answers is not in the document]
robertandremitchell: lmk what you think of `filter_clinical_services`, I think this was my best idea so far.

robertandremitchell:
@DanPaseltiner and @bamader , new version of code ready for review, curious your thoughts. I also merged with main to get this moved over to `trigger-code-reference`

containers/trigger-code-reference/app/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    :param value: string, int, float, list to check
+    :return: A list free of excess whitespace
+    """
+    if isinstance(value, (int, float)):
+        return [str(value)]
+    elif isinstance(value, str):
+        common_delimiters = [",", "|", ";"]
+        for delimiter in common_delimiters:
+            if delimiter in value:
+                return [val.strip() for val in value.split(delimiter) if val.strip()]
```
nickclyde, line 22: Gotta love a good list comprehension 🔥

nickclyde (approved):
Looks great Rob, now that the function concerns are a bit more separated, I think this is much more readable 👍🏻

DanPaseltiner (approved):
Thanks for the refactors here @robertandremitchell. Two minor suggestions and some ramblings about DB connection management below, none of which are blocking from my perspective.

containers/trigger-code-reference/app/utils.py
```diff
@@ -0,0 +1,134 @@
+import sqlite3
+from typing import List
+from typing import Union
+
+
+def sanitize_inputs_to_list(value: Union[list, str, int, float]) -> list:
```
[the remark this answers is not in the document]
robertandremitchell: makes sense. updating

containers/trigger-code-reference/app/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    :return: A one-item list of a cleaned SNOMED code.
+    """
+    clean_snomed_code = sanitize_inputs_to_list(snomed_code)
+    if len(clean_snomed_code) != 1:
+        return {
+            "error": f"{len(clean_snomed_code)} SNOMED codes provided. "
+            + "Provide only one SNOMED code."
+        }
+    else:
+        return clean_snomed_code
```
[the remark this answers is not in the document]
robertandremitchell: makes sense. updating

containers/trigger-code-reference/app/utils.py
```diff
@@ -0,0 +1,134 @@
...
+        code = get_clean_snomed_code(snomed_code)
+        cursor.execute(sql_query, code)
+        clinical_services_list = cursor.fetchall()
+        if not clinical_services_list:
+            return {"error": f"No data found for the SNOMED code: {code}."}
+    except sqlite3.Error as e:
+        return {"error": f"An SQL error occurred: {str(e)}"}
+    finally:
+        conn.close()
+    return clinical_services_list
```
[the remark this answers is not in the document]
robertandremitchell: Good call out. I can play around with this section a bit more this morning. I think having a context manager makes sense and mirrors what I see in other parts of the code.

Yeah, that's a good point re: the opening/closing. I could see moving the sql connection to a separate function that might allow us to run multiple calls.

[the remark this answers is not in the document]
robertandremitchell: switched to `with`.  reading up one some other options to keep the connection open if needed, but I think to your point it may make more sense to revisit once we know better what this scaling might look like on the `ecr-refiner`.

PR #1810, opened 2024-05-08
Render medication requests in TEFCA Viewer & Refactor parseFhirSearch to handle parsing multiple resources at once

### LINKED ISSUE #1717: Add rendering for medication resources

m-goggins:
# PULL REQUEST

## Summary
This makes 3 changes:
1. Renders the medication requestions of a returned bundle according to the STI use cases in the [Connectathon specifications](https://docs.google.com/spreadsheets/d/1m4YTTMYBNMTwXWPG1EkpkY7iWIUHT0Ul4MXU2DTGP94/edit?pli=1#gid=1877733207). 
2. Refactors `parseFhirSearch` to parse multiple resources and appropriately add them to the `queryResponse`. This is useful in the case of the medication requests because I used `_include` in the query, which returns multiple FHIR resources from a single query, and now `parseFhirSearch` can add each resource type to the appropriate section of `queryResponse`, i.e., Medications are added to `queryResponse.Medication`. 
3. Adds `dev-win` to allow for local development and testing on Windows

## Related Issue
Fixes #1717

## Additional Information
The refactoring of `parseFhirSearch` will mean we can update the condition and encounter queries to a single query in another PR.

The Medication Requests render like this:
image

## Checklist


containers/tefca-viewer/package.json
```diff
@@ -5,6 +5,7 @@
   "private": true,
   "scripts": {
     "dev": "NODE_TLS_REJECT_UNAUTHORIZED=0 next dev",
+    "dev-win": "set NODE_TLS_REJECT_UNAUTHORIZED=0 && next dev",
```
DanPaseltiner, line 8: 🚀  glad this working!

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -65,20 +66,21 @@ type UseCaseQueryRequest = {
...
+  MedicationRequest?: MedicationRequest[];
 };
 
 const useCaseQueryMap: {
   [key in USE_CASES]: (
     input: UseCaseQueryRequest,
     queryResponse: QueryResponse,
-  ) => Promise<void>;
+  ) => Promise<QueryResponse>;
 } = {
```
DanPaseltiner, line 84: I'm looking at this again and maybe I just don't understand the syntax here, but I'm not sure these types are correct. We should have a map whose keys are strings of type `USE_CASES` and values are functions (`Callable` in Python). It is true that all of these functions return a `Promise<QueryResponse>`, but is that important typing information for the map?

containers/tefca-viewer/src/app/query/components/MedicationRequestTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <td>
+              {formatCodeableConcept(
+                medicationRequest.medicationCodeableConcept,
+              )}
+            </td>
+            <td>{formatCodeableConcept(medicationRequest?.reasonCode?.[0])}</td>
+          </tr>
+        ))}
+      </tbody>
+    </Table>
```
DanPaseltiner, line 42: I think we might want to include `MedicationRequest.status` in the table as well. We just leave it and ask folks at the connectathon what data they'd like to see and update accordingly.

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -217,147 +222,174 @@ async function newbornScreeningQuery(
...
-      return body.entry.map((entry: any) => entry.resource);
+      for (const entry of body.entry) {
+        if (resourceTypes.includes(entry.resource.resourceType)) {
+          output[entry.resource.resourceType as string].push(entry.resource);
+        }
+      }
     }
   }
-  return [];
+  return output;
```
DanPaseltiner, line 394: This approach makes a lot of sense to, but what about making a slight change to have this function return a `QueryResponse` object? This would allow us not to provide `resourceTypes` as a parameter. Instead the function can instantiate a `QueryResponse`, sort the resources in the bundle into the correct buckets (as defined by `QueryResponse`), and return them without needing to be told explicitly which resources types the query will return. Then we can take the `QueryResponse` retuned and merge with the instance of `QueryResponse` in the calling function and continue on our merry way. This allows us to continue just writing FHIR API calls without also having to state the resources types each call will return. Thinking ahead to when we want to make batches of requests concurrently the simpler the interfaces the better.

DanPaseltiner (commented):
Looks great. One question/suggestion below on our approach to parsing

containers/tefca-viewer/src/app/query-service.ts
m-goggins: This might be a gap in my understanding in typing but I thought you should only use `Promise<void>` when you aren't expecting to return anything of value or a significant result.

containers/tefca-viewer/src/app/query-service.ts
m-goggins: I like your suggested approach!

containers/tefca-viewer/src/app/query-service.ts
DanPaseltiner: This just clicked for me and I think what you are proposing is correct. We are saying that the keys must be members of `USES_CASES` which map to functions with the signature ` (input: UseCaseQueryRequest, queryResponse: QueryResponse, ) => Promise<QueryResponse>`.

m-goggins: Yep exactly! We're just mapping to the use case query functions, which we know return `QueryResponse`s

DanPaseltiner (approved):
LGTM!

PR #1833, opened 2024-05-13
Adding TCR to message-refiner docker-compose

### LINKED ISSUE #1822: Create docker compose for Message Refine and TCR
Let's create a docker compose file to spin up the Message Refiner and Trigger Code Reference.

Acceptance Criteria:
GIVEN this new docker compose file
WHEN `docker compose up` is run
THEN both the eCR Refiner and the TCR are spun up and networked so that the eCR refiner can make requests to the TCR.

robertandremitchell:
# PULL REQUEST

## Summary
This updates and creates a docker-compose.yml that starts up the message-refiner and trigger-code-reference services. I've carried over a lot of the checks/syntax from `orchestration`; I'm not sure if we will want/need to add additional services, so it's probably a little overkill right now, but I figured it'd be better to overcommit in the event we needed to add additional services down the line.

## Related Issue
Fixes #1822 

## Additional Information
I wasn't sure whether to give these services unique port numbers, or follow the orchestration model of having the base service (message-refiner) be 8080, then increment by 1 for the additional services.

## Checklist


containers/message-refiner/.env
DanPaseltiner, line 1: All of this networking should be handled within the docker compose file. Record linkage and orchestration offer examples of this.

containers/message-refiner/Dockerfile
```diff
@@ -9,5 +9,8 @@ COPY ./app /code/app
 COPY ./assets /code/assets
 COPY ./description.md /code/description.md
 
+ENV MESSAGE_REFINER_URL=$MESSAGE_REFINER_URL
+ENV TRIGGER_CODE_REFERENCE_URL=$TRIGGER_CODE_REFERENCE_URL
```
DanPaseltiner, line 13: As with the comment above we can set environment variables in the docker compose file. Having values for environment variables should be a requirement to run the service not build an image for it.

DanPaseltiner (changes requested):
Looks good. I think we just need to clean up how we are using environment variables (comments below). Happy to discuss further if you would like.

containers/message-refiner/Dockerfile
robertandremitchell: makes sense, updated!

containers/message-refiner/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    build:
+      context: .
+    depends_on:
+      - trigger-code-reference-service
+    ports:
+      - 8080:8080
+    logging:
+      driver: "json-file"
+    environment:
+      - MESSAGE_REFINER_URL=http://message-refiner-service:8080
```
DanPaseltiner, line 12: Does the message refiner need to know it's own URL?

containers/message-refiner/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      - TRIGGER_CODE_REFERENCE_URL=http://trigger-code-reference-service:8080
```
DanPaseltiner, line 13: I think this can be simplified to `message-refiner-service:8080`. Record linkage is a good example of this, but docker compose makes networking between containers relatively easy.

containers/message-refiner/docker-compose.yml
robertandremitchell: blah, no, updated!

containers/message-refiner/docker-compose.yml
robertandremitchell: updated!

robertandremitchell: blah, nevermind, update didn't take, one sec...

robertandremitchell: okay, should be good!

DanPaseltiner (approved):
Send it!

PR #1837, opened 2024-05-13
Implement stamp endpoint

### LINKED ISSUE #1794: Implement the `/insert-condition-extensions` endpoint

bamader:
# Stamp Those Conditions

## Summary
This PR implements the `/stamp-condition-extensions` endpoint. Using the functionality Rob built out with the SQLite Database, we now have an endpoint that accepts a FHIR bundle and a list of SNOMED codes to cross-reference, then stamps each resource in the bundle if one or more of their referenced value sets for a particular service type contains such a cross-referenced code. Lots of tests are provided.

## Related Issue
Fixes #1794 

## Additional Information
- The length of this PR comes mostly from testing assets I adapted from the streamline team's DB seeding, do not be alarmed--most of that doesn't need review
- We have a filter set up between FHIR resource types and the types of clinical services we accepted based on some profiling work I did last week
- At one point you'll see a string of like four nested for loops--DO NOT BE ALARMED lol. This is just an unfortunate result of how data is structured in both FHIR bundles (we need to iterate over entries to get resources, and then codes within each resource) and in the return value sets from the DB. Most of these loops are actually constant time and it just looks deceiving--for example, there are never more than 6 types of clinical services we get a value set for, so looping over those six service types for e.g. an Observation is just necessary with checking all value sets we get back from the DB, but will never scale up. Most of the looping is like that, and it's only the number of entries in the FHIR bundle and the number of conditions a user passes to the endpoint (which we would realistically expect to be 1 to 3) that will actually scale up this code to somewhere above linear time. Just wanted to flag that since we typically stay away from grotesque nesting.


containers/trigger-code-reference/app/main.py
```diff
@@ -23,8 +34,8 @@ async def health_check():
     return {"status": "OK"}
 
 
-@app.post("/insert-condition-extensions/")
-async def insert_condition_extensions(input: InsertConditionInput) -> Response:
+@app.post("/stamp-condition-extensions/")
```
DanPaseltiner, line 37: nit: I don't think we need the final slash on the endpoint path. This seems to have crept into a couple of our endpoint definitions recently and it doesn't add anything

containers/trigger-code-reference/app/main.py
```diff
@@ -38,7 +49,51 @@ async def insert_condition_extensions(input: InsertConditionInput) -> Response:
...
+                    if stype in stamp_checks:
+                        for code_sys_obj in stamp_checks[stype]:
+                            for rcode in r_codes:
+                                if rcode in code_sys_obj["codes"]:
+                                    should_stamp = True
+                                    break
+                            if should_stamp:
+                                break
+                    if should_stamp:
+                        break
```
DanPaseltiner, line 90: If the only place that `should_stamp` can be set to true is line 85 and break immediately after on line 86 do we need the subsequent checks and breaks on lines 87 and 89?

containers/trigger-code-reference/app/utils.py
```diff
@@ -129,3 +129,48 @@ def get_clinical_services_dict(
...
+
+    :param resource: The resource in which to locate the code.
+    :return: One or more clinical codes, as a list of strings. If the given
+      resource has no corresponding codes, an empty list is returned.
+    """
+    rtype = resource.get("resourceType")
+    codings = []
+
+    # Grab coding schemes based on resource type
+    if rtype == "Observation" or rtype == "Condition" or rtype == "DiagnosticReport":
```
DanPaseltiner, line 151: I think this could be simplified with `if rtype in ["Observation", "Condition", "DiagnosticReport"]:`

containers/trigger-code-reference/app/utils.py
```diff
@@ -129,3 +129,48 @@ def get_clinical_services_dict(
...
+    :param resource: The resource to stamp.
+    :param code: The SNOMED code to insert an extension for.
+    :return: The resource with new extension added.
+    """
+    if "extension" not in resource:
+        resource["extension"] = []
+    resource["extension"].append(
+        {"url": "http://snomed.info/sct", "coding": [{"code": code}]}
+    )
+    return resource
```
DanPaseltiner, line 176: We should use `"https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code"` as the URL and I think we should include `"system"` in addition to `"code"`. The value of system would be `"http://snomed.info/sct"` assuming `code` is a SNOMED code which at least for now we can assume.


Given how overloaded "code" could we making it clear with a combination of naming and the doc string that these are SNOMED codes that represent conditions.

DanPaseltiner (changes requested):
Generally looks really good! Couple minor suggestions, but the thing I view as truly blocking is modifying the specifics of the extension we add. Our intent here is to use the condition extension that [report stream has already implemented](https://github.com/CDCgov/prime-reportstream/blob/master/prime-router/docs/design/proposals/0023-condition-to-code-mapping/0023-condition-to-code-mapping.md)

Thanks for all your work gathering requirements for this and implementing @bamader

robertandremitchell (approved):
looks good. the nested for loop is definitely a little dizzying, but I couldn't think of another way to really do it, either.

containers/trigger-code-reference/app/main.py
bamader: Yeah, break only terminates the innermost encompassing loop, so the break on line 86 means we'll stop checking for codes we found in the resource, but we'll continue checking all of the `code_sys_objects` returned from the DB as well as all other clinical service types the current resource could get tagged by. And since we only ever need a single copy of a tag to stamp a resource, finding out that an observation could get tags from all of diagnostic, lab observation test name, and substance / organism adds nothing over the first instance. Just a little time saver.

containers/trigger-code-reference/app/main.py
```diff
@@ -38,7 +49,51 @@ async def insert_condition_extensions(input: InsertConditionInput) -> Response:
     :return: HTTP Response containing the bundle with resources extended by
       any linked conditions.
     """
-    # TODO: This method is a stub.
+    # Collate all clinical services for each of the supplied conditions to
+    # extend, collected by service type
+    stamp_codes_to_service_codes = {}
+    for cond in input.conditions:
+        cond_list = get_clinical_services_list([cond])
```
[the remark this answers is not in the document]
bamader: Lol, yeah it's hard to see a scenario where we want to call one without the other, but in fairness I was thinking about doing a dictionary inversion where we map all the possible value set clinical services as keys to the values of tuples of service type and code scheme, but I threw that out. I think what this really reveals is that taxonomy's with four mutually influential keys is messy no matter how you slice it 🤷‍♂️ .

containers/trigger-code-reference/app/main.py
DanPaseltiner: Oh right of course we have to jump out of all the loops. DUH!

bamader:
All outstanding comments should be addressed. I've also modified our testing data to include some more logical resources related to covid (a diagnosis and a positive test) and then used that in our integration testing suite to confirm we can in fact stamp resources correctly.

DanPaseltiner (approved):
🚀

PR #1850, opened 2024-05-14
Ensure namespaces are set in refined root of the `refine` function

### LINKED ISSUE #1820: Fix Namespace Bug
Currently the output of the eCR Refiner puts all elements of the refined XML in the `ns0` namespace. We don't want this. The resulting name spacing should match that of the incoming eCR document.

Acceptance Criteria

GIVEN the eCR Refiner
WHEN and eCR is refined
THEN the name spacing on the refined eCR matches the original eCR exactly

robertmitchellv:
# PULL REQUEST

## Summary
In the `refine` function for the message refiner service, we're setting the namespaces that are usually present in eICR messages for the refined root so that the output can be validated using the 1.1 Schematron. We're also using `lxml.etree.SubElement` to attach the different looped through parts to the sections they belong to so that the nesting is preserved.

## Related Issue
Fixes #1820 
- #1820 

## Additional Information

## Checklist


containers/message-refiner/tests/assets/refined_message_social_history_only.xml
```diff
@@ -1,285 +1,273 @@
...
-                            <ns0:text>Pregnancy Status-Not applicable (qualifier value)</ns0:text>
-                            <ns0:statusCode code="completed" />
-                            <ns0:effectiveTime value="20220513072522+0000" />
-                        </ns0:act>
-                    </ns0:entry>
-                </ns0:section>
-            </component>
-        </structuredBody>
-    </component>
-</ns0:ClinicalDocument>
```
DanPaseltiner, line 285: I would have expected this file and the one above to remain the same length. Could you explain what's going on here? I'm probably just missing something.

DanPaseltiner (approved):
LGTM! Nice work @robertmitchellv our XML guru 🚀

containers/message-refiner/app/main.py
```diff
@@ -119,19 +120,28 @@ def refine(validated_message: bytes, sections_to_include: str) -> str:
...
+    # ensure that the default namespaces are set correctly
+    namespace = "urn:hl7-org:v3"
+    nsmap = {
+        None: namespace,
+        "cda": namespace,
+        "sdtc": "urn:hl7-org:sdtc",
+        "xsi": "http://www.w3.org/2001/XMLSchema-instance",
+    }
+    # creating the root element with our uri namespace and nsmap
+    refined_message_root = ET.Element(f"{{{namespace}}}ClinicalDocument", nsmap=nsmap)
```
m-goggins, line 133: aha! So that's how we do it. Nice!

containers/message-refiner/tests/assets/refined_message_social_history_only.xml
m-goggins: Hmmm I would have thought that the new file would be longer, not shorter, because additional namespaces were added at the top of the document.

m-goggins (approved):
Really happy that you were able to resolve this issue! 😎

containers/message-refiner/tests/assets/refined_message_social_history_only.xml
robertmitchellv: yeah, let me dig into this a bit before merging

containers/message-refiner/app/main.py
robertmitchellv: the `ruff` lsp kept auto changing single quotes to double and this seemed to be the way to make sure it was quoted correctly. this took longer than i expected to figure out 😅

containers/message-refiner/tests/assets/refined_message_social_history_only.xml
robertmitchellv: so both of these files have the same content and tags, i went through line by line and there are a couple of things i noticed:
- in the earlier version there are instances like race and ethnicity were there are more lines used; breaks just moving to the next line where in the newer version it's all technically on one line.
- there's also a namespace added before social history in the earlier version that we don't have in the new one
- the lines and spacing for the table seem a little different from earlier version to later version in social history

i copied the output from an Insomnia response to a locally built container with these changes, which might just be formatted slightly different.

DanPaseltiner: Thanks for looking into this. Sounds reasonable to me, I think we should merge and keep moving forward here. Unless @m-goggins has concerns.

m-goggins: Sounds good to me. Thanks for taking a thorough look at this!

robertmitchellv: @DanPaseltiner @m-goggins let me know if this sounds good to you both and i can resolve and merge

PR #1855, opened 2024-05-16
Document Example Condition Stamping I/O

### LINKED ISSUE #1825: Document condition stamping endpoint
Let's addition example requests and responses to more thoroughly document the condition stamping endpoint.

bamader:
# Stamped Condition Examples

## Summary
This PR is a straightforward implementation of some example requests and responses for the `stamp-condition-extensions` endpoint. It follows the pattern established in the message-parser w.r.t. annotations, and the chosen examples are (dramatically) simplified and modified versions of some of our testing bundles, with extraneous resources deleted and COVID-related codes highlighted.

## Related Issue
Fixes #1825 

## Additional Information
Pointing at the stamp-extensions branch since this code just adds to what I did in another PR. Once we land that one this one becomes mergeable.


robertandremitchell (approved):
`/redoc` looks good to me. Only small notes.


* not needed for this PR, but not sure why the health check shows as null and not status:ok. Not the biggest deal.
* I'm guessing it's a default to include a `422` ValidationError despite one not yet existing (or being needed) for this function? I couldn't think of a scenario where it would be triggered.

PR #1864, opened 2024-05-16
Adding TCR Refinement to message-refiner

### LINKED ISSUE #1823: Implement fine grained condition-specific filtering
Based on the result of the SPIKE #1803 let's implement the fine grained filter

Acceptance Criteria:
GIVEN and eCR and condition code(s)
WHEN they POSTed to `/ecr`
THEN the condition specific filter is applied.

robertandremitchell:
# PULL REQUEST

## Summary
This would add further refinement to the message-refiner service that would look at the eCR for condition codes provided by the user. These condition codes would be looked up in the TCR service to get the relevant clinical services. XPaths will then be generated for those clinical services and then we will search each section for those relevant clinical service codes.

This also assumes the use cases are:
* A user is only interested in relevant sections, no parsing on clinical service codes (already covered)
* A user is only interested in relevant clinical services, no parsing on sections (added here)
* A user is only interested in sections containing relevant clinical service codes (added here)

To confirm, we don't need think about: a user who wants A,B,C sections, but will also be interested in D,E,F sections if D,E,F sections contain clinical service codes X,Y,Z, per conversation with Dan. We would only ever look for clinical services in sections requested.

Open question: If section A has 1,2,3 conditions, as currently written, if the only relevant clinical service code is in A2, the code as written should only return A2, but do we foresee users wanting A1 and A3 as well? Or the information around A2? For example, right now, if a condition is coded provided, right now, it returns this:

```xml
			<component>
				<code code="76078-5" codeSystem="2.16.840.1.113883.6.1" codeSystemName="loinc.org"/>
			</component>
			<component>
				<code code="76078-5" codeSystem="2.16.840.1.113883.6.1" codeSystemName="loinc.org">
					<originalText>
						<reference value="#labTestName15"/>
					</originalText>
				</code>
			</component>
```

versus the full context:
```xml
                    <entry typeCode="DRIV">
                        <organizer classCode="CLUSTER" moodCode="EVN">
                            <templateId root="2.16.840.1.113883.10.20.22.4.1" />
                            <templateId extension="2015-08-01" root="2.16.840.1.113883.10.20.22.4.1" />
                            <id root="10ea9500-d59d-4962-8a8d-81c0749f06f6" />
                            <code code="76078-5" codeSystem="2.16.840.1.113883.6.1"
                                codeSystemName="loinc.org" />
                            <statusCode code="completed" />
                            <component>
                                <observation classCode="OBS" moodCode="EVN">
                                    <templateId root="2.16.840.1.113883.10.20.22.4.2" />
                                    <templateId extension="2015-08-01"
                                        root="2.16.840.1.113883.10.20.22.4.2" />
                                    <id
                                        extension="https://fhir-ehr.cerner.com/r4/da0b5d99-573c-4f5e-aaf6-8d39fa93d033/Observation/M-19459361278/_history/1"
                                        root="2.16.840.1.113883" />
                                    <code code="76078-5" codeSystem="2.16.840.1.113883.6.1"
                                        codeSystemName="loinc.org">
                                        <originalText>
                                            <reference value="#labTestName15" />
                                        </originalText>
                                    </code>
                                    <statusCode code="completed" />
                                    <effectiveTime value="20220513074300+0000" />
                                    <value code="260415000" codeSystem="2.16.840.1.113883.6.96"
                                        codeSystemName="sct"
                                        displayName="Not detected (qualifier value)" xsi:type="CD" />
                                    <interpretationCode code="N" codeSystem="2.16.840.1.113883.5.83"
                                        codeSystemName="v3-ObservationInterpretation"
                                        displayName="Normal" />
                                </observation>
                            </component>
                        </organizer>
                    </entry>
```

## Related Issue
Fixes #1823

## Additional Information
Right now, we look at all potential codes, so if a condition is provided, there can be like 400 xpaths. This can create very long xpaths.

I initially thought we would look at each component then determine which xpath we would use. So we could try to do that or generate a whole bunch of xpaths that have these even if they're not Immunization or Observation? Or generate those additional xpaths if their clinical service type is one of the 3 immunization types only? This could still be a bit complicated/tricky to set up, and could potentially miss things.

More minor, but I also do not know how the URLs will abbreviate for two of the clinical service types. I gotta find them for these two in a sample eICR: 
        "http://www.nlm.nih.gov/research/umls/rxnorm": "?",  # TODO
        "http://hl7.org/fhir/sid/cvx": "?",  # TODO

## Checklist


containers/message-refiner/app/main.py
```diff
@@ -82,46 +98,115 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+    else:
+        xpath_expression = "//*[local-name()='section']"
     elements = validated_message.xpath(xpath_expression, namespaces=namespaces)
+    return add_root_element(header, elements)
 
-    # Create & set up a new root element for the refined XML
-    # we are using a combination of a direct namespace uri and nsmap so that we can
-    # ensure that the default namespaces are set correctly
+
+def add_root_element(header: bytes, elements: list) -> str:
```
robertmitchellv, line 201: i like breaking this out into its own function

robertmitchellv (approved):
this looks good 👍🏻  i'll need to think through the `refine` function now that there's an additional parameter that can be used with `sections_to_include` 🤔 may pick you brain on that!

robertandremitchell:
@DanielPaseltiner , updates made to break up the clinical services code and switch the TCR half to async

DanPaseltiner (changes requested):
This is really coming a long. A couple things I think we should tweak before merging. Thanks for continuing push this a long @robertandremitchell!

containers/message-refiner/app/main.py
```diff
@@ -42,15 +52,22 @@ async def refine_ecr(
...
+            return Response(content=error_message, status_code=422)
 
+    clinical_services_xpaths = None
+    if conditions_to_include:
+        clinical_services, error_message = await get_clinical_services(
+            conditions_to_include
+        )
+        clinical_services_xpaths = create_clinical_xpaths(clinical_services)
         if error_message != "":
             return Response(content=error_message, status_code=422)
```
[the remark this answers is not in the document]
robertandremitchell: updated!

containers/message-refiner/app/main.py
```diff
@@ -1,10 +1,16 @@
...
 
+import requests
 from dibbs.base_service import BaseService
 from fastapi import Request
 from fastapi import Response
 from lxml import etree as ET
 
+from app.utils import _generate_clinical_xpaths
+
+TCR_URL = "http://trigger-code-reference-service:8080"
```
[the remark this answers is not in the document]
robertandremitchell: sorry, missed this one the other day. should be updated now!

containers/message-refiner/app/main.py
```diff
@@ -82,46 +100,128 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+    sections = sections_to_include.split(",")
+    for section in sections:
+        if section not in section_LOINCs:
+            section_loincs = None
+            error_message = f"{section} is invalid. Please provide a valid section."
+            break
+        else:
+            section_loincs.append(section)
 
     return (section_loincs, error_message) if section_loincs else (None, error_message)
```
bamader, line 116: Unrelated to the PR, but I think we might be able to get rid of the ternary in this line. Because `section_loincs` is set to `None` on 110 if the section to lookup isn't in our pre-made list, and then there's a `break`, the second the program errors we'll just drop to the `return` statement and the value of `(section_loincs, error_message)` would already be equivalent.

containers/message-refiner/app/main.py
```diff
@@ -82,46 +100,128 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+    async with httpx.AsyncClient() as client:
+        for condition in conditions_list:
+            response = await client.get(TCR_ENDPOINT + condition)
+            clinical_services = response.json()
+            match response.status_code:
+                case 200:
+                    clinical_services_list.append(clinical_services)
+                case _:  # 400/422/404 cases should return empty list, errors
+                    return ([], response)
+    return (clinical_services_list, error_message)
```
bamader, line 140: Is the `error_message` we return here ever updated? Unless the `response` object somehow automatically manipulates it I can't see where this will ever be anything but the empty string.

bamader (approved):
Looks pretty sick!

containers/message-refiner/app/main.py
robertandremitchell: makes sense. I had to rewrite parts of `refine` that influenced this function, so I think it's worth simplifying. lmk what you think of the update!

containers/message-refiner/app/main.py
DanPaseltiner: Agreed @bamader

robertandremitchell: makes sense. I've refactored this step a little to not require error_message; instead, it returns the whole response, checks the status code, then will tag out to 503 if all responses are not 200. otherwise, it proceeds as usual with creating clinical_service_xpaths. lmk your thoughts!

containers/message-refiner/app/main.py
```diff
@@ -82,46 +99,128 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+    async with httpx.AsyncClient() as client:
+        for condition in conditions_list:
+            response = await client.get(TCR_ENDPOINT + condition)
+            clinical_services = response.json()
+            match response.status_code:
+                case 200:
+                    clinical_services_list.append(clinical_services)
+                case _:  # TODO: need to write 400/422/404 cases
+                    return (None, response)
+    return (clinical_services_list, error_message)
```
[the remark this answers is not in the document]
robertandremitchell: makes sense. with some of the other work I did, I think this works a bit more elegantly, but curious your thoughts.

containers/message-refiner/app/main.py
bamader: Yep, cleaner now!

containers/message-refiner/app/config.py
```diff
@@ -0,0 +1,28 @@
...
+def get_settings() -> Settings:
+    """
+    Load the values specified in the Settings class from the environment and return a
+    dictionary containing them. The dictionary is cached to reduce overhead accessing
+    these values.
+
+    :return: the specified Settings. The value of each key is read from the
+    corresponding environment variable.
+    """
+    return Settings(_env_file=".env").dict()
```
DanPaseltiner, line 28: Thank for moving this over the env var. Unfortunately, we can't use a ".env" file because when these services get deployed we need to be able to set the environment variables at the container level. I think the record linkage service is a good example of how we typical handle environment variables. For testing purposes these values should be set in the docker-compose file.

containers/message-refiner/app/main.py
```diff
@@ -82,46 +100,122 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
     error_message = ""
     section_loincs = []
```
DanPaseltiner, line 101: I think these lines can be removed.

containers/message-refiner/app/main.py
```diff
@@ -82,46 +100,122 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
-        section_loincs = []
-        sections = sections_to_include.split(",")
-        for section in sections:
-            if section not in section_LOINCs:
-                section_loincs = None
-                error_message = f"{section} is invalid. Please provide a valid section."
-                break
-            else:
-                section_loincs.append(section)
+    if sections_to_include is None:
```
DanPaseltiner, line 103: We should probably handle `None` or `""`

containers/message-refiner/app/main.py
```diff
@@ -82,46 +100,122 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+    sections = sections_to_include.split(",")
+    for section in sections:
+        if section not in section_LOINCs:
+            section_loincs = None
+            error_message = f"{section} is invalid. Please provide a valid section."
+            break
+        else:
+            section_loincs.append(section)
+
+    return (section_loincs, error_message)
```
DanPaseltiner, line 116: What about something like this? Since the function should exit whenever the if condition is met we can simply return early, eliminating the need for `break` and `else`. Furthermore, I don't think we need to set `section_loincs` to `None` in the if block. It doesn't hurt, but does it really do anything for us? Our logic in the parent function is to check the value of `error_message` and if there is an error we effectively ignore `section_loincs`.

```suggestion
    section_loincs = []
    sections = sections_to_include.split(",")
    for section in sections:
        if section not in section_LOINCs:
            error_message = f"{section} is invalid. Please provide a valid section."
            return (section_loincs, error_message)
        section_loincs.append(section)

    return (section_loincs, "")
```

containers/message-refiner/app/config.py
robertandremitchell: updated!

containers/message-refiner/app/main.py
robertandremitchell: added!

DanPaseltiner (approved):
🚀 Nice work on this!

PR #1869, opened 2024-05-19
Additional codes for Syphilis

DanPaseltiner:
# PULL REQUEST

## Summary
Updates to correct duplicate Observations and missing medications.

## Related Issue
Fixes #

## Additional Information

## Checklist

### DIFF
--- a/containers/tefca-viewer/src/app/query-service.ts
+++ b/containers/tefca-viewer/src/app/query-service.ts
@@ -344,7 +344,8 @@ async function chlamydiaQuery(
     "434692", // azithromycin 1000 MG
     "82122", // levofloxacin
     "1649987", // doxycycline hyclate 100 MG
-    "1665005", // ceftriaxone 500 MG Injection
+    "1665005", // ceftriaxone 500 MG Injection,
+    "749880",
   ];
   const serviceTypes = [
     "54", // Family planning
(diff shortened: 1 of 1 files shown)


containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -344,7 +344,8 @@ async function chlamydiaQuery(
     "434692", // azithromycin 1000 MG
     "82122", // levofloxacin
     "1649987", // doxycycline hyclate 100 MG
-    "1665005", // ceftriaxone 500 MG Injection
+    "1665005", // ceftriaxone 500 MG Injection,
+    "749880",
```
m-goggins, line 348: Based on testing, I don't think we want to add BC to the list of medications we query for, right? This had more to do with NYC pulling all obs.

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -363,11 +364,11 @@ async function chlamydiaQuery(
 
   queryResponse = await parseFhirSearch(observationResponse, queryResponse);
 
-  // Observation queries for social history
-  const socialHistoryQuery = `/Observation?subject=${patientId}&category=social-history`;
-  const socialHistoryResponse = await fhirClient.get(socialHistoryQuery);
+  // // Observation queries for social history
```
m-goggins, line 367: We also probably want to remove the social history query on gonorrhea too.

m-goggins (changes requested):
Looks good! I think we had talked about removing the birth control from rxnorm list after realizing it was an NYC specific-pull and making sure we remove the social history query for gonorrhea (untill we have a general STI query)

PR #1872, opened 2024-05-20
Add Batch TEFCA Requests 

### LINKED ISSUE #1829: Implement batch queries
**Context**: We can currently only query resources based on "AND" conditions, e.g., 

`/Observation?subject=${patientId}&code=${loincFilter}`

But we may need to query the same resource type using multiple different search parameters, e.g.,

Query for Observations by LOINC: `/Observation?subject=${patientId}&code=${loincFilter}`
Query for Observations by category: `/Observation?subject=${patientId}&category=social-history` 

It seems that the only way to implement "or" statements in FHIR queries is to use [batch querying](https://hl7.org/implement/standards/fhir/http.html#transaction). This would also speed up our processing in general and would apply to all queries. 

**Acceptance Criteria**

- [ ] Batch querying is implemented for all use cases

m-goggins:
# PULL REQUEST

## Summary
Adds `getBatch` to the `FHIRClient` to make a batch of requests and return an array of responses when possible. This seems to speed up returning data, especially when using eHealthExchange. 

## Related Issue
Fixes #1829 

## Additional Information
Tested the following scenarios:

- "GC JMC" direct and eHealth
- "Chlamydia JMC" direct and eHealth
- "Veronica Blackstone" direct and eHealth

## Checklist


containers/tefca-viewer/src/app/fhir-servers.ts
```diff
@@ -87,6 +86,20 @@ class FHIRClient {
...
+          console.error(`Failed to fetch ${url}: ${response.statusText}`);
+          return; // will return undefined if the fetch fails
+        }
+        return response.json();
+      }),
+    );
+    const results = await Promise.all(fetchPromises);
+    return results.filter((result) => result !== undefined); // Filter out undefined results
+  }
 }
```
DanPaseltiner, line 103: I think we should ensure parity in the inputs and outputs between these two methods. Personally, I'd leaning towards having the inputs be a url string and an array of urls, and the inputs be a `Promise<Response>` and `Array<Promise<Response>>>`. Currently one of these methods handles evaluating response status and parsing to JSON while the other does not. This then results in inconsistency in `parseFhirSearch`.

m-goggins: That's a good call out. I modified the functions and pulled out a helper function to ensure that we always check for response.status and handle the responses the same regardless of batching or a single response.

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -446,37 +400,58 @@ async function cancerQuery(
...
+
+/**
+ * Process the response from a FHIR search query. If the response is successful and
+ * contains data, return an array of resources that are ready to be parsed.
+ * @param response - The response from the FHIR server.
+ * @returns - The array of resources from the response.
+ */
+async function processResponse(response: fetch.Response): Promise<any[]> {
+  let resourceArray: any[] = [];
   if (response.status === 200) {
```
DanPaseltiner, line 448: Since we check the status here do we need to do so in `getBatch` as well?

m-goggins: Nope, `processResponse` is applied to both types of responses (arrays from `getBatch` and single responses from `get`) in `parseFhirSearch`, see [here](https://github.com/CDCgov/phdi/blob/tefca-batch-requests/containers/tefca-viewer/src/app/query-service.ts#L419-L426)

m-goggins: Ah yep, good call! I removed it an rechecked the test cases. I think originally I had it in there because I didn't have the status check in the parsing function but this is much cleaner.

DanPaseltiner (approved):
🚀

PR #1875, opened 2024-05-20
Boban/move evaluate table

BobanL:
# PULL REQUEST

## Summary
- Make EvaluateTable a React Component rather than a function. [Read about the difference](https://stackoverflow.com/questions/71658569/react-whats-the-difference-between-a-function-returning-jsx-and-function-compo)
- Make `caption` optional
- Fix an issue with unavailable test


nickbristow (approved):
love it

PR #1908, opened 2024-05-30
Document and cleanup `get-value-sets` endpoint

### LINKED ISSUE #1907: Create documentation for the /get-value-sets endpoint on the TCR

bamader:
# PULL REQUEST

## Summary
This PR makes a couple of modifications to the `get-value-sets` endpoint of the TCR:
1. adds sample request and response examples to the endpoint's annotated documentation; we now have examples of a regular condition query, and a query with clinical-service based filtering
2. eliminates the pesky `/` that's been creeping into our endpoints
3. changes the input signature of the `filter_clinical_services` parameter from a list to a string; in most other places in the codebase (e.g. the message refiner), we simply accept a comma-delineated string and then split to create our lists, so standardizing this input puts us in sync with that service. Moreover, while you can get fastapi to accept explicit python lists by creating and tagging those params with `Query`, in our case, we're using Query as a validation enforcement within `Annotation`, which just indicates that the value isn't part of the request's body payload for purposes of validation; trying to simultaneously use Annotation and pass a python list is causing query parsing failures. Using a string is imo the cleanest way to preserve input uniformity across services and also make things easiest on fastapi.

## Related Issue
Fixes #1907


containers/trigger-code-reference/app/main.py
```diff
@@ -139,6 +154,8 @@ async def get_value_sets_for_condition(
             status_code=422,
         )
     else:
+        if filter_clinical_services not in [None, ""]:
```
robertandremitchell, line 157: I'm not sure if we need. It won't hurt, but that should be handled here:
https://github.com/CDCgov/phdi/blob/main/containers/trigger-code-reference/app/utils.py#L125

robertandremitchell (approved):
looks good! minor quibble about the filter for clinical services, but documentation looks good on my end, no issues in testing.

containers/trigger-code-reference/app/main.py
bamader: Oh good catch! I missed the convert to list util, it'll definitely be handled there for us. The main thing that prompted me to change it was just that I couldn't make actually passing a list as part of the query params work correctly with the annotation (one or the other kept breaking), but yeah this is an easy include.

Issue #1861, opened 2024-05-16
Create example requests/responses for `refine-ecr` endpoint of `message-refiner`


bamader:

bamader:
The message refiner service will be a cornerstone of a lot of our future workflows, including coarse and fine grain filtering. However, the actual ways that we can use this endpoint are still a bit opaque. We want our open source users to be able to make requests without combing over the source code or necessarily reading individual function docstrings; instead, we want our auto-generated fastapi `/redoc` extensions to have some good working examples of the inputs and outputs of the endpoint.

The scope of this ticket is to create sample request and response `.json` asset files, load them into `app.main` of the message-refiner, and annotate the `refine_ecr` endpoint with them. You can follow the same pattern used on the `stamp-condition-extensions` endpoint of the TCR, in the orchestration service, and in the message parser for some examples on how to do this. We want these examples to be real, legitimate looking FHIR data rather than have blank fields, and we want multiple examples of using different parameters and parsing values to show the variety of output possible (e.g. one example where `sections_to_include` is `None` and another where it's an actual query string).

PR #1909, opened 2024-05-30
Adding Models, samples to ECR Refiner

### LINKED ISSUE #1906: Create documentation for the /ecr endpoint on the Message Refiner

robertandremitchell:
# PULL REQUEST

## Summary
This PR adds models for the `/ecr` request and response, along with updated sample data for the `/redoc` endpoint. It also updates the tests accordingly to use json payloads rather than urls to make the POST.

## Related Issue
Fixes #1906 , #1560, #1861

## Additional Information
I am a little confused about the best way to build the response model, where, at least for now, we want to display a formatted XML string in Insomnia/Postman, though eventually, like FHIR->PHDC, we probably would just return a `{refined_message: bytes}` in practice. So I am inclined to have that Model as-is, but curious y'all's thoughts.

## Checklist


bamader (approved):
This looks good to me. I agree with your reasoning around the response model in terms of going from formatted string to raw bytes dump, so I think I'd be good to leave this as is. Would be curious to hear from @DanPaseltiner before merging though to see if the vision should be different.

containers/message-refiner/app/models.py
```diff
@@ -0,0 +1,27 @@
...
+    refiner_input: str = Field(
+        description="The request object containing the XML input."
+    )
+    sections_to_include: str = Field(
+        None, description="The fields to include in the refined message."
+    )
+    conditions_to_include: str = Field(
+        None,
+        description="The SNOMED condition codes to use to search for relevant clinical services in the ECR.",
+    )
```
DanPaseltiner, line 19: These should continue to be query parameters. I'm not actually sure we should deviate from just putting the XML in the body of the request and this adds burden on the client to transform their data and work with a mix of JSON and XML.

containers/message-refiner/app/models.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class RefineECRResponse(BaseModel):
+    """
+    Return for the /ecr endpoint.
+    """
+
+    refined_message: str = Field(description="Refined XML as a string.")
```
DanPaseltiner, line 27: Again here I think we should just just return an XML response because we are working with XML data. What makes the API easiest for a Rhapsody developer to connect to? It may be that there isn't really a need for request and response models here as long as we can get Pydantic to validate that the they are formatted as XML.

DanPaseltiner (changes requested):
I think we are overcomplicating things by moving to JSON formatted requests and responses with serialized XML.

containers/message-refiner/app/models.py
robertandremitchell: makes sense. this update switches back to using just XML as the response and more simplified error messaging.

containers/message-refiner/app/models.py
robertandremitchell: I think that makes sense. The main issue I'm running into, however, is trying to use an kind of Model with raw XML. The only working version uses `Response`, but FastAPI does not permit annotation of `Response` for...reasons, so I'm unable to add in the examples. I've tried a few workarounds and changes, but I'm a bit stumped as to how essentially allow the raw XML string but also in /redoc produce documentation.

Right now, the optional params are documented, we just don't have a way to display example API calls or show an example XML body

DanPaseltiner: Let's review the existing docs during our 1:1 I don't know that we need to have a model per se as long as we can clearly state the input should be an eICR.

DanPaseltiner (approved):
🚀 Nice work!

PR #1919, opened 2024-06-04
Create endpoint for UAT

### LINKED ISSUE #1905: Create Postman Collection for TCR and eCR Refiner Examples

bamader:
# Postman Collection Endpoint

## Summary
This PR creates a simple `GET` endpoint in the message-refiner that returns a FileObject of the UAT Postman Collection. The FileObject is a json representation of the five requests we use for user testing of the TCR and message refiner on anonymized COVID eCR data provided by LAC. It can be imported directly into postman or insomnia.

## Related Issue
Fixes #1905


robertandremitchell (approved):
looks good to me. was able to successfully load and run tests on postman. very cool!

containers/message-refiner/app/main.py
```diff
@@ -4,13 +4,19 @@
...
 from app.config import get_settings
 from app.utils import _generate_clinical_xpaths
 
 settings = get_settings()
 TCR_ENDPOINT = f"{settings['tcr_url']}/get-value-sets/?condition_code="
+UAT_COLLECTION_FILEPATH = (
+    Path(__file__).parent.parent
+    / "assets"
+    / "Message_Refiner_UAT.postman_collection.json"
+)
```
DanPaseltiner, line 19: Not a big deal, but since we aren't using this path anywhere else does it need to be defined globally? Would it be alright to just define within `get_uat_collection`?

containers/message-refiner/app/main.py
```diff
@@ -31,6 +37,18 @@ async def health_check():
     return {"status": "OK"}
 
 
+@app.get("/uat-collection")
```
DanPaseltiner, line 40: I think the value here will likely be beyond just UAT. What do you think about making the name of the endpoint more general to something like "example-requests"?

DanPaseltiner (approved):
Looks great! Two minor comments/suggestions below nothing that's blocking.

containers/message-refiner/app/main.py
bamader: Yep that's totally fine with me. I just did it as a global in case we for some reason changed our directory structure in the future, but I don't realistically see that happening.

containers/message-refiner/app/main.py
bamader: Sure, I don't mind the name change. If we anticipate returning multiple possible sets, what do you think about including a query param specifying what type of examples to include? We could support a `uat` option now but add different cases of the query param in the future as we come up with additional things to return.

PR #1933, opened 2024-06-05
Updating Message Refiner Postman URLs and ordering, initializing documentation

robertandremitchell:
# PULL REQUEST

## Summary
This does a few updates:
1. Reorders the UAT examples.
2. Updates the URLs for the API calls.
3. (WIP) Adds documentation on the scenario and notes on how to use the API.

Step 3 will be worked on parallel with @mattmgoldberg . Actual copy can be found here: https://docs.google.com/document/d/11xOUhqRevskb9_ZJy2TfGOWq3bwOSEEOk5iX8RDIiIk/edit#heading=h.36llvp8tjlr5

## Related Issue
Fixes #

## Additional Information
Noted above, the exact copy is still subject to change. But I am curious y'all's thoughts on whether these are still the 5 tests we want and the order we want:
1. one with sections
2. one with condition
3. a demo of TCR
4. one with section and condition
5. one with a condition that does not exist in the ECR

I did not add one with multiple conditions but rather added the note about how to use multiple conditions to the second example, but I can definitely see the argument for that being an additional, or just having the second example have two conditions to start.

Also, if it is helpful, you can open this JSON, despite its name, in Insomnia and still review the docs.


m-goggins:
I combined the API notes and scenario into one section here but definitely think @mattmgoldberg should review this kind of thing since we think it may sit in front of some less technical folks. Also think there's a point missing from the Example section because there are only two items.

**Scenario**

A public health authority receives a large eCR but only wants to store and review certain sections of the eCR, such as laboratory tests and results, the patient's social history narrative, and the patient's list of reported problems. 
This first API request seeks to use the message refiner to refine an eCR based on the sections_to_include parameter. Each section of an eCR has a corresponding LOINC code that can be used in the API call. For example, the social history narrative section corresponds to LOINC code 29762-2 and the API request would include the following query parameter: `?sections_to_include=29762-2`. 
In this scenario, the user wants to include multiple sections in the refined eCR so the sections LOINCs are comma delimited. For example, to include laboratory tests and results (LOINC 30954-2) , the patient's social history narrative (LOINC 29762-2), and the patient's list of reported problems (LOINC 11450-4), the API request would include the following query parameter: ``?sections_to_include=30954-2,29762-2,11450-4`. 
The full list of eCR sections that could be included and their corresponding  LOINC codes can be found [here](https://dibbs.cloud/message-refiner/redoc#operation/refine_ecr_ecr_post).

bamader:
I generally like Marcelle's combination of things, but I would probably advise a word of caution. Personally, I think the focus on "Scenario" is actually a bit misleading. We've seen many times in the past where non-technical audiences or hyper-focused users that we've demo'ed for have become fixated on particular words or phrases that in the grand scheme of the example are completely irrelevant, to the detriment of the rest of the demo. Responses like "well, that's not how this scenario would really play out," or "well, my workflow is actually more like..." are the last things we want to get into when demo'ing functionality. Those statements have value as feedback, but imo not when they block or obstruct us showing what the tool actually does. I would suggest a language reframe to "Use Case." I think the core reason a "Scenario" is valuable is because it answers the question "why would an epi want to use this tool in terms of problems it solves and value it adds," and so I think getting away from constructing a specific artificial workflow or scenario might make the demo itself easier to digest. We could simply say things like "This call demonstrates the message refiner's ability to filter out sections that aren't of interest when investigating a particular disease. For example, an epi might use this to shrink down the size of an eCR they want to store to avoid file size limitations and make it easier to examine eCRs in the future." This is an extremely semantic point, I'll be the first to acknowledge that, but I think too many times semantics can derail an otherwise productive conversation.

mattmgoldberg:
Suggested edits:

**Use case**

This API request demonstrates the Message Refiner's ability to filter out sections of an eCR that aren't of interest to users when investigating a particular disease. As an example, an epidemiologist might want to use this call to store and review certain sections of an eCR, such as laboratory tests and results, the patient's social history narrative, and the patient's list of reported problems. 

This API call seeks to use the Message Refiner to refine an eCR based on the sections_to_include parameter. Each section of an eCR has a corresponding LOINC code that can be used in the API call. For instance, the social history narrative section corresponds to LOINC code 29762-2 and the API request would include the following query parameter: ?sections_to_include=29762-2.

The full list of eCR sections that could be included and their corresponding LOINC codes can be found [here](https://dibbs.cloud/message-refiner/redoc#operation/refine_ecr_ecr_post).

robertandremitchell:
@bamader , @robertmitchellv , @m-goggins , your and Matt's edits from the Google doc have been incorporated. I think they are best viewed in Postman by uploading this new bundle. lmk if you have any questions!

bamader (changes requested):
Looks great! All requests work for me and I'm able to read the documentation, which clarifies things handily. The only change I'd like to see is that on the "cover page" or first intro explanation of the message refiner, in the section called "How to refine eCR messages," I'd like us to add the following: "... You will also learn how clinical services related to a particular SNOMED code are queried behind the scenes using the DIBBs Trigger Code Reference (TCR) service. [new sentence starts here] Each of the five requests is made up of the request's Body, which consists of the data sent to the API, and Documentation, which explains the possible use cases of the functionality demonstrated in the test. To view the documentation for a test, click the icon of a page with a folded corner near the top of the right-hand bar." I would also personally opt for changing the title of this section to "How to use this testing suite" to better reflect the new inclusion, but I don't feel strongly about that one. I do think we need to tell potential users where to find the explanation of each test case, though.

robertandremitchell:
> Looks great! All requests work for me and I'm able to read the documentation, which clarifies things handily. The only change I'd like to see is that on the "cover page" or first intro explanation of the message refiner, in the section called "How to refine eCR messages," I'd like us to add the following: "... You will also learn how clinical services related to a particular SNOMED code are queried behind the scenes using the DIBBs Trigger Code Reference (TCR) service. [new sentence starts here] Each of the five requests is made up of the request's Body, which consists of the data sent to the API, and Documentation, which explains the possible use cases of the functionality demonstrated in the test. To view the documentation for a test, click the icon of a page with a folded corner near the top of the right-hand bar." I would also personally opt for changing the title of this section to "How to use this testing suite" to better reflect the new inclusion, but I don't feel strongly about that one. I do think we need to tell potential users where to find the explanation of each test case, though.

makes sense. this is definitely an odd button; I did a quick scan of docs to see if there was a way to just have it open automatically as a setting but unfortunately couldn't find anything.

bamader (approved):
Thanks for adding that in. Yeah I agree would be great if we could auto-open, but in the absence of that, a regular text prompt will have to do.

Issue #1935, opened 2024-06-05
Make sure condition based filtering is wrapped in its parent section


robertmitchellv:
## Action Requested
In the `refine` function `clinical_services` were being added to the eICR body outside of their parent section. In order to continue to work toward our eICR output being valid, we need to make sure that condition filtering is returned in the section they were filtered from

PR #1936, opened 2024-06-05
Add minimal sections to 'refined' eICR

### LINKED ISSUE #1884: For refined out sections of an eICR, make sure small empty sections are included in refined output
## Action Requested
During the eICR validation process it was uncovered that all of the sections are required. In practice, we've observed from production eICR messages that there are ways to include empty sections that do not contain data. So when users elects to only include specific sections in a message we need to ensure the sections we're going to refine/remove have minimal empty sections with proper `nullFlavor="NI"`.

AS a $USER, I want to ensure that refined/removed sections contain only what's required SO THAT our message refiner service does not introduce additional errors or warnings to messages

## Acceptance Criteria

**GIVEN** an eICR to be refined
**WHEN** sections are refined/removed
**THEN** minimally required tags are retained to pass validation

## Testing
Test will need to be added and adjusted for this

## Background & Context
This is important because we want to ensure that there are zero additional errors or warnings added when messages pass through the refiner.

robertmitchellv:
# PULL REQUEST

## Summary
Given the four potential outcomes based on the combination of arguments passed to the `refine` function, we are now ensuring that regardless of these different scenarios, that a minimal section is created and added to the eICR output to support validation goals. There are specific CONF# rules from the schematron that require (a cardinality of 1..1 with 'SHALL" language used) certain sections of the body to be present. This PR will make sure that these sections are created and appended.

Additionally the condition based filtering in `refine` was placing the `<entry>` that matches into the body, this PR will also wrap the matching clinical services `<entry>` in its parent `<section>`.

## Related Issue
Fixes #1884 #1935 
- #1884 
- #1935


containers/message-refiner/app/main.py
```diff
@@ -185,25 +248,13 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
-    ]
-
     if sections_to_include in [None, ""]:
         return (None, "")
 
     section_loincs = []
     sections = sections_to_include.split(",")
     for section in sections:
-        if section not in section_LOINCs:
+        if section not in SECTION_LOINCS:
```
robertandremitchell, line 257: we could just use `SECTION_DETAILS.keys()` here so that we don't have to maintain two hardcoded things for sections in the event we have to add/amend.

containers/message-refiner/app/main.py
```diff
@@ -275,31 +333,152 @@ def refine(
...
+    minimal_sections = []
+    sections_to_exclude = (
+        set(SECTION_DETAILS.keys())
+        - set(sections_to_include or [])
+        - set(sections_with_conditions or [])
+    )
+
+    for section_code in sections_to_exclude:
+        minimal_section = create_minimal_section(section_code, empty_section)
+        if minimal_section:
```
robertandremitchell, line 479: very minor, but this generates a warning that is solved by doing `is not None` here.
```bash
  /Users/rob/phdi/containers/message-refiner/app/main.py:479: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
    if minimal_section:
```

robertandremitchell (changes requested):
Overall looks great! I had a few minor notes that are more take-it-or-leave-it.

I think the only other thought is how complex `refine` is at the moment that it may make sense to break it up in some way. We talked in our pairing earlier about moving/refactoring code in `app/main.py`, so that may be a good task for next week's work, rather than tackle in this PR.

containers/message-refiner/app/main.py
```diff
@@ -20,6 +20,70 @@
...
+    "18776-5",  # plan of treatment: care plan
+    "11450-4",  # problem: reported list
+    "29299-5",  # reason for visit
+    "30954-2",  # results: diagnostic tests/laboratory data narrative
+    "29762-2",  # social history: narrative
+]
+
+# dictionary of the section's LOINC, displayName, templateId, extension, and title
+SECTION_DETAILS = {
+    "46240-8": (
```
bamader, line 37: I almost wonder if we might want to offload this whole dictionary into a separate `.json` file and then just `json.loads()` it at the top of the app. It's a fairly long constant that makes this `.py` file more cluttered imo, and it might be easier to maintain down the road. Super non-blocking, just a thought about breaking down our larger files.

containers/message-refiner/app/main.py
bamader: +1 to this. I think we can delete `SECTION_LOINCS` entirely, and maybe just move the comments about what each line represents into the actual dictionary structure of `SECTION_DETAILS`. And if we go with moving the whole dict into a separate file, we can always add another field for a plain language descriptor of what that section represents.

containers/message-refiner/app/main.py
```diff
@@ -259,14 +310,21 @@ def refine(
 
     :param validated_message: The XML input.
     :param sections_to_include: The sections to include in the refined message.
-    :param clinical_services_xpaths: clinical service XPaths to include in the
+    :param clinical_services: clinical service XPaths to include in the
```
bamader, line 313: Not a blocker, but what's the thinking behind this name change? The values are still xpaths at this point so the former name does seem more descriptively clear to me.

containers/message-refiner/app/main.py
bamader: Lol for real? `if [var_name]` seems like a completely legitimate way to check if something is `None`. I have no idea why that would become deprecated.

bamader (commented):
Great looking PR! Outside of my comments (which are non-blocking), I would definitely lean towards breaking up the `refine` function into multiple smaller helpers. I could see a world where we have one helper for each of the four combinations of section and condition filter and use those to disentangle a function that's otherwise getting very difficult to parse. But I also think I agree with Rob that refactoring can wait until a future PR.

containers/message-refiner/app/main.py
robertmitchellv: ah, i realize that i accidentally removed immunizations from `SECTION_LOINCS`; it's supposed to be there. it's not meant to be in `SECTION_DETAILS` because it is suggested but not required by the spec. we need `SECTION_DETAILS` for creating minimal sections and i can see `SECTION_LOINCS` potentially shifting more since it deals more with what kinds of sections we're allowing the API to work with. i'll fix the issue with `SECTION_LOINCS` now; thanks for catching!

containers/message-refiner/app/main.py
robertmitchellv: thanks! will add this now

containers/message-refiner/app/main.py
robertmitchellv: i like this idea; it might be a great addition to the tech debt ticket that also breaks the refiner specific code out of `main` and into its own directory as this work grows. does that sound good?

containers/message-refiner/app/main.py
robertmitchellv: oh, i just noticed that the param was named `clinical_services` but the docstring had `clinical_services_xpaths`, so i just made them the same since i noticed it

containers/message-refiner/app/main.py
bamader: Yep love it!

containers/message-refiner/app/main.py
bamader: Oh good catch! I didn't even catch that discrepancy.

containers/message-refiner/app/main.py
robertmitchellv: @bamader @robertandremitchell do you both feel like this is ok after my explanation? happy to talk more about this as well. don't want to hit resolve without making sure you're both good

robertandremitchell: Yeah, I don't think I realized there was different section LOINCs in each. I could see building out the `SECTION_DETAILS` dict in a future PR to have a `required: True/False` key so that we could try to reduce / manage these LOINC-related details in one constant, but that's probably me being a little extra with it to avoid duplication.

robertmitchellv: no, i think that's a great idea; brandon suggested a json file we read in and i thought that would be a great addition for when we break these functions out of `main` and into a `refiner` directory, which i'm going to add to the tech debt thread

robertandremitchell (approved):

PR #1944, opened 2024-06-07
Refactoring Message Refiner main.py

### LINKED ISSUE #1940: Reorganize main.py to be less cluttered
This should also begin to split up `refine` function to for each scenario of refinement (section only, section and conditions, conditions, no arg).

robertandremitchell:
# PULL REQUEST

## Summary
This moves most of the parsing code out of `app/main.py` to `app/refine.py` or `app/utils.py` to declutter the main function to just be functions associated with API calls.

`app/refine.py` hasn't otherwise really incorporated any major updates beyond pushing the code out of main.

## Related Issue
Fixes #1940

## Additional Information
I did not begin to split up refine in this ticket. Each time I started to think about it, the more it felt like each function (refine_no_params, refine_conditions_only, refine_sections_only, refine_sections_and_conditions) could be a tad duplicative, particularly with things like headers. I still think it's probably the way to go to avoid `refine` being as complex as it currently reads, but curious your thoughts on this preliminary level of breakup, and that can be done in a new PR.

I also want to look at redoing/thinking about the tests, just because so much of the code is carried over to read in the test assets where it's worth having a supplemental program that just creates our test assets to import.

## Checklist


DanPaseltiner:
I'll do a more thorough review it bit, but can we correct the name of this PR? We don't appear to be making changes the Message Parser here.

containers/message-refiner/assets/section_loincs.json
```diff
@@ -0,0 +1,83 @@
+{
+  "46240-8": {
+    "details": [
```
robertmitchellv, line 3: thanks for getting started on this! one suggestion i have is to rename `"details"` to `"minimal_fields"` or something similar. working through the trigger code `templateId`s ticket i was going to create _another_ constant and i now that i see this json file i want to add it to this structure but i think the existing name `"details"` is too vague now--especially if we continue to add other fields. and this is my fault for naming it `SECTION_DETAILS` to begin with 😅

containers/message-refiner/app/utils.py
```diff
@@ -48,3 +65,25 @@ def read_json_from_assets(filename: str) -> dict:
...
+
+    # dictionary of the required eICR sections'
+    # LOINC code, displayName, templateId, extension, and title
+    # to be used to create minimal sections and to support validation
+    section_details = {
+        loinc: details.get("details")
+        for loinc, details in loinc_json.items()
+        if details.get("required")
+    }
+    return (section_list, section_details)
```
robertmitchellv, line 89: i like this!

robertmitchellv (approved):
thanks for taking this on and getting it cleaned up ✨🧹✨ 

i'm also leaning towards _not_ breaking up `refine()` yet. with what i'm currently working on/thinking through, there will need to be some action that happens across the board regardless of the combination of parameters that get passed to it and it may change `refine()` more. either way though, i think it would be great to have a discussion about the design pattern for `refine.py` so we can continue to extend the complexity in a comfortable way

containers/message-refiner/assets/section_loincs.json
robertandremitchell: makes sense! updated

PR #1953, opened 2024-06-10
Remove PHDI from orchestration dependencies

### LINKED ISSUE #1620: Remove `phdi` from Orchestration

bamader:
# Remove PHDI as a dependency from Orchestration Service

## Summary
This PR changes the Dockerfile and the requirements.txt file of the Orchestration Service to finish cutting PHDI out of our lives.

Don't be scared by the massive number of insertions--most of those are grabbing test files that were formerly located in phdi and moving them to be directly accessible within the orchestration container. That way, neither the prod code nor the tests need any knowledge of an external package.

## Related Issue
Fixes #1620


m-goggins (approved):
🚀

robertandremitchell (approved):
lgtm! I couldn't get the ecr-viewer test to work, but the other thread made it sound like that test is a little brittle, so if it's passing on Actions, I think we're fine.

PR #1959, opened 2024-06-12
OAuth with GitHub example for ecr-viewer

### LINKED ISSUE #1920: Add Support for OAUTH
## Action Requested
Add support for OAUTH on top of the support for JWT.

## Acceptance Criteria
Access to data is blocked without proper OAUTH access.
This can be configured so we can turn this off and on.

## Background & Context
We should have a dev way of testing this locally. We may need a test oauth provider

KennethSkylight:
# PULL REQUEST

## Summary
Adds a sign in with github button to the application. By default it will be turned off in development. I will have a demo to show how it all works. I also fixed some of the issues with the middleware that was leftover from trying to get it to work in prod. 

## Related Issue
Fixes #1920


containers/ecr-viewer/src/app/login/page.tsx
```diff
@@ -0,0 +1,23 @@
+import { getServerSession } from "next-auth";
+import { LoginForm } from "./LoginForm";
+import { redirect } from "next/navigation";
+
+/**
+ * @returns a login page
+ */
+export default async function LoginPage() {
+  const session = await getServerSession();
+  console.log(session);
```
nickbristow, line 10: do we need this?

containers/ecr-viewer/src/app/page.tsx
```diff
@@ -10,6 +11,8 @@ export const dynamic = "force-dynamic";
  */
 const HomePage: React.FC = async () => {
   const listFhirData = await listEcrData();
+  const serverSession = await getServerSession();
+  console.log(serverSession?.user?.email);
```
nickbristow, line 15: Do we need this?

containers/ecr-viewer/src/middleware.ts
```diff
@@ -9,13 +10,31 @@ import { importSPKI, jwtVerify } from "jose";
...
-    set_auth_cookie(req) ?? (await authorize_api(req)) ?? NextResponse.next()
-  );
+  const isDevelopment = process.env.NODE_ENV === "development";
+  const needsAuth = !isDevelopment;
+
+  const oauthToken = await getToken({
+    req,
+    secret: process.env.NEXTAUTH_SECRET,
+  });
+  if (needsAuth) {
```
nickbristow, line 20: Can we re-write this to not be nested ifs?
```
if(!needsAuth){
return NextResponse.next();
}

rest of stuff
```

KennethSkylight: I figured this was the best way to write it to accommodate turning it off and having either nbs auth or oauth. I also can't run `set_auth_cookie` earlier because it would cause issues

containers/ecr-viewer/src/app/login/page.tsx
KennethSkylight: removed

nickbristow (approved):

PR #1962, opened 2024-06-12
Add test database cleanup pytest fixture

### LINKED ISSUE #1954: Implement pytest fixture database cleanup for orchestration's save to ecr viewer integration test

m-goggins:
# PULL REQUEST

## Summary
Adds a pytest fixture to clean up the test database after each test to make sure tests are repeatable.

## Related Issue
Fixes #1954 

## Additional Information
`psycopg2` may only be a temporary fix depending on how y'all develop the eCR viewer db; we moved away from this during the record linkage/MPI development when we transitioned to using a data access layer and using pyway for db version control. The DAL was complicated but it meant we had built a bunch of classes to interact more easily and robustly with the DB, and using pyway to handle migrations meant we could drop whole tables during testing (as opposed to just the data in the table_ and rebuild the tables as needed. I'm not sure exactly how much the eCR viewer overlaps with the MPI in terms of DB development but thought I'd put it out there for the streamline team if anyone wants to chat or the need arises in the future.

## Checklist

- [X] If this code affects the other scrum team, have they been notified? (In Slack, as reviewers, etc.)


containers/orchestration/tests/integration/test_orchestration.py
```diff
@@ -4,12 +4,29 @@
...
 from lxml import etree
 from starlette.testclient import TestClient
 
+
+@pytest.fixture
+def clean_up_db():
+    """
+    Removes the test data that is inserted during testing to ensure idempotency.
+    """
+    connection_string = os.environ.get("DATABASE_URL").replace("@db", "@localhost")
```
bamader, line 20: This should be good even in github / deployments because of the DB separation, right?

containers/orchestration/tests/integration/test_orchestration.py
```diff
@@ -163,12 +180,15 @@ def test_success_save_to_ecr_viewer(setup):
     ) as file:
         form_data = {
             "message_type": "ecr",
+            "data_type": "zip",
             "config_file_name": "sample-orchestration-config.json",
         }
+        print(file)
```
bamader, line 186: Ah ha, I have caught you testing and debugging your code 😏

bamader (commented):
Probably good to go once the print statement is cleared up; just wanted to confirm that connection string replacement will still play well with github and deployments?

containers/orchestration/tests/integration/test_orchestration.py
m-goggins: Nooooooooo

containers/orchestration/tests/integration/test_orchestration.py
m-goggins: I _think_ so. They appear to be working in github on the PR, which makes me think they will work for deployment. Do you know of  a way to test ahead of time?

bamader: Alas, I do not. The dibbs.cloud version requires a versioning deploy, but I agree that since the github actions works this probably isn't a concern. If the tests fail when trying to merge to main I think that'll also be a good signifier.

m-goggins: I'll keep an eye when merging

bamader (approved):
LGTM!

Issue #1862, opened 2024-05-16
Fix validation errors for the refined outputs


robertmitchellv:
## Action Requested
Currently for the refined output in `message-refiner/tests/assets/` there are two refined outputs that have been run through the 1.1 schematron that have fatal errors and regular errors associated with them. The output that was parsed from the `svrl` output is included below. We will need to find a way to refine while ensuring validity.

```
1

Severity:
None

Message:
Fatal: An automatically initiated eICR must contain at least one of Initial Case Report Trigger Code Lab Test Order, Initial Case Report Trigger Code Problem Observation, Initial Case Report Trigger Code Result Observation, Initial Case Report Trigger Code Medication Information, or Initial Case Report Trigger Code Immunization Medication Information. (Rule: validate_trigger_presence)

Context:
/Q{urn:hl7-org:v3}ClinicalDocument[1]

Test:
(cda:documentationOf/cda:serviceEvent/cda:code[@code='PHC1464']) or (not(cda:documentationOf/cda:serviceEvent/cda:code[@code='PHC1464']) and (//cda:observation/cda:templateId[@root='2.16.840.1.113883.10.20.15.2.3.4'] or //cda:observation/cda:templateId[@root='2.16.840.1.113883.10.20.15.2.3.3'] or //cda:observation/cda:templateId[@root='2.16.840.1.113883.10.20.15.2.3.2'] or //cda:manufacturedProduct/cda:templateId[@root='2.16.840.1.113883.10.20.15.2.3.36'] or //cda:manufacturedProduct/cda:templateId[@root='2.16.840.1.113883.10.20.15.2.3.38']))
--------------------------------------------------
2

Severity:
fatal

Message:
Fatal Error (Missing Trigger Diagnosis, Result, Lab Order, Medication Information, or Immunization Medication Information): An eICR must contain at least one of Initial Case Report Trigger Code Lab Test Order, Initial Case Report Trigger Code Problem Observation, Initial Case Report Trigger Code Result Observation, Initial Case Report Trigger Code Medication Information, Initial Case Report Trigger Code Immunization Medication Information, or Initial Case Report Manual Initiation Reason Observation. (Rule: validate_one_trigger_presence)

Context:
/Q{urn:hl7-org:v3}ClinicalDocument[1]

Test:
[...]

PR #2005, opened 2024-06-20
Preserve trigger code elements within sections

### LINKED ISSUE #1883: Retain trigger code OIDs in refined eICR output
## Action Requested
During the eICR validation process it was uncovered that we **must** retain the trigger code OIDs that are added to the various sections within the `<structuredBody>` via `templateId` tags in order for messages to be valid. Removing them results in a fatal error and at least one of these OIDs needs to be included in an eICR.

It makes sense that even if we're going to remove sections based on the `sections_to_include` that we retain these OIDs via `templateId` since there will still be a level of provenance as well as preventing fatal errors in XML validation.

AS a $USER, I want to retain OID trigger code values SO THAT fatal errors are not introduced to the refining process

## Acceptance Criteria
When can we call this done?

**GIVEN** an eICR with trigger code OIDs in its sections
**WHEN** those sections are included to be removed
**THEN** the OIDs are retained in the `templateId` tag in the empty section

## Testing
Tests will need to be added

## Background & Context
This is important to users because we want our process to **not** introduce any errors or warnings to messages from our refining process.

robertmitchellv:
# PULL REQUEST

## Summary
There are potentially four ways that an eICR can flag an `<observation>` within a section as containing the data that triggered the message to a STLT:

- Manually triggered
- Lab test order
- Problem observation
- Results observation

Each of these are nested in slightly different ways and at least one is required by the schematron to be present for an eICR to be valid. Adding to the complexity, it is possible that the `clinical_services` XPaths can match one of these observations, so we need to be mindful that we're not collecting matching observations twice. What this means for the `refine` function behaves in different ways based on the combination of parameters passed--there are four cases:

Case 1: No parameters (only `validated_message` provided)
- Check for `templateId`s.
- If found, find matching observations, clean up entries, and update text.
- If not found, create a minimal section.

Case 2: `sections_to_include` provided
- For sections in `sections_to_include`:
    - Ignore processing (leave them as they are, unprocessed).
- For all other sections:
     - Check for `templateId`s.
     - If found, find matching observations, clean up entries, and update text.
     - If not found, create a minimal section.

Case 3: `clinical_services` provided
- Check for both `templateId`s and `code`s.
- If found, find matching observations, clean up entries, and update text.
- If not found, create a minimal section.

Case 4: Both sections_to_include and clinical_services provided
- For sections in `sections_to_include`:
            - Check for both `templateId`s and `code`s.
            - If found, find matching observations, clean up entries, and update text.
            - If not found, create a minimal section.
- For all other sections:
            - Check for `templateId`s.
            - If found, find matching observations, clean up entries, and update text.
            - If not found, create a minimal section.

To handle this, we're:
- extending the `section_loincs.json` file and renaming it `refiner_details.json` to include metadata about the `templateId`s for the trigger code templates.
- transforming the response from the TCR in a different way
- leverage the `refiner_details.json` file to create instructions to guide the processing for cases 1-4 inside `refine`

## Related Issue
Fixes #1883 #1862 #2202 
- #1883 
- #1862 
- #2202 


nickbristow (approved):
This looks good to me, but i haven't worked on refiner before

m-goggins (approved):
Looks good to me! Thanks for the clear description of the 4 cases we might encounter

containers/message-refiner/app/refine.py
```diff
@@ -52,262 +66,496 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+    :return: The <section> element of the section with the given LOINC code.
     """
-    Helper function to create a minimal section element based on the LOINC section code.
+    xpath_query = f'.//hl7:section[hl7:code[@code="{code}"]]'
+    section = structured_body.xpath(xpath_query, namespaces=namespaces)
+    if section is not None and len(section) == 1:
+        return section[0]
+
+
+def _get_entries_for_section(
```
robertandremitchell, line 284: does this get used outside of tests?

containers/message-refiner/app/refine.py
```diff
@@ -52,262 +66,496 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
-    :return: A minimal section element or None if the section code is not recognized.
+    :param section: The <section> element of the section to retrieve entries from or a function that returns the section.
+    :param namespaces: The namespaces to use when searching for elements and defaults to 'hl7'.
+    :param args: Additional arguments to pass to the callable if `section` is a callable.
+    :param kwargs: Additional keyword arguments to pass to the callable if `section` is a callable.
+    :return: A list of <entry> elements of the entries in the section.
     """
-    if section_code not in SECTION_DETAILS:
-        return None
+    if callable(section):
```
robertandremitchell, line 299: I know with some of the message-parser code, we ended up rueing the args/kwargs for future visibility. Especially because of how complex this is, I wonder if we want to address now or have a tech debt ticket for how we might resolve in the future.

containers/message-refiner/app/refine.py
```diff
@@ -52,262 +66,496 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
-        "code",
-        code=section_code,
-        codeSystem="2.16.840.1.113883.6.1",
-        displayName=display_name,
-    )
-    section.append(code)
+    :param section: The <section> element of the section to retrieve entries from or a function that returns the section.
+    :param combined_xpath: this will be either code values from the TCR or templateId root values in one combined XPath
+    :param namespaces: The namespaces to use when searching for elements and defaults to 'hl7'.
+    :param
```
robertandremitchell, line 320: it looks like this didn't get complete :param: and :return:?

containers/message-refiner/app/refine.py
```diff
@@ -52,262 +66,496 @@ def validate_sections_to_include(sections_to_include: str | None) -> tuple[list,
...
+        path.append(current_element)
+        current_element = current_element.getparent()
+        if current_element is None:
+            raise ValueError("Parent <entry> element not found.")
+    path.append(current_element)  # Add the <entry> element
+    path.reverse()  # Reverse to get the path from <entry> to the given element
+    return path
+
+
+def _analyze_structure(path: List[etree.Element]) -> List[Dict[str, object]]:
```
robertandremitchell, line 398: also i didn't see where this was used outside of tests

robertandremitchell (changes requested):
hi Robert, great overview and definitely a lot to review and unpack. Overall, looks great, just had a few minor questions and notes.

containers/message-refiner/app/refine.py
robertmitchellv: ah, right, good catch! this was more of a helper function i used when trying to figure out how to get this to work and `find_path_to_entry` is the final version that is used. i'll remove this

containers/message-refiner/app/refine.py
robertmitchellv: yeah this was initially used but there was just too much iteration and looping over lists and it was simpler to search a combination of all XPaths with a section rather than a list of `<entry>`s that are individually searched. `_get_observations` was using it until i figured out how to bypass this function and simplify. i can take it out

containers/message-refiner/app/refine.py
robertmitchellv: good catch! thanks!

containers/message-refiner/app/refine.py
robertmitchellv: yeah, that's a fair point. since we're pretty sure that we'll be calling/using a function in a really specific way i suppose we don't need to have all of the functions building off of the `_get_sections` callable. these were really useful early on trying to figure out how to get things to work.

@DanielPaseltiner what are your thoughts on this one?

bamader (approved):
Phew, this was an absolute bear to get through but looks amazing! Thanks so much for all your hard work (and probably lost sleep 😴 ) on this, Robert. The only suggestion I'd have is agreement with Rob around paring down some of the kwargs/callables related to `_get_sections()`. I think the assessment that so many functions get used in such specific ways is the right one, so I'd personally opt for reducing kwarg reliance and just making the functions more direct. However, I'm also not the one who was absolutely guts deep in these trenches, so if you think there's a legitimate reason for keeping that modularity around I'm all for it. I also think we could address that in a future tech debt ticket if we just want to get this closed. Either way, amazing work!

containers/message-refiner/app/refine.py
robertmitchellv: with @bamader also agreeing i think it makes sense to make those changes; it was actually a lot simpler because i was already specifically calling them with section parameters inside `refine`. this was really a hold over from trying to figure out how to get the desired outcome and working through it interactively. thanks for the helpful comments!

robertmitchellv: @robertandremitchell do you want to take another look and make sure your changes were folded in?

robertandremitchell (approved):

PR #2018, opened 2024-06-24
Write TEFCA Viewer unit tests for formatAddress

### LINKED ISSUE #1991: Write unit tests for formatAddress
Add tests for `formatAddress` in `format-service.tsx`.

robertandremitchell:
# PULL REQUEST

## Summary
This adds unit tests for formatAddress.

## Related Issue
Fixes #1991

## Additional Information
One of the tests I wrote, if all fields are empty, was still rendering `div`s in an empty <></>. I added a second condition to the `formatAddress` function that checks whether all the text in `address` is null, in which case it should return empty, similar to if no address is provided.

## Checklist


containers/tefca-viewer/src/app/format-service.tsx
```diff
@@ -67,20 +67,31 @@ export function formatName(names: HumanName[]): string {
...
-      </>
-    );
+  // return empty if no items in address
+  if (address.length === 0) {
+    return <></>;
+  }
+
+  const addr = address[0];
+  const allFieldsEmpty = [
+    ...(addr.line || []),
```
bamader, line 77: No matter how hard I try I can't escape the props spread 😭

containers/tefca-viewer/src/app/tests/format-service.test.tsx
```diff
@@ -32,7 +33,7 @@ describe("Format Date", () => {
   });
 });
 
-describe.only("formatName", () => {
+describe("formatName", () => {
```
bamader, line 36: Curious to hear what was happening to you with `.only`. On my tests PR, any test that I _didn't_ put an `only` in front of got skipped while being run. I had to go back and add the `.only` to all instances to get them to even kick off.

bamader (commented):
LGTM. Great catch on the blank addresses rendering weird divs. The only question I have is around the use of `.only` in the testing file. On my branch, if I didn't have that, those tests would be skipped, so I actually had to add `.only` to everything to get things to run. Did you do something different here to not see that, or did that problem just not happen for you?

containers/tefca-viewer/src/app/tests/format-service.test.tsx
robertandremitchell: oh, interesting. so I could not figure out why tests were being skipped, but when I removed the .only that allowed all tests to run.

From what I read, `describe.only` is more of a debugging thing (just run the failing tests and/or tests of interest), so I don't think we want to commit them: https://jestjs.io/docs/api#describeonlyname-fn.

containers/tefca-viewer/src/app/format-service.tsx
```diff
@@ -67,20 +67,31 @@ export function formatName(names: HumanName[]): string {
...
+  if (allFieldsEmpty) {
+    return <></>;
   }
-  return formattedAddress;
+  // else return
+  return (
+    <div>
+      {addr.line?.map((line, index) => <div key={index}>{line}</div>)}
+      <div>
+        {addr.city}, {addr.state} {addr.postalCode}
```
m-goggins, line 91: One of the issues we saw with the JMC data was that their resources don't have cities, just states and zip codes so it renders in an ugly way*

image

I think we can change this to line to `{addr.city}{addr.city && ", "}{addr.state} {addr.postalCode}` to make the comma appear conditionally. It might mess up a test (or require another one).

*Pooh Bear Lane is 😆

containers/tefca-viewer/src/app/tests/format-service.test.tsx
```diff
@@ -86,3 +87,75 @@ describe.only("formatName", () => {
...
+        postalCode: "20000",
+      },
+    ];
+    const { getByText } = render(formatAddress(address));
+    expect(getByText("123 Main St")).toBeInTheDocument();
+    expect(getByText("Apt 1")).toBeInTheDocument();
+    expect(getByText("Washington, DC 20000")).toBeInTheDocument();
+  });
+
+  it("should handle missing line array gracefully", () => {
```
m-goggins, line 140: This is the test condition that might be added to or messed up with the suggestion I made above.

m-goggins (approved):
This looks good to me! I'm not as familiar with typescript testing but the tests you laid out were very clear. 

I added one suggestion for making our address formatter a little more robust when city is missing but I don't think it should hold you up and could be a separate ticket if you want.

bamader (approved):
Huh, weird. If it worked for you I'll try to see what's up on my end by taking them out.

containers/tefca-viewer/src/app/format-service.tsx
robertandremitchell: good call. I'll work on this update this morning; I agree, I wasn't sure the best way to do this without trying to check if every variable is present, which may be a way to go.

I had dreamed/hoped I had escaped the vagaries of address formatting moving out of the election space, but alas 🥲

PR #2020, opened 2024-06-24
Add Interpretation to Lab Observation Value

### LINKED ISSUE #1973: Not converting or displaying lab observation Interpretation Codes
## Action Requested
Interpretation codes are missing from our FHIR conversion and not displayed in the app. Examples below
- Lab interpretation text is not being included in the FHIR conversion. 
- **Example** for seed data `61f13d32-b25c-4fb6-998b-73de2dbf87f2`: 
- **Example** for seed data `09f39529-69b8-4937-8e01-596b7e966d87`: 
- **Relevant code**: The snippet below shows how the interpretation code text appears in the XML. If the interpretationCode text is present in an observation, it needs to be added to `valueString` (surrounded by parentheses to match the HTML text) during FHIR conversion.

>                                     <value xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
>                                         xsi:type="ST">Staphylococcus aureus</value>
>                                     <interpretationCode code="A" codeSystem="2.16.840.1.113883.5.83">
>                                         <originalText>POSITIVE</originalText>
>                                     </interpretationCode>
- **Relevant files**: `Resource/_Observation.liquid`

## Acceptance Criteria

- [ ] For a given lab observation, if an observation also has an interpretation code, it must be included in the value string.
- [ ] Test cases are updated to catch this

## Additional Information
This ticket came out of PR #1921 (scope was getting too big so we moved addressing these issues into a new ticket)

JNygaard-Skylight:
# PULL REQUEST

## Summary
Appends lab observation interpretation to  value.  Plus I was bad and greatly increased the scope of this ticket by fixing our Lighthouse scan to use a prod config of ecr-viewer.

## Related Issue
Fixes #1973 

## Additional Information

## Checklist


containers/ecr-viewer/src/middleware.ts
```diff
@@ -11,7 +11,9 @@ import { getToken } from "next-auth/jwt";
  */
 export async function middleware(req: NextRequest): Promise<NextResponse> {
   const isDevelopment = process.env.NODE_ENV === "development";
-  const needsAuth = !isDevelopment;
+  const isTest = process.env.APP_ENV === "test";
+  console.log("test: " + isTest);
```
nickbristow, line 15: probably don't need

nickbristow (approved):

containers/ecr-viewer/docker-compose.yml
```diff
@@ -25,8 +25,6 @@ services:
       dockerfile: ./Dockerfile.dev
     ports:
       - "3000:3000"
-    volumes:
-      - ./src/app:/app
```
BobanL, line 29: I think this is required for Dockerfile dev. Locally when running this docker compoose I get an error
image

BobanL (changes requested):
I think normal docker-compose is broken now.

containers/ecr-viewer/docker-compose.yml
JNygaard-Skylight: I think I fixed this, can you check this again?

BobanL (approved):
Looks good to me! 
Just as a note - I do get an error, but it seems like it is an issue with main too. It has to do with how the shared services are being bundled, which is being worked on now.

image

Issue #1758, opened 2024-05-02
Introduce Integration testing


DanPaseltiner:
Clean up (i.e., make sure the resource only contains info on Veronica Blackstone) and add the attached resource as an asset to be used during integration tests. The resource is attached to this [PR](https://github.com/CDCgov/phdi/pull/2027) (Zenhub doesn't allow you to attach JSONs)

PR #2027, opened 2024-06-25
Add TEFCA Client API

### LINKED ISSUE #1834: Implement API endpoint for machines to execute use case queries.

m-goggins:
PULL REQUEST

## Summary

**1. Adds API endpoint:**
  This PR introduces an API endpoint `api/query` that can execute the same queries as the TEFCA Viewer. To make a request, the user includes a patient resource as the body from which the patient identifiers are parsed using the `parsing-service` functions. the `use_case` and `fhir_server` are included as query_params.
  
  To use the endpoint, you need to run `npm run dev` and hit `localhost:3000/api/query?use_case=[USE_CASE]&fhir_server=[FHIR_SERVER]` with a patient resource included in the body of the request. I've been using the [veronica_blackstone_patient_resource.json](https://github.com/user-attachments/files/15980133/veronica_blackstone_patient_resource.json) (modified to include some data relevant for good old Veronica Blackstone) so that you can execute queries like:
  
  `localhost:3000/api/query?fhir_server=HELIOS Meld: Direct&use_case=syphilis`

**2. Adds the framework for integration tests:**
- Adds a healthcheck GET request and related integration test in `tests\integration`ai-query.test.ts`
- Modifies config to allow for integration tests 

**3. Creates constants file:**

- I created `constants.ts` to keep track of all the types that are used in multiple places. I'm not tied to the idea of keeping everything in one place vs. where it is applicable but in creating the API I needed to import a bunch of different types from different places so this simplified things.
- I changed the way that we build out types by first creating a `const` version that is then used to build the `type` version. This allows us to check if an incoming param matches one of the values of the const version, e.g., does the query param for use_case in the API request match the use_cases that we support?

## Related Issue
Fixes #1834 

## Additional Information
There are several additional pieces of work that are needed as follow up to this PR:
- The output of the API requests is currently the same as the output of the TEFCA viewer and needs to be modified to return a FHIR bundle. This will require modifying the `query-service` to keep track of the source of the request (`api` or `viewer`) and apply `parseFhirSearch` appropriately. -> Ticket #2029 (stubbed)
- Additional error handling for ensuring that the body of the request includes a valid patient resource -> Ticket #2030 (stubbed)
- Additional integration tests for the POST request -> Ticket #1758 

## Checklist


robertandremitchell (approved):
makes sense. tests work for me and the API call looks great. 

I think my only note/question is whether we want to add the API call as a test, as well as the Veronica Blackstone data as a test asset. but that could also just be its own ticket since we're doing all these integration/unit tests separately anyway.

containers/tefca-viewer/src/app/api/query/parsing-service.ts
```diff
@@ -0,0 +1,65 @@
...
+  dob?: string;
+  mrn?: string;
+};
+
+/**
+ * Parses a patient resource to extract identifiers.
+ * @param patient - The patient resource to parse.
+ * @returns An array of identifiers extracted from the patient resource.
+ */
+export async function parsePatientIdentifiers(
```
DanPaseltiner, line 17: I hate be that guy quibbling about function names but since `identifier` is a field within Patient resources containing an array of `Identifiers` and this function parses out values beyond just identifiers I think we should use a different name. Maybe `parsePatientDemographics`?

containers/tefca-viewer/src/app/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+export type QueryResponse = {
+  Patient?: Patient[];
+  Observation?: Observation[];
+  DiagnosticReport?: DiagnosticReport[];
+  Condition?: Condition[];
+  Encounter?: Encounter[];
+  Medication?: Medication[];
+  MedicationAdministration?: MedicationAdministration[];
+  MedicationRequest?: MedicationRequest[];
+};
```
DanPaseltiner, line 59: Query response is more of a model than a constant as compared to `UseCases` and `FhirServers`, does it really belong in this module?

containers/tefca-viewer/src/app/api/query/route.ts
```diff
@@ -0,0 +1,108 @@
...
+  } else if (
+    !Object.values(FhirServers).includes(fhir_server as FHIR_SERVERS)
+  ) {
+    return NextResponse.json(
+      {
+        message: "Invalid fhir_server. Please provide a valid fhir_server.",
+      },
+      { status: 400 },
+    );
+  }
```
DanPaseltiner, line 85: This is great. Thank you!

DanPaseltiner (approved):
This looks great. A couple small things I think worth considering, but I don't want to be a blocker here.

containers/tefca-viewer/src/app/api/query/parsing-service.ts
m-goggins: Ya, that's a better name!

containers/tefca-viewer/src/app/constants.ts
m-goggins: That's a good point. I think just started moving everything over but agree it should remain in `query-service`

m-goggins:
> makes sense. tests work for me and the API call looks great.
> 
> I think my only note/question is whether we want to add the API call as a test, as well as the Veronica Blackstone data as a test asset. but that could also just be its own ticket since we're doing all these integration/unit tests separately anyway.

@robertandremitchell I think it should be a separate ticket. If you look closely at the test data, you will see that it's pretty much just a mix of Veronica Blackstone and a sample resource I took from the HL7 website; it definitely needs some cleanup. But I can add that as a note on the ticket about integration tests.

PR #2033, opened 2024-06-26
Create unit tests for formatIdentifier

### LINKED ISSUE #1994: Write unit tests for formatIdentifier
Add tests for `formatIdentifier` in `format-service.tsx`.

bamader:
# Create Unit Tests for FormatIdentifier

## Summary
This PR creates--and this is gonna come as a surprise given the overall PR title and this here description box--unit tests for the formatIdentifier function. 

## Related Issue
Fixes #1994 

## Additional Information
Debugged an interesting case here with test robustness: because we sometimes display both an id_type and id_value separated by a colon, React parses these as separate sub-elements within the div that the function creates. It's not a problem at all for app users who visually see nothing different, but it means exact string matching in our test doesn't work because those values get split across lines, punctuation, and different DOM elements that need to be queried. There's an easy fix, and it's just to disable exact matching with the keyword dictionary. This just enables substring searching to represent all characters across the different elements, rather than forcing them to appear consecutively in a single element. Note that this parameter does _not_ make things fuzzy match. To see this yourself, if you change the third `expect` of the first `it` test to say `getByText("InternalReferenceIdentifier")` (deleting the spaces), the test will now fail. That's what we want because it means that the matches are still exactly as characters appear, it's just that we allow for matching across multiple elements.

### DIFF
--- a/containers/tefca-viewer/src/app/tests/format-service.test.tsx
+++ b/containers/tefca-viewer/src/app/tests/format-service.test.tsx
@@ -1,15 +1,16 @@
-import { render } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 import {
   formatAddress,
   formatContact,
   formatDate,
+  formatIdentifier,
   formatName,
   formatMRN,
   formatString,
 } from "@/app/format-service";
 import { Address, HumanName, ContactPoint, Identifier } from "fhir/r4";
 
-describe("Format Date", () => {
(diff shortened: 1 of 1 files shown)


m-goggins (approved):
Looks good to me! Thanks for the explanation on the edge case.

robertandremitchell (approved):
looks good!

PR #2034, opened 2024-06-26
Update TEFCA API Output to Bundle

### LINKED ISSUE #2029: Update output of the API endpoint
**Context**: The output of the API requests is currently the same as the output of the TEFCA viewer and needs to be modified to return a FHIR bundle. This will require modifying the query-service to keep track of the source of the request (api or viewer) and apply parseFhirSearch appropriately.

m-goggins:
# PULL REQUEST

## Summary
This PR updates the format of the API output to include a bundle of data. I added the `createBundle` function that puts all of the parsed resources from a `UseCaseQueryResponse` into a single `Bundle`. 

Originally, I tried to have branching logic for the way that we parse (so that we don't break apart and then put back together). It quickly became a headache to pass around the request type (api or viewer) and parse differently. With the current implementation, the queries function exactly the same whether you query from the API or the viewer. Plus, this would allow for future users to operate the queries from the viewer and still get back machine readable data as well (with say, an "export" button) that could be passed to a surveillance system; a hybrid of a manual and an automatic workflow. 

## Related Issue
Fixes #2029

## Additional Information

## Checklist


bamader (approved):
Looks great! Thanks for walking through your reasoning on moving away from passing the request type as well. Based on how clean the code turned out without that, I definitely agree. I think this is much simpler and fits the general case perfectly well.

robertandremitchell (approved):
makes sense, looks great!

tried out the two 400 cases and love the error messaging.

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -468,3 +473,30 @@ async function processResponse(response: fetch.Response): Promise<any[]> {
...
+ */
+export async function createBundle(
+  queryResponse: QueryResponse,
+): Promise<APIQueryResponse> {
+  const bundle: Bundle = {
+    resourceType: "Bundle",
+    type: "searchset",
+    total: 0,
+    entry: [],
+  };
```
DanPaseltiner, line 490: Can we use the `Bundle` class instead of hand rolling this ourselves?

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -468,3 +473,30 @@ async function processResponse(response: fetch.Response): Promise<any[]> {
...
+  Object.entries(queryResponse).forEach(([key, resources]) => {
+    if (Array.isArray(resources)) {
+      resources.forEach((resource) => {
+        bundle.entry?.push({ resource });
+        bundle.total = (bundle.total || 0) + 1;
+      });
+    }
+  });
+
+  return { Bundle: bundle };
```
DanPaseltiner, line 501: We should just return the bundle without nesting within another key as this is typically how FHIR APIs behave.

containers/tefca-viewer/src/app/api/query/route.ts
```diff
@@ -81,7 +82,7 @@ export async function POST(request: NextRequest) {
   ) {
     return NextResponse.json(
       {
-        message: "Invalid fhir_server. Please provide a valid fhir_server.",
+        message: `Invalid fhir_server. Please provide a valid fhir_server. Valid fhir_servers include ${FhirServers}.`,
       },
       { status: 400 },
     );
```
DanPaseltiner, line 88: In a future PR let's update all of these to use `OperationOutcome` resources. There is also no need to include the status code in the body of the response as it will be available in the response as a separate attribute i.e response.status

DanPaseltiner (changes requested):
This looks really great. There are a couple things I think we should consider to gradually move towards a more standard FHIR API

containers/tefca-viewer/src/app/classes.ts
```diff
@@ -0,0 +1,25 @@
...
+    type?: "searchset";
+    total?: number;
+    entry?: Array<{ resource: any }>;
+  } = {}) {
+    this.resourceType = resourceType;
+    this.type = type;
+    this.total = total;
+    this.entry = entry;
+  }
+}
```
DanPaseltiner, line 25: I think this is making things a little more complicated that needed. Apologies if I led you down this path. I was thinking would just do something like this:

```
// Create the bundle
const bundle: Bundle = {
  resourceType: 'Bundle',
  type: 'collection',
  entry: [entry]
};
```

Where bundle is import from `fhir/r4`.

containers/tefca-viewer/src/app/query-service.ts
m-goggins: Here the `Bundle` after the colon is typing the `bundle` to conform to the fhir/r4 specification of a Bundle, which is what we want

DanPaseltiner: Oh oops this is great. Sorry I goofed on this before totally my fault

DanPaseltiner (approved):

PR #2037, opened 2024-06-27
Add function to combine multiple orchestration outputs

nickbristow:
# PULL REQUEST

## Summary
This allows you to update the config to include multiple responses in your request to orchestration

## Related Issue
Fixes #

## Additional Information

## Checklist


containers/ecr-viewer/seed-scripts/sql/01-init.sql
```diff
@@ -9,13 +9,13 @@ CREATE TABLE IF NOT EXISTS fhir (
 
 CREATE TABLE IF NOT EXISTS fhir_metadata(
   ecr_id VARCHAR(200) NOT NULL,
-  data_source VARCHAR(2) NOT NULL, -- S3 or DB
-  data_link VARCHAR(500) NOT NULL, -- Link to the data
+  data_source VARCHAR(2), -- S3 or DB
```
BobanL, line 12: Is it accurate that these are nullable?

containers/orchestration/app/custom_configs/seed-ecr-viewer-config.json
```diff
@@ -16,8 +16,24 @@
...
+        "credential_manager": "azure"
+      }
     }
-  ]
+  ],
+  "outputs": [
+    "message_parser_values",
+    "fhir_bundle"
+  ],
+  "default-response": false
```
BobanL, line 38: Do we need to define this if we already have outputs?

BobanL (commented):
Just a couple of qs!

containers/ecr-viewer/seed-scripts/sql/01-init.sql
nickbristow: this can go back to not null, it just broke when i was getting it working

containers/orchestration/app/custom_configs/seed-ecr-viewer-config.json
nickbristow: So in order to not break anything currently relying on it, I put in default-response, so you can tell it not to include stuff outside `responses`. 

`outputs` puts the responses in a key of `responses`.

The default response, is just in the top level

BobanL (approved):
🏖️

PR #2039, opened 2024-06-28
parse RCKMS docs for trigger code reference data

ericbuckley:
# PULL REQUEST

## Summary
Adding the first part of a script to load RCKMS CodeSet data into a database for the trigger-code-reference API to use.  This PR focuses on the first of 3 steps, parsing the ValueSet data out of RCKMS reporting specification word documents.

## Related Issue
Fixes #

## Additional Information
The next two steps of completing the script will entail
a) Querying the NIH UMLS API for CodeSets using the OID ValueSet information parsed
b) Loading a SQlite database with the values obtained

## Checklist


containers/trigger-code-reference/seed-scripts/test_seed_rckms_database.py
```diff
@@ -0,0 +1,198 @@
+"""
```
robertandremitchell, line 1: is there a reason this can't be in the test folder directly? If it's because of the need to access the rckms database, we could put an `__init__.py` in the `seed-scripts` so that it could be imported more seamlessly in tests. I think we should try to keep all tests together, if possible, or failing that, create a `tests/` subdirectory here to pick up on tests.

robertandremitchell (changes requested):
mostly makes sense, just a note about the test structure, but overall makes sense.

containers/trigger-code-reference/seed-scripts/test_seed_rckms_database.py
ericbuckley: Great question.  I don't think adding a duner init to seed-scripts will work, as `seed-scripts` is not a [valid python module name](https://peps.python.org/pep-0008/#package-and-module-names).  You could rename `seed-scripts` to `seed_scripts`, and it should work then (something I briefly tried in one of my commits).  However, I noticed there is a larger convention throughout the repo where the "seed-scripts" directory format is used in other containers, so I rolled back that change to stick with the existing convention.

robertandremitchell: ah yeah, makes sense. I think I've be in favor of just renaming to `seed_scripts/` or putting the test script into a `seed-scripts/test` folder, but I don't feel that strongly toward either option. I'll defer to @DanielPaseltiner on if there's a better way to handle this.

ericbuckley: renaming to `seed_scripts/` is an option.  

However, moving the test file to `seed-scripts/test` will also fail.  The test script needs to import the function from the `seed_rckms_database` module, if that module still resides within the `seed-scripts` directory, the import will fail because the statement `from .. import seed_rckms_database` will fail to resolve.

ericbuckley:
canceling this request in favor of a newer one that has more functionality

PR #2059, opened 2024-07-03
Add Loading Skeleton to Loading Screen

KennethSkylight:
# PULL REQUEST

## Summary
Adding the loading state to use the ecr loading skeleton. After researching how loading.js works, that mainly applies to server components which we are not using. For client components, state management is the way to go.


BobanL (approved):
Looks great!

PR #2118, opened 2024-07-11
adding a function to load the RCKMS data into a sqlite DB

ericbuckley:
# PULL REQUEST

## Summary
Updating the script to use argparse for easier CLI argument parsing and adding a new function to load the condition/valueset/codeset data into the database.

## Additional Information
There is an ongoing discussion about changing the database schema, I didn't attempt to make any of those changes in this PR.  The objective here was to load into the current schema and prove that we have all the data elements necessary for that.  

## Checklist


robertandremitchell (commented):
makes sense. Is there supposed to be migration scripts in this PR?

ericbuckley:
> makes sense. Is there supposed to be migration scripts in this PR?

This works with the existing [migrations](https://github.com/CDCgov/phdi/tree/main/containers/trigger-code-reference/seed-scripts/migrations), so you should be able to run

```sh
python seed_rckms_db.py archive.zip rckms.db --migrations migrations/
```

robertandremitchell:
> > makes sense. Is there supposed to be migration scripts in this PR?
> 
> This works with the existing [migrations](https://github.com/CDCgov/phdi/tree/main/containers/trigger-code-reference/seed-scripts/migrations), so you should be able to run
> 
> ```shell
> python seed_rckms_db.py archive.zip rckms.db --migrations migrations/
> ```

ah okay, sounds good. I'll check in with Dan for next sprint on updating the migrations with the new ERD, but this looks good!

robertandremitchell (approved):

ericbuckley:
> > > makes sense. Is there supposed to be migration scripts in this PR?
> > 
> > 
> > This works with the existing [migrations](https://github.com/CDCgov/phdi/tree/main/containers/trigger-code-reference/seed-scripts/migrations), so you should be able to run
> > ```shell
> > python seed_rckms_db.py archive.zip rckms.db --migrations migrations/
> > ```
> 
> ah okay, sounds good. I'll check in with Dan for next sprint on updating the migrations with the new ERD, but this looks good!

Cool.  I'd love to be included in that conversation, I have some reservations about reusing the same schema for both ersd and rckms data considering all the M2M relationships.  I was thinking storing the data in multiple databases or using namespaced tables, would make it much easier to a) periodically update our mappings and b) attribute any mapping to a source.

PR #2122, opened 2024-07-12
Updating createNewRelease to have workflow_dispatch

### LINKED ISSUE #2019: Address eCR Viewer Build failure
Currently the eCR Viewer is failing to build during the release process. Depending on the order that building blocks get scheduled for builds this can block the release of other DIBBs services.

robertandremitchell:
# PULL REQUEST

## Summary
This PR is making two changes:
1. Adding a `workflow_dispatch` to the createNewRelease.yaml in order to be able to test the release YAML without cutting a new release to resolve
2. a potential update to fix this issue: https://github.com/CDCgov/phdi/actions/runs/9908733094/job/27375565577

Long-term, this might allow us to test a new release and check what breaks a bit more easily, as well.

## Related Issue
Fixes #2019 , or at least, future-proofs

## Additional Information

## Checklist


.github/workflows/createNewRelease.yaml
```diff
@@ -1,5 +1,6 @@
 name: Create New Release
 on:
+  workflow_dispatch:
```
BobanL, line 3: Will this pass since we check the commit message in the first job?

BobanL (approved):
I'm ok to test this out!

.github/workflows/createNewRelease.yaml
robertandremitchell: so my thought, at least in the short run, was that this would only be available for RELEASE PRs, and that this would allow us to run github actions on those PRs. So the next release PR should allow us to confirm if that's actually true.

PR #2127, opened 2024-07-15
save bundle locally when running create-seed-data

BobanL:
# PULL REQUEST

## Summary
- Save fhir bundle to folder for easy viewing
- Add the bundle.json to gitignore to prevent it from being saved


lina-roth (approved):
LGTM!

PR #2146, opened 2024-07-16
Change location identifier to use extension rather than root

### LINKED ISSUE #2092: Update IDs across eCR Viewer to use eCR data (not FHIR UUIDs)
## Action Requested
Currently, the IDs present in the eCR Viewer show the FHIR UUID values rather than the ID values that came in with the eCR directly. The FHIR UUIDs have no meaning to our users, but the native ID values do. The ask is to update the fhirpaths for eCR Viewer to reference the eCR ID value, not the FHIR ID value.

## Acceptance Criteria
* All IDs across the eCR Viewer reference the value present from the original eCR
    * Patient ID is known to need an update, but there may be others - part of this ticket will be looking through the Viewer to determine what other ID fields are present and if they need to be updated.
* No FHIR UUIDs are shown to end users

## Testing
Manual testing across several eCRs will be required to ensure the updated FHIR paths work correctly.

## Designs
No designs, but here is an example of what is being shown/what should be shown:

## Background & Context
As a case investigator, I want the IDs present in an eCR to use data from the native eCR (not the FHIR UUIds) so that they are meaningful data.

## Additional Information
The IDs _should_ be converted from FHIR correctly, and we just need to update the FHIR paths that are referenced in the eCR Viewer. But this needs to be confirmed with a few sample eCRs. 

One example from the FHIR bundle:
```
“identifier”: [
          {
            “value”: “urn:uuid:8589d447-a267-11ec-81e0-9e7eef92c510”,
            “system”: “urn:ietf:rfc:3986”
          },
          {
            “value”: “0000016868”,
            “system”: “urn:oid:1.3.6.1.4.1.24310”
          }
        ],
```
We should be extracting the second `value` (where the `system` references an OID) rather than the first value out of the `identifier`.

BobanL:
# PULL REQUEST

## Summary
- Replace Location [IdentifierRoot](https://github.com/CDCgov/phdi/blob/1bb427275135f94225b267b7c969ad4f30163276/containers/fhir-converter/Templates/eCR/DataType/_IdentifierRoot.liquid#L4) with Location [Identifier](https://github.com/CDCgov/phdi/blob/1bb427275135f94225b267b7c969ad4f30163276/containers/fhir-converter/Templates/eCR/DataType/_Identifier.liquid)

## Related Issue
Fixes #2092

## Demo
image


containers/ecr-viewer/src/app/services/evaluateFhirDataService.ts
```diff
@@ -505,3 +496,24 @@ export const evaluateIdentifiers = (fhirBundle: Bundle, path: string) => {
...
+) => {
+  const encounterLocationRef =
+    evaluate(fhirBundle, mappings.facilityLocation)?.[0] ?? "";
+  const location: Location = evaluateReference(
+    fhirBundle,
+    mappings,
+    encounterLocationRef,
+  );
+
+  return location?.identifier?.[0].value;
```
angelathe, line 518: Confirming: Will it always be the value of the first object?

This may just be a data quality issue, but for 1.2.840.114350.1.13.478.3.7.8.688883.230886 (seed data that starts with 61f), it has the OID as the "value" instead of the "system."
```
        "identifier": [
          {
            "value": "1.2.840.114350.1.13.478.3.7.2.686980"
          }
        ],
```

BobanL: > Confirming: Will it always be the value of the first object?

I'm keeping with the existing assumption that it will be the first one since the rest of facility information is coming from the first one

> This may just be a data quality issue, but for 1.2.840.114350.1.13.478.3.7.8.688883.230886 (seed data that starts with 61f), it has the OID as the "value" instead of the "system."

Is it possible that it is still old data you are looking at? I am looking a fresh db build and see the correct value. 
Now since we are using [`Identifier`](https://github.com/CDCgov/phdi/blob/1bb427275135f94225b267b7c969ad4f30163276/containers/fhir-converter/Templates/eCR/DataType/_Identifier.liquid) rather than [`IdentifierRoot`](https://github.com/CDCgov/phdi/blob/1bb427275135f94225b267b7c969ad4f30163276/containers/fhir-converter/Templates/eCR/DataType/_IdentifierRoot.liquid#L4) it should always display the `extension` value in that place.


image
image

angelathe: Yup looking at old data 🤦‍♀️🤦‍♀️ This looks good then!!

angelathe (approved):

PR #2162, opened 2024-07-17
Implement self-service demo mode

### LINKED ISSUE #2134: Implement Self-Service Demo Mode

m-goggins:
# PULL REQUEST

## Summary
This PR introduces a drop down menu on the search page that allows users to select from a list of demo patients. It fills in their appropriate data, allowing for an easier self-serve experience. 

It also removes the restriction on MRN only containing numbers as some of the synthetic data does have other character types. I used just one patient example per use case, but can add others if we want (e.g., multiple patients are surface scenario or a query to the Cerner Helios server). I also tended to opt for the `HELIOS Meld: Direct` server but the newborn patient example uses eHealth.

## Related Issue
Fixes #2134 

## Additional Information
There isn't a way to cleanly remove the filled data once it is there other than refreshing the page or selecting another demo patient. I'm not sure that this will be a problem for actual users who likely _only_ want to use the demo data but thought I'd point it out in case others feel strongly about it and want to add a ticket to the backlog for a 'clear fields' button, for example.

## Checklist


bamader (approved):
This looks good to me! Was able to get it running smoothly locally and all the options behaved well when clicking through. I like the different background color for auto population as well.

PR #2177, opened 2024-07-18
Create query workflow e2e

### LINKED ISSUE #2176: Create e2e for tefca viewer query workflow

bamader:
# Create end to end test for basic TEFCA query

## Summary
This PR creates an end to end test suite that walks through a full browser stack of a couple basic TEFCA scenarios:
(1) it makes sure the landing page loads appropriately before doing anything else
(2) it runs through a query that successfully returns a single patient and verifies that the elements for that patient are what we expect
(3) it runs through a query that doesn't find any results and confirms that the user is told as much.

## Related Issue
Fixes #2176


m-goggins (approved):
This testing suite looks really awesome! 🚀

PR #2217, opened 2024-07-22
Updating TEFCA landing page UI

robertandremitchell:
# PULL REQUEST

## Summary
Updates the landing page to match the UI, functionality proposed here: https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=2636-10410&t=YrOFBTt94cTW3nm9-0

## Related Issue
Fixes #2217

## Additional Information
I set up the demo buttons, but I fell down a rabbit hole of trying to update the `/query` page a little yesterday. However, because of the UI changes there, I decided instead to focus on the infrastructure aspect, so I will be curious if we think that code looks okay for the proposed change (i.e., adding the prefill option to the `/query` URL, then that ticket will pick up work on loading that to the prefill).

I think my only other question is that right now if you select nothing, it just redirects to `/query` without any prefiil. This felt like the right user experience expectation, but I wasn't sure if we would want to make it mandatory to select one of the demo options.

## Checklist


m-goggins (approved):
Looks good to me!

PR #2220, opened 2024-07-23
Add error handling to check for valid patient resources

### LINKED ISSUE #2030: Add error handling to check for valid patient resource in API
Context: the `api/query` endpoint [currently](https://github.com/CDCgov/phdi/blob/main/containers/tefca-viewer/src/app/api/query/route.ts#L34) accepts a patient resource as the body of a POST request and parses patient demographics from the resource to make the query. We need to add some error handling to check whether the patient resource is a valid resource. 

I haven't looked into whether there is an existing resource we could use or possibly check if the `requestBody` has the expected `fhir/r4` elements that we parse in [`parsePatientDemographics`](https://github.com/CDCgov/phdi/blob/main/containers/tefca-viewer/src/app/api/query/parsing-service.ts).

Acceptance Criteria:

- [x] The `requestBody` is checked to be a valid Patient resource 
- [x] If there is an error, the user sees an error message that explains the patient resource is not valid
- [x] The TODO for creating additional error handling for this issue is removed the `route.ts` file

m-goggins:
# PULL REQUEST

## Summary
The `api/query` endpoint accepts a patient resource as the body of a POST request and parses patient demographics from the resource to make the query. This PR adds a check in the API route to ensure that the body of an API request contains a Patient resource by checking the `resourceType`.

 It also adds some tests to ensure the other errors are being handled appropriately. 

## Related Issue
Fixes #2030 

## Additional Information
I'm not completely happy that the only check is the `resourceType` but I went down a small rabbit hole (a shallow rabbit ditch?) trying to figure out if we could leverage the `Patient` type from `fhir/r4` as a way to validate the resource. Alas, it does not seem feasible without re-creating a Patient resource schema that we store (and is not completely tied to the `type`) using a package like `superstruct`. tl;dr: I think this solution is good enough for now.

## Checklist


DanPaseltiner (approved):
LGTM. Glad to see the OperationOutcome resources.

PR #2223, opened 2024-07-24
Adding docker compose down to playwright tests

robertandremitchell:
# PULL REQUEST

## Summary
This adds a global teardown to playwright to ensure docker is closed: https://playwright.dev/docs/test-global-setup-teardown.

I wanted to add this because right now you have to close out docker each time you run playwright locally.

## Related Issue
Fixes #n/a

## Additional Information
Should we also add a setup script, given that we are gonna add a database that needs to spin up, etc.? That work can also probably be held off.

In the shorter run, though, I would if we want the webserver command to be `docker compose up --build`. When running tests locally, I didn't realize it was working off a stale image, so tests passed locally because it didn't look at any of my changes. I'm of two minds because it will make running the tests locally slower. Curious others' thoughts.

Also, has anyone else ran into this `ingestion` issue on `npm ci`??? I've tried to install and uninstall locally every way I can think of to get a new package.json/package-lock.json. It passes locally, but keeps throwing me this error and I cannot replicate it locally...

I think it's this issue:

```bash
rob@Roberts-MacBook-Pro tefca-viewer % npm audit
# npm audit report

lodash.set  *
Severity: high
Prototype Pollution in lodash - https://github.com/advisories/GHSA-p6mc-m468-83gw
No fix available
node_modules/lodash.set
  archetype  >=0.7.0
  Depends on vulnerable versions of lodash.set
  Depends on vulnerable versions of mpath
  node_modules/archetype
    ingestion  *
    Depends on vulnerable versions of archetype
    node_modules/ingestion

mpath  <0.8.4
Severity: moderate
Type confusion in mpath - https://github.com/advisories/GHSA-p92x-r36w-9395
No fix available
node_modules/mpath

4 vulnerabilities (1 moderate, 3 high)

Some issues need review, and may require choosing
a different dependency.

```

## Checklist


bamader (commented):
The code itself looks totally fine, but why is the `package-lock.json` changing so much? Do we need that `ingestion` dependency include in the tefca-viewer's `package.json`?

robertandremitchell:
> The code itself looks totally fine, but that level of change in `package-lock.json` worries me. I'm not seeing those ingestion issues at all, but it looks related to what Nick posted about this morning. From a pedagogical standpoint, I'm not sure the TEFCA viewer _should_ need to know about the ingestion service, so I hesitate to include it as a dependency here.
> 
> As far as the webserver command, I do see what you're saying about the stale image. That's unfortunate but we could always do a `docker compose build && docker compose up`, although I guess if the image is already built that might not work. I do think we should try to work out a way to make the tests run against a version with the changes incorporated. If you rebuild the image before you run the tests, it'll actually take those changes into account, but that might defeat the purpose. In either case, it sounds like I'm in favor of some kind of additional setup script.

totally agree re: the package-lock situation. I _think_ I fixed it, not sure why it was such a mess with `ingestion`, but it appears being on `1.0.0` does not cause the same issue. Overall changes are far more modest.

Yeah, I'm of two minds regarding setup. We will 100% need it eventually once TEFCA incorporates a DB and other services. And GitHub Actions ultimately still catches whether someone ran the playwright tests on a stale image. I can add a backlog ticket to further consider setup since that one is a little less annoying imo.

bamader (approved):
Cool, this looks much better. Thanks for stripping out that ingestion mess, this is a lot more manageable. I like your suggestion of writing up an additional setup consideration ticket--for the time being, we can continue to just rebuild locally before running the tests (minor annoyance but you can rebuild with caching to grab layers that didn't get changed, takes about 8 seconds on my machine) and then down the line we can make things work better. For now, though, LGTM!

PR #2232, opened 2024-07-24
Handle multiple phone number formats in search

### LINKED ISSUE #1840: Implement phone number search

bamader:
# PULL REQUEST

## Summary
This PR handles the MVP of phone number formatting for our query search; the phone number passed in the search field gets stripped to just be digits, and then based on some profiling of the most common formatting schemes present in HAPI, Epic, and Cerner, we spit back out an array of possible formattings with some combinations of spaces, dashes, and parentheticals. These all get logically `OR`'ed in the query URL search proper.

## Related Issue
Fixes #1840 

## Additional Information
The idea for this approach came out of some conversations with Dan, but the idea is to basically take a much wider shotgun-style approach to hitting phone numbers, since the big FHIR servers all enforce _exact only_ matching on the telecom field (that means we need to perfectly hit how they distribute all their delimiters, parentheses, etc.). We're making the choice right now to kind of ignore country code parsing--eventually, we'll want to have a separate country code field that starts off with a default value of 1 (for US numbers) because then we can more explicitly handle country digits. If we tried to handle it now, we run into some very hairy and not-worth-it-to-sort-out cases. For example, there's no way to tell if a number with 10 digits includes a country code and omits local-area routing, or if it's some other partial phone number, etc. So we're just gonna ignore it for this PR. Later on though, we can also use the country code text input field for switch-statement logic to check different formatting options more common to international numbers. That code will just have to come after we have a country code box.

### DIFF
--- a/containers/tefca-viewer/src/app/query/components/SearchForm.tsx
+++ b/containers/tefca-viewer/src/app/query/components/SearchForm.tsx
@@ -20,6 +20,28 @@ import {
 } from "../../query-service";
 import { Mode } from "../page";
 
+/**
+ * @todo Add country code form box on search form
+ * @todo Update documentation here once that's done to reflect switch
+ * logic of country code
+ *
+ * Helper function to strip out non-digit characters from a phone number
+ * entered into the search tool. Right now, we're not going to worry about
+ * country code or international numbers, so we'll just make an MVP
+ * assumption of 10 regular digits.
+ * @param givenPhone The original phone number the user typed.
+ * @returns The phone number as a pure digit string. If the cleaned number
+ * is fewer than 10 digits, just return the original.
+ */
+export function FormatPhoneAsDigits(givenPhone: string) {
+  // Start by getting rid of all existing separators for a clean slate
+  const newPhone: string = givenPhone.replace(/\D/g, "");
+  if (newPhone.length != 10) {
+    return givenPhone;
+  }
+  return newPhone;
+}
+
 interface SearchFormProps {
   setOriginalRequest: (originalRequest: UseCaseQueryRequest) => void;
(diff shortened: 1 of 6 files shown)


containers/tefca-viewer/src/app/query/components/SearchForm.tsx
```diff
@@ -21,6 +21,28 @@ import {
...
+ */
+export function FormatPhoneAsDigits(givenPhone: string) {
+  // Start by getting rid of all existing separators for a clean slate
+  const newPhone: string = givenPhone.replace(/\D/g, "");
+  if (newPhone.length != 10) {
+    return givenPhone;
+  }
+  return newPhone;
+}
+
```
DanPaseltiner, line 45: I think we should move this functionality into the query service of utils so that it is available for use on the API side of things as well.

DanPaseltiner (changes requested):
Looks really great, but we need to make sure that we can handle phone numbers when provided via either the UI or API.

containers/tefca-viewer/src/app/query/components/SearchForm.tsx
bamader: Oh duh, I was so focused on the different formats I forgot the API side. I'll get it sorted.

DanPaseltiner: No worries! I was starring at this for a while before I realized what was bugging me about it. We are building two interfaces that expose the same functionality so we just need to keep that in mind with structuring the code.

robertandremitchell (approved):
looks good!

DanPaseltiner (approved):

PR #2238, opened 2024-07-25
TCRS now finds Conditions on its own.

### LINKED ISSUE #2123: Update endpoint in the Trigger Code Reference service
## Action Requested
We need to make some additions in the Trigger Code Reference building block to better suit our needs. We need to update the endpoint so rather than sending a list of codes to stamp, it automatically looks inside in the bundle for the reportable condition codes from the Reportability Response (RR), then stamps the eCR with those conditions.

## Acceptance Criteria

In the endpoint:
- [ ]  The RR condition codes are extracted from the FHIR bundle
- [ ]  The TCR then stamps the bundle with each of those codes
- [ ]  The stamped bundle is returned in the response from the endpoint
- [ ]  It no longer needs/accepts a list of conditions sent it alongside the FHIR bundle
- [ ]  Unit tests for all functions

## Testing
Yes, this ticket requires unit tests for new functions. No integration tests yet as that will come in in the next ticket.

## Designs
No as this is entirely backend

## Background & Context
The current stamping endpoint on the TCR isn't ideal for us as it requires us to pass in a list of conditions to the endpoint. We don't want to do this code gathering and passing in orchestration because it adds complexity and makes orchestration "too smart" when it should just be passing around bundles and calling endpoints. This change will  let us do all the work in the TCR to keep orchestration simple.

JNygaard-Skylight:
# PULL REQUEST

## Summary
Instead of requiring a given list of SNOMED condition codes the Trigger Code Reference Service now finds all present in the given bundle and stamps additional data for what it found.

## Related Issue
Fixes #2123 

## Additional Information

## Checklist
This can be confirmed by:
1) Set up the TCRS following the instructions in it's `description.md`
2) Using your favourite method for making API calls  send a post command to `/stamp-condition-extensions` with the following body:
```
{
    "bundle": <JSON of a FHIR bundle, I used the bundle with ID: "1.2.840.114350.1.13.478.3.7.8.688883.230886">
}
```
3) Compare the returned JSON with the original one. For the bundle I used, three conditions were found: COVID, Hepatitis, and plague.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,6 +10,19 @@
       "cwd": "${workspaceFolder}/containers/orchestration",
       "justMyCode": false,
       "python": "${workspaceFolder}/containers/orchestration/.venv/bin/python"
+    },
+    {
+      "name": "Python Debugger: Trigger Code Reference",
+      "type": "debugpy",
+      "request": "launch",
+      "program": "${workspaceFolder}/containers/trigger-code-reference/app/main.py",
+      "console": "integratedTerminal",
+      "cwd": "${workspaceFolder}/containers/trigger-code-reference",
+      "justMyCode": false,
+      "python": "${workspaceFolder}/containers/trigger-code-reference/.venv/bin/python",
+      "env": {
+        "PYTHONPATH": "${workspaceFolder}/containers/trigger-code-reference"
+      }
     }
   ]
 }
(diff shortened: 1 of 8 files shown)


BobanL (approved):
Just tested this locally and got Hep C stamped! Thanks for you work on this!

PR #2281, opened 2024-07-30
Updating TCR ERD

### LINKED ISSUE #2224: Migrate TCR to new schema

robertandremitchell:
# PULL REQUEST

## Summary
This change adopts the ERD here: https://drive.google.com/file/d/1nxm0HpiJsONizM5NzvSIH2dbUCcDif6f/view
for the eRSD table, then updates the code in utils and main accordingly. It also retires the term `clinical service` in favor of `concepts` to better match the vocabulary used in other data sources.

This ERD will then be used as the basis for the "super" database for eRSD, RCKMS, and any other potential data sources.

The code also implements the VSAC FHIR API to get the version for ValueSet and Conditions.

## Related Issue
Fixes #2224

## Additional Information
* I did take a peek at the message-refiner postman, which is the only other documentation place i'm aware of that mentions TCR. That documentation is still good, but I did want to tag a member of the streamline team in case it impacts any documentation or use-cases for `ecr-viewer`.
* This also had an issue with dev-requirements, where I removed the testcontainers versioning. I did it based on this warning I was getting `WARNING: testcontainers 3.7.1 does not provide the extra 'compose'`. 
* More minor, but with the ~1500 API calls we now make to VSAC, it's a lot slower, like 5 minutes to rebuild the database. Because this is not something we would run often, it's probably fine, but if we want/need to eventually optimize, we might need to either async or better parallelize the API calls. I did try to write the API calls such that we avoid duplicative calls on ValueSets or conditions.

## Checklist


containers/trigger-code-reference/seed-scripts/migrations/V02_01__create_tables_add_indexes.sql
```diff
@@ -0,0 +1,68 @@
+-- create tables to link conditions to clinical services
+CREATE TABLE IF NOT EXISTS valueset_types (
```
ericbuckley, line 2: @robertandremitchell in this new schema, how do we distinguish between valuesets sourced from ersd vs rckms?

robertandremitchell: good question. my initial thought was that the `author` would be different, but I don't think that's actually true; I believe both sources will have `CSTE Steward`, or similar language that will make it not useful as a distinguishing. 

@DanielPaseltiner , curious your thoughts; we could add a `provenance` column to this and maybe all tables without a huge fuss to the code. But to your point in the past, at least the two sources we are planning on now (RCKMS and eRSD) are in theory both sourced from APHL.

DanPaseltiner: > good question. my initial thought was that the `author` would be different, but I don't think that's actually true; I believe both sources will have `CSTE Steward`, or similar language that will make it not useful as a distinguishing.
> 
> @DanielPaseltiner , curious your thoughts; we could add a `provenance` column to this and maybe all tables without a huge fuss to the code. But to your point in the past, at least the two sources we are planning on now (RCKMS and eRSD) are in theory both sourced from APHL.

The first thing I think we need to clarify is some out language. Our two sources of data are the eRSD and the reporting specification documents. Both of these artifacts are produced by APHL to support RCKMS. I realize it is a little bit pedantic but referring to these two sources as eRSD and RCKMS doesn't make much sense.

Second @ericbuckley I think you raise a good point. From the standpoint of tracking provenance we want to know two things.

1. Who produced a given value set i.e. `CSTE Steward`. This is the organization that collected a whole bunch of concepts into a single group for one particular purpose.
2. Which resource indicates that a given value set is related to specific condition. This is what eRSD and the reporting specification documents tell us. For example, CSTE may publish a value set containing LOINC codes for covid labs with OID x.y.z. The eRSD will tell us that value set x.y.z should be associated with the SNOMED condition code for covid. We want to know both that CSTE published value set x.y.z and that the eRSD indicates the concepts it contains are pertinent to covid.

 To address this could we a column to `condition_valueset_junction` called something like `source` that indicates the source we references to establish the link between a value set and a condition? We do need to consider that there is likely overlap between the eRSD and reporting spec documents so we have yet another many to many relationship. What about something like this?
 
 
image

robertandremitchell: thanks for the updated ERD! I've implemented, but quick question re: `condition_to_valueset_source`. Do we want ID to be a sequential_row_number? Also did you have another notion in mind for what `map_id` should be?

Screenshot 2024-08-01 at 11 42 06 AM

ericbuckley: Re `terminology_sources` and `valueset_types`.   Both tables have a single value and are only referenced once in the schema.  Is the overhead/complexity of having a lookup table necessary here?  In my experience, lookup tables are helpful if you have storage concerns and/or believe the values might be referenced in 2+ tables in the future.  If the answer is no for both of those questions, we might want to opt for simplicity and just remove them.

bamader: Yeah I kind of skew towards a smaller DB with less need to reference tables. Perhaps counter intuitively, I find that lookup tables can make it harder for me to parse what's happening because it represents another level of removal from what's going on.

DanPaseltiner: How about we take a pass on all of this for now just know that ultimately the source of all value set to condition mapping is from APHL. I was jumping ahead to a future state where we have additional terminology sources from orgs beyond APHL. @robertandremitchell let's go with your initial suggestion the author of the value sets is sufficient even if it is pretty much CSTE for everything.

DanPaseltiner: Basically we are just talking about the difference between a look up table or an enum styled value. Either works for me.

robertandremitchell:
per convo on next steps, we may want to add code system to value set level

robertandremitchell:
honestly forgot that this hasn't gotten merged, but can @m-goggins @bamader or @DanPaseltiner review when they have a chance todya or tomorrow?

bamader (approved):
Sorry for not getting back to this Rob, thanks for the reminder! Everything looks good to me, thanks for addressing the conversations and changes!

PR #2290, opened 2024-07-31
Modify eCR Library Table Formatting Part II

### LINKED ISSUE #2244: eCR Library Table Formatting Part II | Non-Integrated eCR Viewer
## User Story

As somebody viewing the eCR Library table, I want to see a user-friendly table, so that I can easily understand and scan through the eCRs my public health department has received

## Acceptance Criteria

The table must match the Figma designs linked below, specifically:
- [x] There should be a drop shadow below the table header (Y offset: -4px, Blur: 4px, Color: 000000 15% opacity) ~AND below the eCR Library + search panel~
- [x] Alternating grey and white rows, making sure colors in table match Figma designs
- [x] The column widths should look visually similar to the designs. There should be a minimum width on each column such that the column header stays on one line when making the table narrower. If the user makes the browser narrower than the min width, the table should not get any narrower (the user would have to just scroll right and left on the page to see the full page and table) 
- [x] The table should be responsive to resize when the browser is resized
- [x] Set table height so that scrolling is inside the table, not the whole page

## Scenarios

_Use the scenarios below to help in testing the ticket before/during/after design review to ensure it meets the criteria above_

Scenario 1:
- I resize my browser window to be wider and taller
- The table smoothly resizes to match the browser window, the edge of the table following the browser edge (not leaving any padding anywhere)

Scenario 2:
- I resize my browser window to be narrower
- The columns all decrease in width, until they reach their minimum width
- All column headers stay on one line
- When I make it super narrow, the browser forces me to scroll right and left to see the full table and page

Scenario 3:
- I try to scroll while hovering outside of the table
- I should not be able to scroll
- I try scrolling within the table area, below the table header
- I am able to scroll through all of the eCRs, and I see a scroll bar to indicate how far I am in the results
- I also see that every row alternates in background color between white and light gray

## Design Notes

_Owned by design team - Attach link to design in dev mode and/or prototype, as well as any notes on visual details, interaction expectations, accessibility considerations, copy requirements_

[Link to Figma design in Dev mode](https://www.figma.com/design/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?node-id=4955-47612&m=dev&t=zYuGs1ZKcquAFSSW-1)
[Link to Figma prototype](https://www.figma.com/proto/tXWUeV3Kjehd7VNqHojZYq/eCR-Viewer?page-id=4955%3A47258&node-id=4955-47612&viewport=-4385%2C2455%2C0.95&t=eC6BEpGPg9yXeSuW-1&scaling=scale-down&content-scaling=fixed)

angelathe:
# PULL REQUEST
Co-authored w/ @BobanL and @joshnygaard 

## Summary
**Modifies eCR Library table formatting:**
- Adds drop shadow below table header & below table (above pagination) -- (Y offset: -4px, Blur: 4px, Color: 000000 15% opacity)
- Make the size of the pagination banner at the bottom match the Figma designs
- Alternating grey and white rows, starting with white
- Table spans the whole browser width and height (not including the top header and bottom pagination banner).
- Table is responsive to resizing when the browser is resized
- Set table height with scrolling on the inside of the table (not the whole page)
    - Sticky DIBBs header & table header
- Add min widths for each column such that the column header stays on one line when making the table narrower. When the user makes the browser narrower than the min width, the table should not get any narrower (they would scroll right and left on the page to see the full page/table).

## Related Issue
Fixes #2244

## Checklist
- [ ] See original ticket for testing scenarios
- [ ] Please also do a visual check of the eCR viewer to make sure nothing has changed on that end


angelathe:
@sarahtress Let me know if you want me to attach screenshots or a loom for design review!

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -146,16 +146,81 @@ h4 {
...
 
-  tbody td:first-child {
-    width: 644px;
+  th:nth-child(4), td:nth-child(4) {
+    min-width: 11.75rem;
+  }
+
+  th:nth-child(5), td:nth-child(5) {
+    min-width: 11.5rem;
   }
```
BobanL, line 224: Similar to above, I would suggest doing this in the JS is easier for me to understand. 
```css
.minw-20 {
  min-width: 10rem;
}

.minw-1605 {
  min-width: 8.25rem;
}

.minw-1705 {
  min-width: 8.75rem;
}

.minw-2305 {
  min-width: 11.75rem;
}

.minw-23 {
  min-width: 11.5rem;
}
```
```jsx
  const header = [
    { value: "Patient", className: "minw-20" },
    { value: "Received Date", className: "minw-1605" },
    { value: "Encounter Date", className: "minw-1705" },
    { value: "Reportable Condition", className: "minw-2305" },
    { value: "RCKMS Rule Summary", className: "minw-23" },
  ];
  \\ other stuff
    {header.map((column) => (
    <th
      key={`${column.value}`}
      scope="col"
      className={column.className}
    >
      {column.value}
    </th>
  ))}
```

sarahtress:
> @sarahtress Let me know if you want me to attach screenshots or a loom for design review!

screen recording or pairing live works for me - whatever is easiest for you!

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -230,28 +295,31 @@ td {
...
+  align-self: stretch;
+  position: fixed;
+  bottom: 0px;
+  background-color: white;
 
-  .usa-pagination {
-    margin-top: 0 !important;
-  }
+  box-shadow: 0 -0.25rem 0.25rem rgba(0, 0, 0, 0.1);
+  z-index: 999;
```
BobanL, line 314: Maybe we don't need the z-index?

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -230,28 +295,31 @@ td {
...
-  flex-wrap: wrap;
-  flex-direction: column;
-
-  .usa-table {
-    margin-bottom: 0 !important;
-  }
+  align-self: stretch;
+  position: fixed;
+  bottom: 0px;
+  background-color: white;
```
BobanL, line 311: I don't know if we need this either?

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -230,28 +295,31 @@ td {
...
 }
 
-.homepage-wrapper {
+.pagination-bar {
+  .usa-pagination {
+    margin: 0;
+  }
+
+  width: 100%;
   display: flex;
```
BobanL, line 304: Tested removing `display: flex`, `align-items: center`, `justify-content: center` and wasn't able to notice any difference.

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -230,28 +295,31 @@ td {
...
   align-items: center;
-  flex-wrap: wrap;
-  flex-direction: column;
-
-  .usa-table {
-    margin-bottom: 0 !important;
-  }
+  align-self: stretch;
+  position: fixed;
+  bottom: 0px;
```
BobanL, line 310: These two might be able to be removed?

BobanL (changes requested):
I tested the app and it looks great! Accessibility passed and noticed no problems with zooming or anything like that. When I went through turning off some css on and offI noticed some CSS that might not need to be there? You might have a different experience or know more why those are needed, but left some suggestions there. 

Then just generally speaking, I know we have a conversation later today at 4 est where we are going to talk about css with the whole team. I left some comments with some bigger changes, that we should hold off until that meeting.

Generally, I think I just prefer using the [built in utility classes](https://designsystem.digital.gov/utilities/) wherever possible.

Pros: 
* Easier to read/understand what css is being applied in JSX files
* Snapshots become more valuable
* Knowing for sure making a change to one class won't break something else that is also using it
* Actual reuse of classes

Cons:
* Super wordy
  * This `className="ecr-library-viewer-container"` becomes `className="display-flex flex-column maxh-viewport overflow-hidden"`
* USWDS kind of lacks a bunch of steps that [something like tailwind](https://nerdcave.com/tailwind-cheat-sheet)** has. If we look at [min width](https://designsystem.digital.gov/utilities/height-and-width/#minw), it steps by 1/2 a rem at each step until 5 rem and then has one more at 7.5 rem. If we need to use another value we would have to create it.

Another possible approach would be to create multiple `.scss` files that would relate to each specific `.tsx` file and we could import it specifically in that file. (That would def be out of scope for this ticket though)

** Tailwind uses a ratio of 1/4 rem and USWDS uses 1/2 rem. So Margin top of 8 px would be `m-2`,`margin-1` respectively. I know in our code we use things like `margin-1` in a bunch of places so we would likely have to replace all of those with the tailwind equivalent.

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -146,16 +146,81 @@ h4 {
...
+  display: flex;
+  flex-wrap: wrap;
+  flex-direction: column;
+  overflow: auto;
+}
+
+.table-ecr-library {
+  margin: 0 0;
+  font-size: 1rem;
+  border-collapse: separate;
```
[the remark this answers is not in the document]
angelathe: This was necessary for keeping the table header bottom border sticky, otherwise it disappers when you scroll

containers/ecr-viewer/src/styles/custom-styles.scss
angelathe: @BobanL Was there a numbering system to the classnames above or was that random? I couldn't match these with the utilities in USWDS

angelathe:
@BobanL Ah sorry!! I tested these all individually at one point and they were necessary at the time, but recent changes made these unecessary. Fixing now and thanks for finding!

containers/ecr-viewer/src/styles/custom-styles.scss
angelathe: FYI, I ended up naming each class by pixel size that's equiv. to the min-width rem

BobanL: Sorry for the late reply, but the values provided were not random random, this just continued to follow the USWDS standards of 1 = 1/2 rem; I think we should probably stick with that naming scheme than moving to pixels.

angelathe: Got it! Changed it back

angelathe:
@BobanL I applied as many USWDS utility classes as I could, but kept running into some problems: 
1) I couldn't find utility classes for a few things (i.e., [font size w/o specifying family](https://designsystem.digital.gov/utilities/font-size-and-family/)? [border collapse=separate](https://designsystem.digital.gov/utilities/border/#utilities-section)?), 
2) Using the utility classes would be ineffective bc they would get overwritten by `usa-<>` classes, and 
3) I would've had to apply the classes to each item (i.e., adding classes to each `<td>` in a `<tr>`

I pushed my latest code, lmk if you have thoughts

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -146,16 +146,62 @@ h4 {
...
 
-  tbody td:first-child {
-    width: 644px;
+  .minw-2305 {
+    min-width: 11.75rem;
+  }
+
+  .minw-23 {
+    min-width: 11.5rem;
   }
```
BobanL, line 205: This looks like it's in the `table-ecr-library` class. Can we move it to the top level?

BobanL (approved):
One small nit, but otherwise looks great. Thank you for your hard work on this!

PR #2298, opened 2024-08-01
Add design review script for eCR Viewer

### LINKED ISSUE #2289: Yoink VIPER's cool Design Review script
## What needs to be done
Take VIPER's nifty design review script, copy it, and make it work for the eCR viewer!

## Why it needs to be done

We chatted with VIPER about their design review process and learned that they have a fancy automatic script for helping their designer get branches of the app running and viewable. Copying and updating the script to work for our needs will help Sarah more efficiently do design reviews without having to pair with someone or just look at screenshots.

## Acceptance Criteria

- [ ] Create a copy of VIPER's readme and design-review.sh found [here](https://github.com/CDCgov/phdi/tree/main/containers/tefca-viewer/design-review) and store it in the eCR Viewer portion of the repo in a directory
- [ ] Update the script to apply to the eCR Viewer instead of the TEFCA viewer
- [ ] Update the readme to apply to eCR Viewer not TEFCA, especially the initial setup portion
- [ ] The script should accept a branch name as an input, spin up the viewer from that branch, spin up the local dev DB, and automatically open a browser window to the eCR Viewer home page

## Technical Notes

All of the functionality is already written, it just needs to be tweaked to fit us. Should be pretty easy, no need for tests etc

angelathe:
# PULL REQUEST

## Summary
- Change VIPER's design review script for the eCR Viewer
- Add an argument to indicate whether or not to view the non-integrated vs. integrated eCR Viewer

## Related Issue
Fixes #2289

## Additional Information
Ty VIPER!

## Checklist

- [ ] Perform a mock design review using the script, and make sure it works :)

**To test this PR locally (Usage):**
1. Make sure you're on this branch: `angela/2289-design-review-script`
2. `cd containers/ecr-viewer/design-review`
3. `./design-review.sh <branch-to-test> <is_non_integrated>` where is_non_integrated is `true` or `false`

**To Test `Initial Setup` from the README:** I think this can only be tested once the PR merges to main? Because there's no design review script in the `ecr-viewer` container to download yet.

### DIFF
--- a/containers/ecr-viewer/design-review/README.MD
+++ b/containers/ecr-viewer/design-review/README.MD
@@ -1,0 +1,22 @@
+# Design Review Script
+
+A bash script intended to allow designers and members of the team to easily spin up an instance of the eCR Viewer on their local machines for UI/UX review. Currently only MacOS is supported.
+
+## Initial Setup
+
+In order to use this script a copy must be downloaded onto your computer from GitHub. To do this follow these instructions.
+1. Open the Terminal application.
+2. Copy and paste `curl https://raw.githubusercontent.com/CDCgov/phdi/main/containers/ecr-viewer/design-review/design-review.sh -O` into the Terminal prompt and press enter. This command uses the wget program to download a copy the `design-review.sh` file from this directory to the root level of your user directory e.g. `Users/johndoe`.
+3. Copy and paste `chmod +x design-review.sh` into Terminal and press enter. This command assigns executable permissions to `design-review.sh` allowing it to be run.
+
+
+## Usage
+
+Follow these steps to run script and spin up a local instance of the eCR Viewer.
+
+1. Ensure you have completed the initial setup instructions.
+2. Open the Terminal application.
+3. Copy and paste `./design-review.sh <MY-BRANCH>` into the Terminal prompt.
+4. Replace `<MY-BRANCH>` with the name of the GitHub branch you would like to conduct a review on. For example, `./design-review.sh main` will spin up an instance of the eCR Viewer based on the current state of the `main` branch of repository.
+5. Press enter. The script will now ensure that all required dependencies are installed on your machine, build and run the eCR Viewer, and finally navigate to the landing page in your system's default browser. Please note that because certain dependencies may need to be installed the script make take a few minutes the first time it is run on a machine. Additionally, depending on what needs to be installed you may be prompted at points during installation of dependencies to provide a password or click through some installation screens.
+6. When you are done with your review to shut the eCR Viewer down return to Terminal and press enter.
--- a/containers/ecr-viewer/design-review/README.MD
+++ b/containers/ecr-viewer/design-review/README.MD
@@ -6,8 +6,12 @@ A bash script intended to allow designers and members of the team to easily spin
 
 In order to use this script a copy must be downloaded onto your computer from GitHub. To do this follow these instructions.
 1. Open the Terminal application.
-2. Copy and paste `curl https://raw.githubusercontent.com/CDCgov/phdi/main/containers/ecr-viewer/design-review/design-review.sh -O` into the Terminal prompt and press enter. This command uses the wget program to download a copy the `design-review.sh` file from this directory to the root level of your user directory e.g. `Users/johndoe`.
-3. Copy and paste `chmod +x design-review.sh` into Terminal and press enter. This command assigns executable permissions to `design-review.sh` allowing it to be run.
+2. Copy and paste the following command into the Terminal prompt and press enter.
+```
+curl https://raw.githubusercontent.com/CDCgov/phdi/main/containers/ecr-viewer/design-review/design-review.sh -O && chmod +x design-review.sh
+```
+- The first command uses the wget program to download a copy the `design-review.sh` file from this directory to the root level of your user directory e.g. `Users/johndoe`.
+- The second command (`chmod +x design-review.sh`) assigns executable permissions to `design-review.sh` allowing it to be run.
 
 
 ## Usage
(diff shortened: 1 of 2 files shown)


angelathe:
**Question for reviewers:** When I test this locally where I'm looking at the non-integrated viewer, the loading page that comes up displays `Something went wrong!` -- and it takes one reload to get the eCR Viewer to display properly. Is anyone else seeing this (/ know why this may be happening)?

containers/ecr-viewer/design-review/design-review.sh
```diff
@@ -0,0 +1,101 @@
...
+# Checkout the specified branch
+git checkout $BRANCH_NAME
+
+# Write necessary env vars to .env.local
+echo "APP_ENV=test" > .env.local
+echo "DATABASE_URL=postgres://postgres:pw@db:5432/ecr_viewer_db" >> .env.local
+echo "NEXT_PUBLIC_NON_INTEGRATED_VIEWER=$IS_NON_INTEGRATED" >> .env.local
+
+# Build and run docker-compose with APP_ENV=TEST
+docker-compose build --no-cache && docker compose --env-file .env.local up -d
```
BobanL, line 87: Could we just simplify this line by doing `docker compose up ecr-viewer db --build` so we don't end up including any of the logging stuff?

containers/ecr-viewer/design-review/design-review.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "Waiting for $URL to be available..."
+    sleep 5
+done
+
+# Open in default browser
+open http://localhost:3000/
+
+# Prompt to end review session
+read -p "Press enter to end review"
+docker compose down
```
BobanL, line 101: I think we should add [`-v`](https://docs.docker.com/reference/cli/docker/compose/down/) here so that the db volume gets deleted too. `docker compose down -v`

containers/ecr-viewer/design-review/README.MD
```diff
@@ -0,0 +1,24 @@
+# Design Review Script
+
+A bash script intended to allow designers and members of the team to easily spin up an instance of the eCR Viewer on their local machines for UI/UX review. Currently only MacOS is supported.
+
+## Initial Setup
+
+In order to use this script a copy must be downloaded onto your computer from GitHub. To do this follow these instructions.
+1. Open the Terminal application.
+2. Copy and paste `curl https://raw.githubusercontent.com/CDCgov/phdi/main/containers/ecr-viewer/design-review/design-review.sh -O` into the Terminal prompt and press enter. This command uses the wget program to download a copy the `design-review.sh` file from this directory to the root level of your user directory e.g. `Users/johndoe`.
+3. Copy and paste `chmod +x design-review.sh` into Terminal and press enter. This command assigns executable permissions to `design-review.sh` allowing it to be run.
```
BobanL, line 10: I'm curious if we did something like [this](https://stackoverflow.com/a/40979016), would the file always have execute permissions even if downloaded?

containers/ecr-viewer/design-review/design-review.sh
BobanL: Also I don't think we should be using [docker-compose since it's deprecated](https://docs.docker.com/compose/migrate/).

containers/ecr-viewer/design-review/design-review.sh
```diff
@@ -0,0 +1,101 @@
...
+open /Applications/Docker.app
+echo "Waiting for Docker to launch..."
+while ! docker system info > /dev/null 2>&1; do
+    sleep 1
+done
+
+# Install Docker Compose if it's not already installed
+if ! command_exists docker-compose; then
+    brew install docker-compose
+fi
```
BobanL, line 57: Also I don't think we should be using [docker-compose since it's deprecated](https://docs.docker.com/compose/migrate/).  This functionality is in docker as `docker compose` (no dash)

containers/ecr-viewer/design-review/design-review.sh
```diff
@@ -0,0 +1,101 @@
...
+
+if ! command_exists node; then
+    echo "Node.js not found, installing it now..."
+    brew install node
+fi
```
BobanL, line 62: Do we need to install node its just running docker? I have no idea how to test this since I'm not going to uninstall node

BobanL (commented):
Tested locally and didn't see the `Something went wrong!` page, just left a couple of nits/questions

containers/ecr-viewer/design-review/design-review.sh
angelathe: Hm, for some reason, the command above didn't work for me (docker compose gets stuck on attaching to the containers).

Running in detached mode does work though: `docker compose --env-file .env.local up -d ecr-viewer db --build` -- does that sound alright?

BobanL: Yeah that works!

containers/ecr-viewer/design-review/README.MD
angelathe: Hmm, do you know how could we test this?

BobanL: Commit it with the +x and I'll download it locally and see if it says I need execute permissions still.

angelathe: Just pushed

BobanL: Actually it seems like it didn't work and [it won't work like that 😢](https://askubuntu.com/questions/742448/download-executable-script-from-github-preserving-x-permissions). I do like the suggestion in that thread of just appending `&& chmod +x ./design-review.sh` to that command so it can be like a one stop command to run rather than two different ones.

BobanL (approved):
Looks great 👍

containers/ecr-viewer/design-review/README.MD
angelathe: Ooh I like this change. Just updated the initial setup cmd + readme

PR #2301, opened 2024-08-01
Query struct

### LINKED ISSUE #2299: Refactor query-service to use JSON data structures

bamader:
# PULL REQUEST

## Summary
This PR creates a CustomQuery data class that holds the various condition codes and generated query strings used for queries to the FHIR server. It also refactors all of our demo use cases to take advantage of this unified representation, and condenses all the different query service function calls down into a single generalized function.

## Related Issue
Fixes #2299 

## Additional Information
Once we move to creating a database, importing the query structs will no longer be important, and we'll use an API to retrieve them as JSON text from the DB proper. For now, though, this is the easiest way to keep things moving.


bamader:
Oh, and just to note it here, I confirmed that all of our demo cases look the exact same using the new condensed code and query generation as they currently do in main, with each one having its own separate function. All the same sections are still displayed in the viewer in both cases.

DanPaseltiner (approved):
This looks great to me. I think there will be work do on refining the query template structure to support multiple valuesets with codes from different systems and such, but this is an excellent start.

containers/tefca-viewer/src/app/CustomQuery.ts
```diff
@@ -0,0 +1,135 @@
...
+  getQuery(desiredQuery: string): string {
+    switch (desiredQuery) {
+      case "observation":
+        return this.observationQuery;
+      case "diagnostic":
+        return this.diagnosticReportQuery;
+      case "condition":
+        return this.conditionQuery;
+      case "medication":
+        return this.medicationRequestQuery;
```
m-goggins, line 124: nit: can this case be `medicationRequest`? There are several "medications" (request, administration, etc.)

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -171,299 +196,42 @@ export async function UseCaseQuery(
...
-
-  // Query for encounters. TODO: Add encounters as _include in condition query & batch encounter queries
-  if (queryResponse.Condition && queryResponse.Condition.length > 0) {
-    const conditionId = queryResponse.Condition[0].id;
-    const encounterQuery = `/Encounter?subject=${patientId}&reason-reference=${conditionId}`;
-    const encounterResponse = await fhirClient.get(encounterQuery);
-
-    queryResponse = await parseFhirSearch(encounterResponse, queryResponse);
-  }
+  await generalizedQuery(
```
m-goggins, line 199: Nice 😎

m-goggins (approved):
Looks much cleaner! 

Unrelated to the new code in this PR, but I think the phone number functions should be moved out of `query-service` to either their own service or maybe in formatting-service?

containers/tefca-viewer/src/app/CustomQuery.ts
bamader: For sure, no problem at all!

containers/tefca-viewer/src/app/query-service.ts
bamader: I get an embarrassing hit of dopamine from either dropping or receiving sunglasses on a PR 🤣

bamader:
@m-goggins Yeah I think you're right about moving the phone standardization utils out of query-service. Initially I had them on the client side but Dan made the point they should be accessible on the server side too so the API service could also use them, but format-service looks like an excellent place for both to still reach them. Probably a small separate ticket (which I'll write up) but a good call out.

PR #2315, opened 2024-08-05
Fix modal formatting

### LINKED ISSUE #2288: Correct demo modal styling
The current layout and styling of the modal that allows users to select their demo experience does not exactly match our [designs](https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=2923-12508&m=dev). Let's bring out implementation of this component into alignment with the designs.

DanPaseltiner:
# PULL REQUEST

## Summary
This PR updates styling and layout for the modal. Additionally, we manually apply focus styling in order to support Safari. In the future we should probably expand this solution beyond just the modal. The next button on the modal is now always disabled until the user makes a selection. Finally, there were some issues with phone number auto population when a user would change their selection from a demo patient with a phone number to one without or vice versa.

## Related Issue
Fixes #2288 

## Additional Information

## Checklist


containers/tefca-viewer/src/app/page.tsx
```diff
@@ -37,6 +37,16 @@ export default function LandingPage() {
...
+    document.getElementById(option)?.focus();
+  };
+
+  const modalOptions = {
+    chlamydia: "Chlamydia case investigation",
+    gonorrhea: "Gonorrhea case investigation",
+    syphilis: "Syphilis case investigation",
+    cancer: "Cancer case investigation",
+    "newborn-screening": "Newborn screening follow-up",
+    "social-determinants": "Gather social determinants of health for a patient",
```
m-goggins, line 49: We have a very similar const in `constants.ts` called `demoQueryOptions` that I think we should try to re-use instead of redefining here. Or if that doesn't work, I would move modalOptions to the `constants.ts` file.

containers/tefca-viewer/src/app/page.tsx
```diff
@@ -133,70 +144,35 @@ export default function LandingPage() {
...
+              {Object.keys(modalOptions).map((option) => (
+                <Button
+                  type="button"
+                  className={`modal-option ${
+                    selectedOption === option ? "selected" : ""
+                  }`}
+                  id={option}
+                  onClick={() => handleOptionClick(option)}
+                >
+                  {modalOptions[option as keyof typeof modalOptions]}
```
m-goggins, line 165: nice!

containers/tefca-viewer/src/app/constants.ts
```diff
@@ -65,6 +65,7 @@ export const demoData: Record<PatientType, DemoDataFields> = {
     LastName: "JMC",
     DOB: "2001-05-07",
     MRN: "b50z-wayszq-ofib",
+    Phone: "",
```
m-goggins, line 68: We could add this as another ticket, but I think a more robust solution would be to change `fillFields` to reset fields before setting with new data (or something like that).

m-goggins (commented):
Looks pretty good to me! The selection looks sweet. Just had a couple of comments about filling the fields and moving constants to another spot or reusing constants that already exist.

containers/tefca-viewer/src/app/page.tsx
DanPaseltiner: Yeah, I looked at trying to reuse `demoQueryOptions`, but the problem is that for the UI designs we need this specific copy. I'm open to suggestions, but can definitely move into `constants.ts`.

lg-king (approved):

robertandremitchell (approved):
looks good. minor quibble I don't know if it matters: if you click on one of the modal boxes, it does highlight in blue. however, if you then click elsewhere, it does seem to remember your selection, but loses the blue highlight. not sure if it should retain the blue highlight; it does appear to retain the lighter blue font.

containers/tefca-viewer/src/app/constants.ts
bamader: Yeah very firm agree here. This is also making e2e tests somewhat wonky depending on the use case we're testing. I would be okay with making it a separate ticket.

containers/tefca-viewer/src/app/page.tsx
bamader: If we have to have this specific copy I would move to `constants.ts`

bamader (commented):
Overall looks solid, just some thoughts around the constants file and the field clearing

containers/tefca-viewer/src/app/page.tsx
m-goggins: Maybe this is an opportunity to standardize the copy across different places?

bamader: Would love that tbh. Too many variations of `chlamydia` `chlamydia-sti` `chlamydia-sti-positive` etc. 😅

DanPaseltiner: I have moved to constants. I think a large effort to improve copy consistency should wait for another ticket as it requires coordination with design

containers/tefca-viewer/src/app/constants.ts
DanPaseltiner: Let's address this in another ticket if needed. I believe a "reset" in this context is simply setting the value to an empty string which is what the current solution already does. I could very well be missing something here, but I don't much value in a dedicated reset step. I'm going to merge as is and we can discuss next week when we are all back in the office if a follow up ticket here is required.

PR #2327, opened 2024-08-06
Add hover to multiple results

### LINKED ISSUE #2283: Add hover to multiple patients found page
When we render the results a search that returns multiple patient records let's implement the interactive highlighting on hover called for in [these](https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=2923-11351&t=cGjEWcedWbgUXg40-4) designs.

bamader:
# Hover On Table Rows of Multiple Records Found

## Summary
This PR adds a hover styling to the "multiple patient results" table following a user's search. Color and opacity pulled from figma designs.

## Related Issue
Fixes #2283 

## Additional Information
Anything else the review team should know? Definitely not how long it took me to correctly locate the `td` within the styling subclass.

### DIFF
--- a/containers/tefca-viewer/src/styles/custom-styles.scss
+++ b/containers/tefca-viewer/src/styles/custom-styles.scss
@@ -266,8 +266,7 @@ body {
   .usa-table--borderless {
     tr {
       &:hover td {
-        background-color: #d9e8f6;
-        opacity: 50%;
+        background: rgba(217, 232, 246, 0.5);
       }
     }
   }
--- a/containers/tefca-viewer/src/styles/custom-styles.scss
+++ b/containers/tefca-viewer/src/styles/custom-styles.scss
@@ -1,525 +1,527 @@
 @use "uswds-core" as *;
 
 .section__line {
-  border-top: 1px solid #99deea;
-  margin-top: 0.5rem;
-  margin-bottom: 0.5rem;
+    border-top: 1px solid #99deea;
+    margin-top: 0.5rem;
+    margin-bottom: 0.5rem;
 }
 
 .section__line_gray {
-  border-top: 1px solid #dfe1e2;
-  margin-top: 0.5rem;
-  margin-bottom: 0.5rem;
+    border-top: 1px solid #dfe1e2;
+    margin-top: 0.5rem;
+    margin-bottom: 0.5rem;
 }
 
 .card__line {
-  border-top: 8px solid #edeff0;
+    border-top: 8px solid #edeff0;
 }
 
 h1 {
-  margin-bottom: 0.25rem;
+    margin-bottom: 0.25rem;
 }
 
 h2 {
-  font-size: 40px;
+    font-size: 40px;
 }
 
 h3 {
-  font-size: 22px;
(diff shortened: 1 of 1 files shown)


containers/tefca-viewer/src/styles/custom-styles.scss
```diff
@@ -270,6 +262,17 @@ body {
   justify-content: center;
   margin: 0 auto;
   max-width: 1000px;
+
+  .usa-table--borderless {
+    tr {
+      &:hover td {
+        background-color: #D9E8F680;
```
m-goggins, line 269:
```suggestion
        background-color: #d9e8f6;
```

I think this is the color from the Figma file

containers/tefca-viewer/src/styles/custom-styles.scss
```diff
@@ -270,6 +262,17 @@ body {
...
+        opacity: 50%;
```
m-goggins, line 270: I know the Figma file says 50% opacity but something about the way it's rendering on my screen looks really different from the Figma file. The text looks washed out? I tried changing the opacity to 100% and it honestly looks better to me, but wanted to see if it looks different for anyone else.

m-goggins (commented):
I left some more specific comments in the code but something about the hover looks "off" to me.

containers/tefca-viewer/src/styles/custom-styles.scss
bamader: It totally is, I have literally no idea where the random 80 came from

containers/tefca-viewer/src/styles/custom-styles.scss
bamader: The text does look a little grayer to me with the (now correct) background color changed but I actually think it looks pretty slick. I wouldn't describe it as washed out, it looks more to me like a "contrast complement" color, if that makes sense. I don't mind bumping up to 100% opacity if we think it's a problem, but I'm happy at 50 too. Whatever works!

robertandremitchell: it looks gray to me, as well, but I'm ambivalent.

bamader:
@lg-king Tagging you in for a design question! We're not sure about the opacity and color setting here, so just want to check that this is the configuration we want. It's pulled from figma but it's rendering washed out for Marcelle, and Rob and I are fine with whatever color we pick.

lg-king:
Yeah I think the hover looks off with the grey text. Is there a way to apply the opacity just to the background but not the text? If not I think just changing the background color without changing the opacity would work - try hex code #ecf3fa (that should match the lighter opacity of D9E8F6).

bamader:
Yeah I think we can just apply opacity to the background. We'll need the color as an RGB value though because we won't be using the opacity parameter, we'll use a different setting instead.  I think` rgb(217, 232, 246)` should be the right value?

bamader:
@m-goggins Opacity should be fixed now, see if the text renders better for you with this new setting.

m-goggins (approved):
Looks great!

PR #2333, opened 2024-08-07
Add python linting workflow

### LINKED ISSUE #1985: Add linting workflows
**Context**:
In separating out the workflows to run per container, we ended up with a lot of duplicate code that we'd like to make more reusable. For example, each of the python containers has a `python-linting` step that is repeated code. We'd like to make separate workflows for `python-linting` and `javascript-linting` that can be called from within each container workflow.

**Acceptance Criteria**:

- [ ] A `python-linting` workflow has been added that includes all the steps in the `python-linting` step of a container workflow
- [ ] A `javascript-linting` workflow has been added that includes all the steps in the `javascript-linting` step of a container workflow
- [ ] The current linting steps are replaced with references to the newly created linting workflows

m-goggins:
# PULL REQUEST

## Summary
This PR introduces a python linting workflow that container workflows can refer to instead of having the same code copied across all the container workflows.

There are two caveats to the new workflow:
1. You cannot pass env vars through to workflows, such as the `env.TEST_RUNNER_PYTHON_VERSION`. There are workarounds that we could introduce, e.g., add the python version to secrets which can be passed to/inherited by workflows. However, I think this introduces a level of complexity that we do not yet need because a dev would have to know to update the python version in secrets, etc. 
2. I was unable to get the python linting in the orchestration workflow to run using the new workflow. There doesn't appear to be any differences between  the `orchestration-python-linting` job and what's in the `linting-python.yaml` workflow but whenever I switched the orchestration container workflow to use the linting workflow, the linting never ran; the test sat on "queued" for several hours. I'm curious if anyone has insight into this or strong opinions on calling this 'good enough for now' or if we should wait to figure this out before introducing the linting workflow (I won't be offended if you tell me to let this ticket/work sit until we have more dedicated tech debt time 😅)

## Related Issue
Fixes #1985 

## Additional Information

## Checklist

### DIFF
--- a/.github/workflows/container-dibbs.yaml
+++ b/.github/workflows/container-dibbs.yaml
@@ -26,29 +26,9 @@ env:
 
 jobs:
   python-linting:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Setup python ${{env.TEST_RUNNER_PYTHON_VERSION}}
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{env.TEST_RUNNER_PYTHON_VERSION}}
-          cache: pip
-
-      - name: Install dependencies
-        run: |
-          pip install -U pip
-          pip install ruff==0.4.3
-
-      - name: Run linter (ruff)
-        run: |
-          ruff check --output-format=github .
-
-      - name: Run formatter (ruff)
-        run: |
-          ruff format --check
+    uses: ./.github/workflows/linting-python.yaml
+    with:
+      python_runner_version: 3.11 # must be hardcoded; cannot pass in from env
 
   unit-test-python-containers:
     runs-on: ubuntu-latest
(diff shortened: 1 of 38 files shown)


bamader (approved):
I'm for sure of the "this is good enough, merge this bad boi in" opinion. It's a solid refactor and even if orchestration isn't yet benefiting from it, this still cuts down on a lot of kruft. I also wholeheartedly agree that we don't need the complexity of trying to pass in secrets and having devs update those every time.

robertandremitchell (approved):
very weird it would not work for orchestration. I don't see any reason why it wouldn't, but simplifying 8 of 9 containers feels like a good W for this ticket. I agree adding python version to secrets feels like a layer of abstraction/obfuscation too many.

DanPaseltiner (approved):
LGTM, but I think we should make sure folks outside of Viper have a chance put eyes on this before merging. @gordonfarrell and @nickbristow

nickbristow (approved):

PR #2338, opened 2024-08-07
Correctly find reportable conditions

### LINKED ISSUE #2331: TCR Service Not Correctly Finding Reportable Conditions
## Description

The goal #2123 was for the Trigger Code Reference Service to automatically find all reportable conditions within a FHIR bundle. Instead of looking at the "Reportability Response Information Section" of the FHIR bundle I instead looked up resources within the bundle that contained a SNOMED condition code. While the found conditions would indeed likely be a reportable condition there are additional things that can be a reportable.

## Steps to reproduce

This is tricky to reproduce an error specifically as there are several issues currently with stamping conditions. 

## Expected Result

The `find_conditions` in the TCRS `utils.py` module should use the `entry` list in the "Reportability Response Information Section" to find appropriate condition codes.

## Actual result

It currently looks for any resource with a SNOMED condition.

## Priority

#### Priority
- [x] **High priority** _High priority bugs must be addressed immediately. They often affect the app’s functionality and impact customers and their user experience. They must take priority._

## Notes

JNygaard-Skylight:
# PULL REQUEST

## Summary
Uses the reportable conditions section to find the SNOMED codes of those conditions

## Related Issue
Fixes #2331 

## Additional Information
I'm unsure how best to test this, as there appear to be several issues with reportable conditions, stamping relevant labs, etc. 

Check the eCR pictured, and similar conditions and lab should be listed in the summary.
Screenshot 2024-08-07 at 17 12 07
Screenshot 2024-08-07 at 17 12 17
Screenshot 2024-08-07 at 17 11 51

## Checklist


containers/trigger-code-reference/app/utils.py
```diff
@@ -201,31 +201,50 @@ def read_json_from_assets(filename: str) -> dict:
...
-    # Extract unique SNOMED codes from the observations
-    snomed_codes = {
-        coding["code"]
-        for obs in resources_with_conditions
-        for coding in obs.get("valueCodeableConcept", {}).get("coding", [])
-        if coding["system"] == SNOMED_URL
-    }
-
-    return snomed_codes
+    # Get list of compositions
```
bamader, line 214: I think the cleanest way we want to actually do this is by just leveraging the `fhirpathpy` module. That allows us to go directly to a particularly resource of interest and manipulate it without worrying about trying to filter by codes just within the reportability trigger. You can see examples of this in a number of our existing containers, including record linkage and ingestion, but the gist is basically the body of the util function `extract_value_with_resource_path`. 

```
path_to_triggering_entries = fhirpathpy.compile("Bundle.entry.resource.where(resourceType='Composition').section.where(title = 'Reportability Response Information Section').entry")
trigger_entries = path_to_triggering_entries(bundle)
```

will directly give you the `entry` array you want on the specific Composition resource in the input bundle. From there, you just need need to grab the reference IDs out of the list with e.g.

```
triggering_IDs = [x["reference"] for x in trigger_entries]
```

You can then use similar fhirpathpy compiling and path parsing to look up the referenced resources, with the `.where()` clause of where ID = whatever we pulled out of the composition. Using fhirpathpy like this will not only be more efficient since the FHIR resources gets directly compiled, but will also protect against situations in which SNOMED or LOINC codes don't necessarily appear in the reportability composition and so we'd miss them by not just grabbing based on title (the cardinality of which is 1, hence required, in the eICR FHIR spec).

bamader (changes requested):
It's a good start, but let's switch over to using fhirpathpy since it excels in this area and will be more robust in the long term.

bamader (approved):
🔥 Looks great! Thanks for making that change to fhirpathpy. That'll ensure really fast compilation on the server side when we go to work with these codes, and now we'll be sure we get everything that the eHR flagged as reportable.

PR #2385, opened 2024-08-14
add extended ecr data schema

### LINKED ISSUE #2349: Create new extended database schema for Philly
## What needs to be done

We need a new, alternate DB schema for use with Philly's on prem DB which will include significantly more information than our standard eCR viewer DB schema.

## Why it needs to be done

We're going to be using Philly's on prem DB for two purposes. First, it will hold the information needed to populate the eCR Library table, just like we use a local or cloud DB to. Second, Philly wants to store a large amount of additional eCR information in this DB so they can easily make SQL queries into the eCR data. This new schema will set up the DB in such a way that it can work for both functions.

## Acceptance Criteria

- [x] Create a new extended_db.sql file
- [x] Store the file in a new directory under ecer-viewer/src/app call sql
- [x] Write the sql to create the necessary tables, similar to how we're doing it in the file `01-init.sql`
- [x] There will be three tables, `ecr_data` `ecr_rr_conditions` and  `ecr_labs`. Details on columns and typing can be found [HERE](https://cdc.sharepoint.com/:x:/r/teams/OPHDST-IRD-DIBBs/_layouts/15/Doc.aspx?sourcedoc=%7BCB3CFBAA-5D2D-44C1-AA15-F6760CF97126%7D&file=Philadelphia%20Suggested%20Schema.xlsx&action=default&mobileredirect=true) in the tabs along the bottom

## Technical Notes

This isn't actually standing up a DB or using this for anything, just creating the schema for future use.

BobanL:
# PULL REQUEST

## Summary
- Create new `extended_metadata_db.sql` file 
- **Written for SQL Server**
  - UUID type is supported in Postgres and SQL Server, but not MySQL or Oracle. 
  - DATETIME is supported in SQL Server and MySQL but not in Oracle or postgres.
  - UUID would be pretty easy and seamless to change to varchar, but if we convert DATETIME to varchar we would probably lose out a lot.

## Related Issue
Fixes #2349 

## Additional Information
- Acceptance calls out to be stored in `src/app`, but I believe it makes more sense to store the schema outside of the `src` folder to keep it totally separate from the typescript folder structure we have now
- An additional table was created to handle `ecr_rr_conditions.rule_summary` since it is an array of data and SQL server does not support array types.


containers/ecr-viewer/sql/extended_metadata_db.sql
```diff
@@ -0,0 +1,83 @@
...
+    current_job_title        VARCHAR(255),
+    current_job_industry     VARCHAR(255),
+    usual_occupation         VARCHAR(255),
+    usual_industry           VARCHAR(255),
+    preferred_language       VARCHAR(255),
+    pregnancy_status         VARCHAR(255),
+    rr_id                    VARCHAR(255),
+    processing_status        VARCHAR(255),
+    eicr_version_number      VARCHAR(50),
+    authoring_datetime       DATETIME,
```
BobanL, line 33: `DATETIME` is supported in SQL Server and MySQL but not in Oracle or postgres.

containers/ecr-viewer/sql/extended_metadata_db.sql
```diff
@@ -0,0 +1,83 @@
...
+    encounter_type           VARCHAR(255),
+    encounter_start_date     DATETIME,
+    encounter_end_date       DATETIME,
+    reason_for_visit         VARCHAR(255),
+    active_problems          VARCHAR(255)
+);
+
+CREATE TABLE IF NOT EXISTS ecr_rr_conditions
+(
+    UUID      UUID PRIMARY KEY,
```
BobanL, line 47: UUID type is supported in Postgres and SQL Server, but not MySQL or Oracle.

gordonfarrell:
Hmmmm we may need to up the varchar counts for certain fields, for example the rule summary on the RR. Basically anything thats like a comment field where any amount of stuff might be written. I just did a character check on a rule summary from a random RR and it was ~1200 characters

BobanL:
> Hmmmm we may need to up the varchar counts for certain fields, for example the rule summary on the RR. Basically anything thats like a comment field where any amount of stuff might be written. I just did a character check on a rule summary from a random RR and it was ~1200 characters
> 
> Edit: Whoops just realized you already have that summary as TEXT. Are there any others which might be longer than 255? Hard to tell. Are comments being pulled into this in any way? I've seen a few random fields with a comment thats over 255 for sure, but I dont know if this will include those

Yeah, unsure which fields will be long but for now I set `reason_for_visit`, `active_problems`, `condition`, and `ecr_summary` to be just a [variable length VARCHAR](https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver16#:~:text=Use%20varchar%20when%20the%20sizes%20of%20the%20column%20data%20entries%20vary%20considerably.). The rest of the fields look OK with their limit I think, but I could just make them variable?

gordonfarrell:
> > Hmmmm we may need to up the varchar counts for certain fields, for example the rule summary on the RR. Basically anything thats like a comment field where any amount of stuff might be written. I just did a character check on a rule summary from a random RR and it was ~1200 characters
> > Edit: Whoops just realized you already have that summary as TEXT. Are there any others which might be longer than 255? Hard to tell. Are comments being pulled into this in any way? I've seen a few random fields with a comment thats over 255 for sure, but I dont know if this will include those
> 
> Yeah, unsure which fields will be long but for now I set `reason_for_visit`, `active_problems`, `condition`, and `ecr_summary` to be just a [variable length VARCHAR](https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver16#:~:text=Use%20varchar%20when%20the%20sizes%20of%20the%20column%20data%20entries%20vary%20considerably.). The rest of the fields look OK with their limit I think, but I could just make them variable?

How it is now is probably fine, we can always change it later if we feel some squeeze.

gordonfarrell (approved):
LGTM!

PR #2400, opened 2024-08-19
add environment types for DB type and schema

### LINKED ISSUE #2347: Create two new environment variables for the eCR Viewer
## What needs to be done

We need two new variables to allow the non-integrated viewer to be configured depending on what the database needs are, one for the type of database being used and one for which schema the DB will be implementing.

## Why it needs to be done

Very soon we'll have several different possible DB configurations:

1. The first is for local dev, basically what we have now, a small subset of fields being grabbed from the eCR to populate the eCR library.

2. The second is almost exactly the same except it will be interacting with an AWS SQL server for AWS setups
 
3. Same again but Azure for New Mexico.
 
4. Finally, we'll have a configuration for Philly that uses their own on-prem MS SQL Server and has an expanded set of eCR fields that capture most of an eCRs important data

Creating these env vars will let us swap around DB setups to make any of these scenarios possible.

## Acceptance Criteria

- [ ] A new env var called DATABASE_TYPE is implemented in the eCR viewer
- [ ] A new env var called DATABASE_SCHEMA is implemented in the eCR Viewer

## Technical Notes

These don't have to do anything right now, just exist to be used later in this epic.

BobanL:
# PULL REQUEST

## Summary
- Add environment.d.ts file to host all environment variable types

## Related Issue
Fixes #2347 

## Additional Info
- Another PR will be made in the [dibbs-aws repo](https://github.com/CDCgov/dibbs-aws) to add the two new environment variables.

### DIFF
--- a/containers/ecr-viewer/environment.d.ts
+++ b/containers/ecr-viewer/environment.d.ts
@@ -1,0 +1,18 @@
+/* eslint-disable unused-imports/no-unused-vars */
+namespace NodeJS {
+  interface ProcessEnv {
+    APP_ENV: "test" | "middleware" | "prod";
+    AWS_REGION: string;
+    DATABASE_SCHEMA: "core" | "extended";
+    DATABASE_TYPE: string;
+    DATABASE_URL: string;
+    ECR_BUCKET_NAME: string;
+    GITHUB_ID: string;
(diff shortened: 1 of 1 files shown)


gordonfarrell (approved):
LGTM!

PR #2425, opened 2024-08-22
Read/write bundles to/from Azure Blob Storage

### LINKED ISSUE #2398: Implement writing to and reading from Azure Blob Storage
## What needs to be done

Add the ability for the app to read and write fhir bundles to and from Azure blob storage.

## Why it needs to be done

We have a working setup for storage using AWS S3 buckets, but that will only work for some of our pilot partners. Since New Mexico is one of the STLTs closest to implementing the viewer, and they're also an Azure shop, we need to add flexibility to the viewer so it knows which cloud service it's being hosted in and can pick the proper storage option.

## Acceptance Criteria

- [x] Adjust existing env vars or create a new one to allow configuring the app to save to Azure Blob storage rather than AWS
- [x] Implement an Azure Blob Storage read and write (See linked docs below)
- [x] Use conditional logic with the env var to save to blob storage when the app is configured for Azure
- [x] Tests
- [ ] S3 read and write should work as normal if the app is configured for AWS

## Technical Notes

Azure Blob Storage Typescript docs [here](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-typescript-get-started?tabs=azure-ad#create-a-blobserviceclient-object)

angelathe:
# PULL REQUEST

## Summary
- Add functionality to read/write FHIR bundles to/from Azure blob storage
- Add env vars
- Add mock API tests

Note: Currently, we don't have access to Azure, so I used placeholder env vars and had to move a couple lines around so things wouldn't error out.

## Related Issue
Fixes #2398

## Checklist
Don't know what testing can currently be done without DIBBs access to Azure?

## Acceptance Criteria
- [ ] Adjust existing env vars or create a new one to allow configuring the app to save to Azure Blob storage rather than AWS
- [ ] Implement an Azure Blob Storage read and write (See linked docs below)
- [ ] Use conditional logic with the env var to save to blob storage when the app is configured for Azure
 Tests
- [ ] S3 read and write should work as normal if the app is configured for AWS


containers/ecr-viewer/src/app/api/fhir-data/route.ts
```diff
@@ -1,8 +1,9 @@
 import { NextRequest, NextResponse } from "next/server";
-import { get_s3, get_postgres } from "./fhir-data-service";
+import { get_s3, get_azure, get_postgres } from "./fhir-data-service";
 
 const S3_SOURCE = "s3";
 const POSTGRES_SOURCE = "postgres";
+const AZURE_SOURCE = "azure";
```
austin-hall-skylight, line 6: It's not a showstopper for me but it would be cleaner to use an enum here.

containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -1,8 +1,13 @@
...
-import { saveToS3, saveToPostgres } from "./save-fhir-data-service";
+import {
+  saveToS3,
+  saveToAzure,
+  saveToPostgres,
+} from "./save-fhir-data-service";
 
 const S3_SOURCE = "s3";
 const POSTGRES_SOURCE = "postgres";
+const AZURE_SOURCE = "azure";
```
austin-hall-skylight, line 10: Enum here too.

austin-hall-skylight (approved):
Tested with Azure storage emulator and everything is working

containers/ecr-viewer/src/app/api/fhir-data/route.ts
BobanL: Even better if they were just in a single place and imported to both of these files.

angelathe: @BobanL @austin-hall-skylight Updated so that the source consts are in `api/utils` & imported to both route files

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -85,3 +86,53 @@ export const saveToS3 = async (fhirBundle: Bundle, ecrId: string) => {
...
+ * @function saveToAzure
+ * @param fhirBundle - The FHIR bundle to be saved.
+ * @param ecrId - The unique ID for the eCR associated with the FHIR bundle.
+ * @returns A promise that resolves when the FHIR bundle is successfully saved to Azure Blob Storage.
+ * @throws {Error} Throws an error if the FHIR bundle cannot be saved to Azure Blob Storage.
+ */
+export const saveToAzure = async (fhirBundle: Bundle, ecrId: string) => {
+  // TODO: Make this global after we get Azure access
+  const blobClient = BlobServiceClient.fromConnectionString(
+    process.env.AZURE_STORAGE_CONNECTION_STRING!,
```
gordonfarrell, line 102: Is the ! on the end of this supposed to be there?

gordonfarrell (approved):
Looks really good, great work on this Angela! Just one tiny comment/question, gonna approve though so you don't have to hunt me down for a re-approval if it needs a change. Thanks!

PR #2427, opened 2024-08-22
Add ICD-9 codes to DB

### LINKED ISSUE #2426: Add ICD9 codes to SQLite DB

bamader:
# Represent ICD9 In Resource Stamping

## Summary
This PR:
* Adds the GEM files downloaded directly from CMS into the repo as accessible data sources for our ETL process (I opted to include documentation about them as well in case either a. anyone was ever curious what things meant, and b. in case we ever wanted to do some more sophisticated processing around the boolean match flags, which we currently ignore)
* Augments the SQLite DB creation to now include a crosswalk table that references our ICD10 information
* Updates the TCR endpoint and utils functions to now also search for ICD9 equivalencies when querying for an ICD10 code (the GEM files are formatless so we needed a little bit of extra checking to make this work)

## Related Issue
Fixes #2426 

## Additional Information
THIS PR IS NOT AS LONG AS IT SEEMS DO NOT BE SCARED. Most of those changes are the GEM files, documentation, and the updated DB.


robertandremitchell (approved):
looks good. one minor quibble about naming conventions, but I think the implementation makes sense, and definitely good to save the docs.

PR #2432, opened 2024-08-23
Fix TEFCA CustomizeQuery Checkboxes

### LINKED ISSUE #2411: Fix CustomizedQuery checkboxes
The color scheme is off and we need to have the check turn into a minus sign if items are unchecked

robertandremitchell:
# PULL REQUEST

## Summary
This does a few things:
* Changes the checkboxes to the the Icon.Check from USWDS
* Adds the indeterminate state with the Icon.Remove from USWDS
* Bonus: Fixes the Expand Less / Expand More arrows to change direction based on opening/closing the Accordions.

## Related Issue
Fixes #2411 

## Additional Information
As far as I can tell, the last work on the UI should be to add borderlines in the accordion and to make every other row gray. There's some CSS code that should be doing this; I tried to tweak it a bit in this ticket without much success, so wanted to break that into a new ticket instead.

## Checklist

### DIFF
--- a/containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
+++ b/containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
@@ -89,10 +89,7 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
       ? [
           {
             title: (
-              <div
-                className="accordion-header display-flex flex-no-wrap flex-align-start"
-                onClick={handleToggleExpand}
-              >
+              <div className="accordion-header display-flex flex-no-wrap flex-align-start">
                 <div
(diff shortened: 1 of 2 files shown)


lg-king:
The checkboxes look great! 

A couple other things I noticed in general (in addition to the borderlines in the table): 
- The background color should be rgba(247, 249, 250, 1) (it's "gray-cool-2" in USWDS colors)
- The button copy should be "Apply changes" (lowercase "c")
- The cancel button should be the secondary button type (white with blue outline)
- Right now clicking anywhere in the accordion header (including the checkbox) expands/collapses the accordion. Can we change this so it's just the area by the expand/collapse carat?

lg-king (approved):

PR #2459, opened 2024-08-28
TEFCA Customize Query Accordion Updates

### LINKED ISSUE #2449: TEFCA Customize Query Accordion UI fixes

robertandremitchell:
# PULL REQUEST

## Summary
This fixes up the last few nits in the accordion, like the lines between rows and the perimeter and different colored rows every other line.

This also has a few of the other updates raised by @lg-king regarding background color, button color and copy.

The one that is still bedeviling me is the Accordion behavior to only open and close with the click of the ExpandMore/ExpandLess element. I will keep digging, but if anyone has any ideas, all ears. At most, I've figured out how to get the arrow to change direction when we click it, which could be a decent middle ground until we could solve it.

Per accessibility guidance, we may not want to disable making the whole accordion clickable to expand, or we might need to look into an outside-the-box accordion: https://designsystem.digital.gov/components/accordion/#usability-guidance.

## Related Issue
Fixes #2449 

## Acceptance Criteria

## Additional Information
I changed from grid to table on the accordion, because some of the reading on css grid I found was that it wasn't optimal for handling nth row type changes we would need to properly format the rows. I kind of prefer grid, but I wasn't making really any progress in getting the alternate coloring to work.

## Checklist

### DIFF
--- a/containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
+++ b/containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
@@ -120,7 +120,8 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
                     cursor: "pointer",
                     borderRadius: "4px",
                   }}
-                  onClick={() =>
+                  onClick={(e) => {
+                    e.stopPropagation();
                     handleSelectAllChange(
                       items,
                       (updatedItems) =>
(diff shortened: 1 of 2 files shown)


lg-king (approved):

lg-king:
I think the most important thing with the accordion is that the checkbox itself doesn't expand/collapse the accordion if possible.

robertandremitchell:
> I think the most important thing with the accordion is that the checkbox itself doesn't expand/collapse the accordion if possible.

makes sense. should be fixed now!

PR #2460, opened 2024-08-28
Support pgSQL Infrastructure

### LINKED ISSUE #2436: Add dev infrastructure for Postgres
Now that the TEFCA Connector will interact with a Postgres DB we need to out the necessary tooling in place to support local development. 

- [ ] `docker-compose.yaml` is updated to spin up an instance of Postgres and values required to connect to PG are set as environment variables on the TEFCA Viewer
- [ ] `npm run dev` is updated so that it spins up a PG DB and sets environment variables appropriately on a dev instance of the app running directly from source code NOT docker. This may require reviewing and updating `docker-compose-dev.yaml`
- [ ] Makefile is reviewed and updated as necessary so that `make` commands spin up PG as needed.

bamader:
# PULL REQUEST

## Summary
This PR modifies the docker_compose file and the .env file (for `npm run dev` to use as its config) to add support for instantiating a postgres database. Currently, since there's nothing to migrate, the DB just kind of sits there, but I added some relevant commented instructions about how we'll need to use the docker entrypoint once we have the migrations figured out.

## Related Issue
Fixes #2436 

## Additional Information
Particularly curious to see how the entrypoint scripts directory will line up with @m-goggins 's work on migrating the SQLite file. Let me know if I can change this to support you in some way.

### DIFF
--- a/containers/tefca-viewer/docker-compose.yaml
+++ b/containers/tefca-viewer/docker-compose.yaml
@@ -1,4 +1,29 @@
 services:
+  # PostgreSQL DB for custom query and value set storage
+  db:
+    image: "postgres:alpine"
+    ports:
+      - "5432:5432"
+    # Once we have the migrations set up, we may need to use a mounted volume
+    # copy to get them into the container and create our table structure.
+    # The excerpt below is copied from the ECR Viewer, which has the most
+    # similar structure to what we're doing.
+    # volumes:
+    #  - ./sql/:/docker-entrypoint-initdb.d/
+    #  - ./seed-scripts/sql/:/docker-entrypoint-initdb.d/
+    #  - ./seed-scripts/sql/.pgpass/:/usr/local/lib/.pgpass
+    environment:
+      - POSTGRES_USER=postgres
+      - PGUSER=postgres
+      - POSTGRES_PASSWORD=pw
+      - POSTGRES_DB=tefca_db
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready"]
+      interval: 2s
+      timeout: 5s
+      retries: 20
+
+  # Next.js app
   tefca-viewer:
     platform: linux/amd64
     build:
@@ -7,4 +32,8 @@ services:
     ports:
       - "3000:3000"
     environment:
-      NODE_ENV: production
+      - DATABASE_URL=${DATABASE_URL:-postgres://postgres:pw@db:5432/tefca_db}
+      - NODE_ENV=production
+    depends_on:
+      db:
+        condition: service_healthy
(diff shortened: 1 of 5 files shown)


containers/tefca-viewer/docker-compose.yaml
```diff
@@ -7,4 +32,8 @@ services:
     ports:
       - "3000:3000"
     environment:
-      NODE_ENV: production
+      - DATABASE_URL=${DATABASE_URL:-postgres://postgres:pw@db:5432/tefca_db}
```
DanPaseltiner, line 35: I'm confused why we need the URL value set to an env var in two places.

DanPaseltiner (changes requested):
Unless I'm missing something these changes won't ensure that we spin up Postgres with `npm run dev`. I believe we need to modify the `dev` definition in `package.json`. This is where we can use `docker-compose-dev` to spin up just Postgres and then start the app directly from source. Looks like there is a fair amount of junk we can remove from `docker-compose-dev`.

@m-goggins are you using `dev-win`? Looks like it has reverted to be equivalent to `dev`. If no one if using this command we can probably remove it.

Can we evaluate if we still need to `NODE_TLS_REJECT_UNAUTHORIZED=0`? I believe since all of the FHIR servers we are working with HTTPS this is no longer necessary.

bamader:
Ugh you're totally right, when I ran it on my local it seemed to work fine but that's because it was using a prebuilt, cached version of the DB. Will need to modify the underlying script. Thanks for catching!

robertandremitchell (approved):
makes sense to me and looks in line with ECR (including the odd situation dan raised with defining the DB URL.

my guess is that the env is needed more so for tests, or at least I recall some very funky errors I ran into months back, but idk.

bamader:
Yeah Rob's right, it's mostly for testing. The formal definition in the `docker-compose` file also uses variable substitution which is what allows the environment for testing to substitute if needed.

DanPaseltiner:
> Yeah Rob's right, it's mostly for testing (and was indeed cribbed from the ecr viewer). The formal definition in the `docker-compose` file also uses variable substitution which is what allows the environment for testing to substitute if needed.
 
Would it be worth trying something like [this](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute) to reference `.env` from `docker-compose.yaml`? I think this would let us define env vars in a single place. Also we could move all of the PG values in there and use the same `.env` in both services, this would have the benefit of providing these values to the app which will need them to create a connection.

bamader:
Yup I like that. I'll migrate env variables as part of the docker compose dev cleanup.

m-goggins:
I'm still using `dev-win`. The `NODE_TLS_REJECT_UNAUTHORIZED=0` from `dev` doesn't work on windows but I have a local .env file to do the same thing. Please don't delete!

bamader:
Deleting the cached images and re-launching with a dev compose worked like a charm. 

Thanks for chiming in on `dev-win` Marcelle, I'll make sure to edit that for PGDB spinup as well. I'm like 99% sure we can straight up delete the NODE_TLS_REJECT variable altogether, and if it's causing difficulties for you, maybe that's just the best play here?

DanPaseltiner (approved):
Looks right to me. Before merging @m-goggins as our resident Windows user can you please test both `dev` and `dev-win` on your local? We are no no longer

PR #2478, opened 2024-08-30
Run ECR tests, when fhir converter changes

nickbristow:
# PULL REQUEST

## Summary
Because the ecr viewer relies on things being converted a certain way, run ecr-viewer tests when the fhir converter changes


robertmitchellv (approved):
perfect! looks good 👍🏻

PR #2485, opened 2024-09-03
Launch flyway with docker

### LINKED ISSUE #2441: Begin running Flyway as part of application start up
When the TEFCA Connector spins up Flyway should be run.

We should consider if Flyway should be run

1. via Docker (our preference)
2. at the application level with in the TEFCA Connector

For the scope of this ticket let's just use a dummy migration that creates a single dummy table with one column.

bamader:
# PULL REQUEST

## Summary
This PR gets Flyway going with the tefca viewer's postgres DB using the migration Marcelle defined for our tables. I've opted to use the `flyway-docker` preferred flyway repo structure, which just meant that our migrations folder is now the `/sql` subdirectory within `/flyway`.

## Related Issue
Fixes #2441 

## Additional Information
I encountered a lot of connection refusal errors and timeouts when pulling the latest version of flyway (as did some users reporting on github), so I'm pinning us to the relatively recent 10.16 for alpine. It came out in the last two months and has no connection problems resolving underlying drivers, so I don't think we're losing anything, but just calling it out.

### DIFF
--- a/containers/tefca-viewer/flyway/conf/flyway.conf
+++ b/containers/tefca-viewer/flyway/conf/flyway.conf
@@ -1,0 +1,5 @@
+flyway.url=jdbc:postgresql://db:5432/tefca_db
+flyway.locations=filesystem:/flyway/sql
+flyway.user=postgres
+flyway.password=pw
+flyway.baselineOnMigrate=false
--- a/containers/tefca-viewer/docker-compose-dev.yaml
+++ b/containers/tefca-viewer/docker-compose-dev.yaml
@@ -1,16 +1,19 @@
 services:
+  # Flyway migrations and DB version control
+  flyway:
+    image: flyway/flyway:10.16-alpine
+    command: -configFiles=/flyway/conf/flyway.conf -schemas=public -connectRetries=60 migrate
+    volumes:
+      - ./flyway/sql:/flyway/sql
+      - ./flyway/conf/flyway.conf:/flyway/conf/flyway.conf
+    depends_on:
+      - db
+
+  # Postgresql DB
   db:
     image: "postgres:alpine"
     ports:
       - "5432:5432"
-    # Once we have the migrations set up, we may need to use a mounted volume
-    # copy to get them into the container and create our table structure.
-    # The excerpt below is copied from the ECR Viewer, which has the most
-    # similar structure to what we're doing.
-    # volumes:
-    #  - ./sql/:/docker-entrypoint-initdb.d/
-    #  - ./seed-scripts/sql/:/docker-entrypoint-initdb.d/
-    #  - ./seed-scripts/sql/.pgpass/:/usr/local/lib/.pgpass
     env_file:
       - "./tefca.env"
     healthcheck:
--- a/containers/tefca-viewer/docker-compose.yaml
+++ b/containers/tefca-viewer/docker-compose.yaml
@@ -14,14 +14,6 @@ services:
     image: "postgres:alpine"
     ports:
       - "5432:5432"
-    # Once we have the migrations set up, we may need to use a mounted volume
-    # copy to get them into the container and create our table structure.
-    # The excerpt below is copied from the ECR Viewer, which has the most
-    # similar structure to what we're doing.
-    # volumes:
-    #  - ./sql/:/docker-entrypoint-initdb.d/
-    #  - ./seed-scripts/sql/:/docker-entrypoint-initdb.d/
-    #  - ./seed-scripts/sql/.pgpass/:/usr/local/lib/.pgpass
     env_file:
       - "./tefca.env"
     healthcheck:
--- a/containers/tefca-viewer/docker-compose.yaml
+++ b/containers/tefca-viewer/docker-compose.yaml
@@ -1,4 +1,14 @@
 services:
+  # Flyway migrations and DB version control
+  flyway:
+    image: flyway/flyway:10.16-alpine
+    command: -configFiles=/flyway/conf/flyway.conf -schemas=public -connectRetries=60 migrate
+    volumes:
+      - ./flyway/sql:/flyway/sql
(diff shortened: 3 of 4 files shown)


containers/tefca-viewer/flyway/conf/flyway.conf
```diff
@@ -0,0 +1,5 @@
+flyway.url=jdbc:postgresql://db:5432/tefca_db
+flyway.locations=filesystem:/flyway/sql
+flyway.user=postgres
+flyway.password=pw
+flyway.baselineOnMigrate=false
```
m-goggins, line 5: I don't know if there's a way to sync the info that's here with the info in `tefca.env` so it's not spread across multiple places

m-goggins (commented):
Looks good to me, just left a comment about `tefca.env` and the flyway conf file

containers/tefca-viewer/flyway/conf/flyway.conf
bamader: Yeah I was mulling on that as well. Since it's a configuration loaded up from a mount rather than the postgres environment file, doing stuff like `${PG_USER}` didn't work. I could try to find some sort of pythonic way to load the environment variables at the time of file read, but flyway and the DB actually have different container environments themselves once they're pushed into docker. Short of just copying everything into the single `tefca.env` file and referencing there I'm not sure I can consolidate any more (though to be fair maybe just dumping everything into a single `.env` is how we want to go--it's unclear how serious flyway is that it needs a `.conf` or `.toml` to mount).

robertandremitchell (commented):
is this error expected?
```bash
rob@Roberts-MacBook-Pro tefca-viewer % npm run dev               

> tefca-viewer@1.0.1 dev
> docker-compose -f docker-compose-dev.yaml up && next dev

[+] Running 10/10
 ✔ flyway Pulled                                                              9.5s 
   ✔ c926b61bad3b Pull complete                                               0.7s 
   ✔ e5820a814e8c Pull complete                                               1.6s 
   ✔ 7b4bda58fd3f Pull complete                                               2.6s 
   ✔ 2e8eef6580a1 Pull complete                                               2.6s 
   ✔ a2f4070fc478 Pull complete                                               2.6s 
   ✔ 4277b7e4ad6d Pull complete                                               2.7s 
   ✔ a6f607247699 Pull complete                                               2.7s 
   ✔ c7a84da88e73 Pull complete                                               7.3s 
   ✔ e3590ede9c15 Pull complete                                               8.4s 
WARN[0009] Found orphan containers ([tefca-viewer-load_tcr_data-1 tefca-viewer-tefca-viewer-1]) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up. 
[+] Running 2/2
 ✔ Container tefca-viewer-db-1      Created                                   0.0s 
 ✔ Container tefca-viewer-flyway-1  Crea...                                   0.8s 
Attaching to db-1, flyway-1
db-1      | 
db-1      | PostgreSQL Database directory appears to contain a database; Skipping initialization
db-1      | 
db-1      | 2024-09-03 21:06:26.739 UTC [1] LOG:  starting PostgreSQL 16.4 on x86_64-pc-linux-musl, compiled by gcc (Alpine 13.2.1_git20240309) 13.2.1 20240309, 64-bit
db-1      | 2024-09-03 21:06:26.743 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db-1      | 2024-09-03 21:06:26.743 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db-1      | 2024-09-03 21:06:26.746 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db-1      | 2024-09-03 21:06:26.756 UTC [30] LOG:  database system was shut down at 2024-09-03 20:57:20 UTC
db-1      | 2024-09-03 21:06:26.778 UTC [1] LOG:  database system is ready to accept connections
flyway-1  | A more recent version of Flyway is available. Find out more about Flyway 10.17.3 at https://rd.gt/3rXiSlV
flyway-1  | 
flyway-1  | Flyway OSS Edition 10.16.0 by Redgate
flyway-1  | 
flyway-1  | See release notes here: https://rd.gt/416ObMi
flyway-1  | Database: jdbc:postgresql://db:5432/tefca_db (PostgreSQL 16.4)
flyway-1  | Schema history table "public"."flyway_schema_history" does not exist yet
flyway-1  | Successfully validated 1 migration (execution time 00:00.135s)
flyway-1  | ERROR: Found non-empty schema(s) "public" but no schema history table. Use baseline() or set baselineOnMigrate to true to initialize the schema history table.
flyway-1 exited with code 1
```

bamader:
@robertandremitchell Hmm no, that's not expected. Did you happen to have the DB spun up before trying to run the migrations? Flyway is very strict about having absolute provenance over the version history of the DB. If if it encounters literally anything not contained in its initial migration or a subsequent validation, it'll throw an error. Might be worth clearing some caches / deleting any images of the DB and trying again?

containers/tefca-viewer/flyway/conf/flyway.conf
m-goggins: I thought that might be the case. I do think flyway expects a `.conf` file vs `.env` so this is probably okay for now

m-goggins:
> @robertandremitchell Hmm no, that's not expected. Did you happen to have the DB spun up before trying to run the migrations? Flyway is very strict about having absolute provenance over the version history of the DB. If if it encounters literally anything not contained in its initial migration or a subsequent validation, it'll throw an error. Might be worth clearing some caches / deleting any images of the DB and trying again?

+1 If you were running the `pgloader` script without deleting images/containers, I'd expect you'd run into this when trying out flyway because of the reason Brandon said above.

robertandremitchell (approved):
looks good!

PR #2488, opened 2024-09-04
Begin populating codes in the TEFCA query customization page

### LINKED ISSUE #2311: Begin populating codes in the query customization page
Now that the query customization page is built out let's begin taking the limited hard coded sets of codes we currently have and using them to populate this page.

Acceptance Criteria:
When a user clicks on the `Customize Query` button
Then they are brought to the query customization page and the codes we currently have on file for the selected use case query are displayed.

robertandremitchell:
# PULL REQUEST

## Summary
This updates the `customize-queries` to begin to take the `UseCaseQueryResponse` into that response. We will need that to then update that response based off the concepts returned by the TCR, then pare down the `UseCaseQueryResponse` based off selections made on the Customize Query page. So if I understand the sequence right:
1. User selects demo of interest.
2. We run the condition against the eRSD.
3. Customize Queries displays the full gamut of potential concepts based off the condition(s) selected.

A smaller, minor fix is to also move `LoadingView.tsx` to its own component to be able to take advantage of it while rendering other pieces, as the the `/query` work gets more complex.

## Related Issue
Fixes #2311 

## Acceptance Criteria

## Additional Information
My original intention was to do more with the demo query data, but I was not actually sure if the sequencing of the work was correct, and if we wanted to get further along with the the eRSD DB call work before implementing it as a full replacement for our demo work.

Basically, I started thinking of how to retrofit our dummy data and realized it would be a lot of work when I think we are closer to maybe just all focusing on the DB connection the updating/removing our dummy data entirely? Curious if y'all think that makes sense.

I could pull data for the sample queries, but idk if we would want to get further the eRSD SQL queries and then just that as the basis for updating.

I think my main concern is releasing the full branch in what will not be a completed state. Right now, the customize query will take you to that page and display just dummy data regardless of demo type. We could implement everything as-is, then just hide the button when we release the full feature branch, I suppose.

## Checklist

### DIFF
--- a/containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
+++ b/containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
@@ -5,8 +5,10 @@ import { Accordion, Button, Icon } from "@trussworks/react-uswds";
 import { AccordianSection } from "../../query/component-utils";
 import { ValueSet } from "../../constants";
 import { AccordionItemProps } from "@trussworks/react-uswds/lib/components/Accordion/Accordion";
+import { UseCaseQueryResponse } from "../../query-service";
 
 interface CustomizeQueryProps {
+  useCaseQueryResponse: UseCaseQueryResponse;
   queryType: string;
   ValueSet: ValueSet;
   goBack: () => void;
@@ -15,12 +17,14 @@ interface CustomizeQueryProps {
 /**
  * CustomizeQuery component for displaying and customizing query details.
  * @param root0 - The properties object.
+ * @param root0.useCaseQueryResponse - The response from the query service.
  * @param root0.queryType - The type of the query.
  * @param root0.ValueSet - The value set of labs, conditions, and medications.
  * @param root0.goBack - Back button to go from "customize-queries" to "search" component.
  * @returns The CustomizeQuery component.
  */
 const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
+  useCaseQueryResponse,
   queryType,
   ValueSet,
   goBack,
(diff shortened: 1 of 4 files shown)


containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -30,6 +34,10 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
   const [valueSetState, setValueSetState] = useState<ValueSet>(ValueSet);
   const [isExpanded, setIsExpanded] = useState(true);
 
+  if (!useCaseQueryResponse) {
+    return <div>Loading...</div>;
```
fzhao99, line 38: Do we think it's appropriate here to pull out a loading component that we could define / reuse? See that we have an used `LoadingView` component in the `query/page.tsx` file that we could maybe start with.

Would make it easier for us to iterate on that loading state in the future (with a gif / loading icon), plus would give us a starting point for loading states from here

containers/tefca-viewer/src/app/query/components/SearchForm.tsx
```diff
@@ -160,12 +160,20 @@ const SearchForm: React.FC<SearchFormProps> = ({
     setOriginalRequest(originalRequest);
     const queryResponse = await UseCaseQuery(originalRequest);
     setUseCaseQueryResponse(queryResponse);
-    if (!queryResponse.Patient || queryResponse.Patient.length === 0) {
-      setMode("no-patients");
-    } else if (queryResponse.Patient.length === 1) {
-      setMode("results");
+    // Check if it's a customize query or a standard search
+    if (customizeQuery) {
```
fzhao99, line 164: question mostly for my understanding: have inferred (but please correct if wrong!) that we have a set of pre-configured query responses for demo purposes that we're building out customization for here. 

what's the shape of the predefined queries and would there be any telltale signs for a customized vs dummy query? Trying to get an understanding of how we're using some of these terms / what their shapes look like. Doubly appreciate if there's a type contract / schema definition anywhere that you could point me to.

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
robertandremitchell: 100% agree

robertandremitchell: added!

containers/tefca-viewer/src/app/query/components/SearchForm.tsx
robertandremitchell: for higher viz: we discussed during a pairing that doing more work to build out the demo seemed potentially at odds or better handled by focusing on the TCR work.

curious @bamader's and @m-goggins' thoughts on that, or if we want to have the customize query fully built out.

PR #2491, opened 2024-09-04
fix spacing on the query results page

fzhao99:
# PULL REQUEST

## Summary
I was playing around with the query connector demo site and noticed that the results page had some wonky spacing issues interacting with the accordions. Was annoyed by it so went ahead and fixed it as a starter PR. Resisting the urge to fix the view on mobile viewports for now.

Pulled a few spacing variables from Figma for ease of use in future frontend PR's.

Using the 12 column grid system to translate design intent to the spacing, which is implemented [in USWDS here](https://designsystem.digital.gov/utilities/layout-grid/) and is something I think we should formalize. A brief intro for the overall approach is [here](https://medium.com/@designwithabishek/a-beginners-guide-to-grid-system-57045b3f7035)

## Related Issue 
Didn't find a bug issue but going to pretend there was a starter bug ticket / flash my new team member badge as permission to fix something arbitrarily :). Won't work on something that isn't already prioritized in the future!

Before: 
https://github.com/user-attachments/assets/94049273-414f-495c-9bc7-b639e46ab765

After:
https://github.com/user-attachments/assets/9bc27375-b9d9-4036-83bc-21238e15dba6

### DIFF
--- a/containers/tefca-viewer/src/styles/custom-styles.scss
+++ b/containers/tefca-viewer/src/styles/custom-styles.scss
@@ -150,15 +150,13 @@ h4 {
 
 .main-container {
   max-width: calc($ecr-viewer-max-width - 2 * $ecr-viewer-content-margin);
-  padding: 0 0 4.5rem 0;
+  padding: 3rem 0 4.5rem 0;
 }
-
 
 .nav-wrapper {
   display: flex;
   flex-direction: column;
   align-items: flex-start;
-  margin-top: 3rem;
 }
 
 .sticky-nav {
--- a/containers/tefca-viewer/src/styles/custom-styles.scss
+++ b/containers/tefca-viewer/src/styles/custom-styles.scss
@@ -2,7 +2,7 @@
 
 // taken as values from Figma
 $ecr-viewer-max-width: 90rem;
-$ecr-viewer-content-margin: 13.75rem;
+$ecr-viewer-content-margin: 6.5rem;
 
 .section__line {
   border-top: 1px solid #99deea;
--- a/containers/tefca-viewer/src/styles/custom-styles.scss
+++ b/containers/tefca-viewer/src/styles/custom-styles.scss
@@ -1,5 +1,9 @@
 @use "uswds-core" as *;
 
+// taken as values from Figma
+$ecr-viewer-max-width: 90rem;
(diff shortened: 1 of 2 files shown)


containers/tefca-viewer/src/styles/custom-styles.scss
```diff
@@ -1,5 +1,9 @@
 @use "uswds-core" as *;
 
+// taken as values from Figma
+$ecr-viewer-max-width: 90rem;
+$ecr-viewer-content-margin: 6.5rem;
```
fzhao99, line 5: pulling these as initial values based on the values in Figma for us to propagate on other pages / pull into a template component so that the spacing works correctly on other pages.

Screenshot 2024-09-04 at 5 39 38 PM


Screenshot 2024-09-04 at 5 38 29 PM

containers/tefca-viewer/src/styles/custom-styles.scss
```diff
@@ -1,5 +1,9 @@
 @use "uswds-core" as *;
 
+// taken as values from Figma
+$ecr-viewer-max-width: 90rem;
+$ecr-viewer-content-margin: 13.75rem;
```
fzhao99, line 5: Pulling these values from Figma so we can propagate them across other pages as appropriate. Probably worth reworking some common spacing elements into a barebones page component so that we're building that once for all pages 

Screenshot 2024-09-04 at 5 39 38 PM

containers/tefca-viewer/src/styles/custom-styles.scss
fzhao99, line 5: Pulling these values from Figma below. I went ahead and halved the margins for now since the content in the accordions were getting cut off / the white space was a bit wide, but an easy fix if we want to change it back.

Probably worth (if we're not already doing this!) pulling some of these common spacing elements into a skeleton page template component that other individual pages can implement with their page-specific UI so that we're only solving this once at the root of the app.

Screenshot 2024-09-04 at 5 39 38 PM

m-goggins (approved):
Looks good to me, thanks for finding and fixing! We're all fairly new to front end work so we haven't honed 'best practices' for application design.

For reference, we usually tag Laura (lg-king) as a reviewer for any design changes, though she is out this week so feel free to skip on this one.

PR #2496, opened 2024-09-05
Remove pgloader, insert TCR data via Flyway migration, add migration to populate query tables

m-goggins:
# PULL REQUEST

## Summary
This PR removes the `pgloader` bash script from the docker-compose file and replaces it with migration V01_02 and adds a migration (V01_03) to populate the query tables based on data in the TCR. Now, every condition has a query that could be used by the TEFCA Connector.

V01_02 is an ugly file that is the result of dumping the `ersd.db` file and updating for compatibility with postgres. The file is intended to serve as a temporary measure to get data into the TEFCA db because the pgloader option was not viable. We missed an obvious flaw in the flyway + pgloader set up; _all_ flyway migrations run before the pgloader script:

What we wanted:
1. Flyway runs migration V01_01
2. pgloader loads TCR data
3. Flyway runs migration V01_02 using the loaded TCR data

What we got:
1. Flyway runs migration V01_01
2. Flyway attempts to run migration V01_02 using the loaded TCR data but fails because there is no data
3. pgloader loads TCR data but the query tables do not contain any data

## Related Issue
Fixes #

## Acceptance Criteria

## Additional Information

## Checklist


robertandremitchell (approved):
code worked for me.

that's crummy about pgloader. I;m guessing there's not really a benefit to trying to dynamically create V01_02 to load data. I don't think it matters too much, beyond needing to up this any time we update the eRSD database, which in theory shouldn't be frequent, but I wonder if we would want to then maybe have a script there that outputs the data to migrations?

In which case, I think we've talked before about it, but I do think we should move eRSD seeding to its own root folder.

m-goggins:
> code worked for me.
> 
> that's crummy about pgloader. I;m guessing there's not really a benefit to trying to dynamically create V01_02 to load data. I don't think it matters too much, beyond needing to up this any time we update the eRSD database, which in theory shouldn't be frequent, but I wonder if we would want to then maybe have a script there that outputs the data to migrations?
> 
> In which case, I think we've talked before about it, but I do think we should move eRSD seeding to its own root folder.

@robertandremitchell We could have a script do that, though using `sqlite3 .dump` does most of the work. I had to do a little cleanup to ensure the code is compatible with postgresql since the output is compatible with sqlite and there were a couple of slight differences. I think the fact that we want to port the TCR data and cannot do so easily means we should probably revisit the idea of making an API request to the TCR. I know Dan is against spinning up two services but I'd like to hear more about the issues this would pose for a STLT before we totally rule it out.

This is a separate note and probably could be another ticket but when I did the `.dump` I noticed that the TCR has both a `crosswalk` and an `icd_crosswalk` with the same data. I ended up deleting the `crosswalk` one from the script to match the [schema](https://app.diagrams.net/#G1i3Bzsu0moS5I7_Qw_Tw75uTMAlWipo2r#%7B%22pageId%22%3A%22WHVhyQuvTKeJfM7__wGX%22%7D), which has `icd_crosswalk`.

bamader (approved):
Looks great and lines up with all those sweet sweet objectives we talked about. FWIW I'm also definitely interested in revisiting the conversation around spinning up two services given the struggle we've had. It seems like going from 0 to 1 is a way harder barrier than going 1 to 2, but that can be a separate conversation.

PR #2529, opened 2024-09-10
add dev script for a local version of `tefca.env`

fzhao99:
adds in a package.json script that enables running the Next app locally against the new db config work


bamader (approved):
Looks great, good catches on hot reloading and thanks for fixing that environment fetch for the port number!

PR #2507, opened 2024-09-06
Marcelle/tefca db client

m-goggins:
# PULL REQUEST

## Summary
This PR creates the DB client that allows the front end to securely interact with the back end DB through server actions. There's also a smattering of new data structures and parameter changes to the UI to set the groundwork for populating the page with real query value set data.

## Related Issue
Fixes #2438 
Fixes #2510 

## Additional Information
Resolving the merge conflicts was a lossy nightmare so I'm proposing we don't land anything new in these areas until this PR lands and we can work from fresh `mains` again.

PR #2530, opened 2024-09-10
Ensure all hardcoded values are included in queries

### LINKED ISSUE #2520: Write migration to insert hardcoded values in existing demo queries

m-goggins:
# PULL REQUEST

## Summary
This PR adds a migration to insert and link all of the values that we have hard coded in our demo queries right now. For example, when you use the hard coded query for Chlamydia, you included medications in the query. The TCR query for Chlamydia, on the other hand, does not include medications. This PR addresses this issue by adding linking additional value sets to the TCR condition for Chlamydia so that when you run a query for Chlamydia using the data from the TCR, you will also get the hard coded concepts. 

Importantly, all newly linked concepts are added via new valuesets, e.g., `Missing Chlamydia mrtc` rather than the TCR Chlamydia mrtc value set, so if we want to undo these linkages in the future (to keep the base queries true to the data in the TCR only), we can easily unlink by changing the condition_to_valueset links. Additionally, some concepts were already included in the TCR data but were not linked to the condition in the way that they are in our hard coded queries, e.g., the gonorrhea test was not linked to chlamydia.

I also added `Newborn Screening`, `Cancer (Leukemia)`, and `Social Determinants of Health` as new conditions, though the Social Determinants of Health one does not actually run using any codes to make the query; it just uses `category=social-history`.

## Related Issue
Fixes #2520 

## Acceptance Criteria

## Additional Information
Once the PR for #2507 lands, I'll do a quick follow up PR to update the names of the queries in `constants.ts`.

## Checklist


fzhao99 (approved):
🚢

PR #2534, opened 2024-09-10
Fix S3 config

### LINKED ISSUE #2537: Fix S3 writes for running instances
## What needs to be done

The recent work on S3 emulation for local host has inadvertently broken writing to S3 for deployed applications. 

## Why it needs to be done

We need to fix this so that the eCR Viewer works in an AWS-deployed environment.

## Acceptance Criteria

- [ ] eCR Viewer can write to a live S3 bucket
- [ ] Local dev emulation for S3 still works as intended

## Technical Notes

The DevOps team spent some time digging into this already, and they think the following line is the issue: https://github.com/CDCgov/phdi/commit/3df0e4f9617a51f664525ca81b823e358543facc#diff-9e5aa20f890ffc305765c4741f858a502b19409897d892ff42604f70b180f16fL12

Their suggested solution is something like the following:

```javascript
const s3Client = process.env.APP_ENV == 'dev' ? new S3Client({
  region: process.env.AWS_REGION,
  endpoint: process.env.AWS_CUSTOM_ENDPOINT,
  forcePathStyle: process.env.AWS_CUSTOM_ENDPOINT !== undefined,
}) : new S3Client({ region: process.env.AWS_REGION });
```

austin-hall-skylight:
# PULL REQUEST

## Summary
Remove AWS_CUSTOM_ENDPOINT, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AZURE_STORAGE_CONNECTION_STRING from .env to prevent them from being deployed. Updates S3 client constructor.

## Related Issue
Fixes #2537 

## Checklist


containers/ecr-viewer/.env
```diff
@@ -1,13 +1,6 @@
...
-AWS_CUSTOM_ENDPOINT="http://localhost:4566"
 AWS_REGION=us-east-1
 ECR_BUCKET_NAME=ecr-viewer-files
-# These are the default credentials for localstack
-AWS_ACCESS_KEY_ID=123
-AWS_SECRET_ACCESS_KEY=xyz
-# NOT A REAL CONNECTION STRING! This is the default connection string for Azurite
-# https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio%2Cblob-storage#well-known-storage-account-and-key
-AZURE_STORAGE_CONNECTION_STRING="DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://0.0.0.0:10000/devstoreaccount1;"
 AZURE_CONTAINER_NAME=ecr-viewer-files
```
angelathe, line 4: I'm wondering if we should remove the bucket & container names as well here since these are only for dev & will be overwritten by the actual container/bucket names in the TF? @emmastephenson

emmastephenson: Yes, I think that's the right call! But cc-ing @alismx just to make sure.

austin-hall-skylight: Ok good call. I wasn't sure since the aws bucket name was already in there before I made my changes but I'm ready to take these out as soon as we have a consensus.

alismx: We set those values later so it's not needed here. With that said, do y'all devs have a place for your local env vars once these are removed? I haven't looked at this outside of the AWS/Azure context.

austin-hall-skylight: Yep, we have a .env.local that doesn't get checked in.

angelathe (approved):
LGTM!

angelathe:
Oops, jumped the gun a bit -- you may need to add that Azure container name env var back in to fix that failing test

PR #2541, opened 2024-09-11
introduce new accordion component and refactor  results table

fzhao99:
# PULL REQUEST

## Related Issue
Part one of [2523](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/2523) split out for size / ease of review. An upcoming PR will do something similar to Customize Query using this componentization code.

## Summary
Defines a new accordion wrapper component to the Truss implementation and then propagates it to the Results view, as well as starting to modularize some of our CSS classes. Various styling tweaks to get things aligned with the existing views.

The implementation is based on the [Atomic Design approach](https://atomicdesign.bradfrost.com/chapter-2/) to building UI: basically, UI can be taxonimized between UI building block components (atoms) that are agnostic of context (think buttons, accordions, banners) vs the final pages that they arrange themselves on to make an experience. Will try to introduce some of this philosophy into the way we're building out our UI moving forward.

## Acceptance Criteria
Should be a no-op on the Results View since this is just a frontend component refactor


containers/tefca-viewer/src/styles/custom-styles.scss
```diff
@@ -59,16 +55,6 @@ h4 {
   font-weight: bold;
 }
 
-.info-container>.usa-accordion__heading>.usa-accordion__button {
```
fzhao99, line 62: moving these styles into the smaller, modularized files to make this css file less giant

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -236,7 +235,8 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
             id: items[0].author + ":" + items[0].system,
             className: "accordion-item",
             content: (
-              <AccordianSection>
```
fzhao99, line 239: This was the one place that was still using this component as used to exist in `component-utils.tsx`, so swapping it out with the markup so i can delete that file

containers/tefca-viewer/src/app/query/components/ResultsViewTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+    ? queryResponse.DiagnosticReport
+    : null;
+  const medicationRequests = queryResponse.MedicationRequest
+    ? queryResponse.MedicationRequest
+    : null;
+
+  const accordionItems = [
+    {
+      title: "Patient Info",
+      id: "patient-info",
```
m-goggins, line 43: Under our current designs we have the Accordion header say Patient Info and the subheader is Demographics. Personally, doesn't matter to me but we should have @lg-king review this PR.

image

containers/tefca-viewer/src/styles/custom-styles.scss
m-goggins: Yay, thank you! I suspect a lot of them are cruft from various versions of components we've created and some can be deleted

m-goggins (commented):
Thanks for fixing up our cruft! Left one comment about Patient Info vs. Demographics

containers/tefca-viewer/src/app/query/components/ResultsViewTable.tsx
fzhao99: Thanks for catching! fixed

lg-king (approved):

PR #2544, opened 2024-09-11
Fix modal and test flow

### LINKED ISSUE #2538: Bug: fix`test` user journey option selection issues

bamader:
# PULL REQUEST

## Summary
This PR fixes the dropdown issues on the test user journey in the Query and Patient selection options. Both now render correctly and allow values to be successfully changed. This PR also addresses an issue where the demo journey search form was defaulting to the cancer use case each time instead of what was chosen in the modal; now, search params are again used to set the correct initial value.

## Related Issue
Fixes #2538


containers/tefca-viewer/src/app/query/test/page.tsx
```diff
@@ -32,13 +32,9 @@ const Query: React.FC = () => {
             setUseCaseQueryResponse={setUseCaseQueryResponse}
             setOriginalRequest={setOriginalRequest}
             userJourney="test"
-            setUseCase={function (useCase: USE_CASES): void {
```
fzhao99, line 35: 🙇‍♂️

fzhao99 (approved):

containers/tefca-viewer/src/app/query/test/page.tsx
```diff
@@ -32,13 +32,9 @@ const Query: React.FC = () => {
...
-              throw new Error("Function not implemented.");
-            }}
             setQueryType={function (queryType: string): void {
               throw new Error("Function not implemented.");
```
m-goggins, line 36:
```suggestion
              throw new Error("Function not implemented.");
```
I don't think we want to throw this error. Possibly a leftover?

m-goggins (changes requested):

containers/tefca-viewer/src/app/query/test/page.tsx
bamader: Shockingly not a leftover but actually a deliberate part of the built CustomizeQuery component. In practice, this function does absolutely nothing though; the demo user journey just uses `setQueryType` to assign state so that the right `queryName` can be looked up in `CustomizeQuery`. Since the customization flow isn't available in the `test` workflow, there's actually no effect at all; you can still hit `Fill Fields` and submit patients and get results back. That being said, because it does absolutely nothing, happy to replace with a `void` do nothing function altogether.

m-goggins (approved):

PR #2545, opened 2024-09-11
TEFCA Custom Query Multiple Accordion Support

### LINKED ISSUE #2522: Write back end function that separates DB returned rows by coding system

### LINKED ISSUE #2523: Modify CustomizeQuery component to show one accordion per coding scheme on each tab.

robertandremitchell:
# PULL REQUEST

## Summary
Updates code to use multiple accordions based on the unique combination of code system and author. We need to include author in the event we have future authors.

Does some styling / frontend refactoring work to match the designs and make the frontend code a little more readable. Might add a followon PR to fix a bit more of the frontend code, but most of the functionality should be there.

## Related Issue
Fixes #2522  and fixes #2523 

## Acceptance Criteria
2522 
Now that we have a DB client set up and can successfully retrieve rows from the postgres instance in TEFCA, we want to be able to filter and group these results according to the coding scheme to which they belong. To do this, we need a back end server function in the database-service.ts file that separates results into different groups based on their coding scheme. This function should be called after splitting up the rows into their respective clinical service types (labs, conditions, and meds).

2523
Now that we have a DB client set up and can successfully retrieve rows from the postgres instance in TEFCA, we want to be able to filter and group these results according to the coding scheme to which they belong. To do this, we need a back end server function in the database-service.ts file that separates results into different groups based on their coding scheme. This function should be called after splitting up the rows into their respective clinical service types (labs, conditions, and meds).

## Additional Information
Do we want to shorten code system URLs, like SNOMED, LOINC, RxNorm, ICD-10, ICD-9, etc? I lean toward no, mostly because when we ingest more systems, we may end up having to maintain that, as that occurred even within the same eCR in past work I did where the URLs differed slightly enough.

## Checklist


containers/tefca-viewer/src/app/database-service.ts
```diff
@@ -80,21 +80,37 @@ export const filterQueryRows = async (
...
-      include: r["include"],
-      author: r["author"],
-    };
-    return vsTranslation;
-  });
-  return vsItems;
+  // Group by author and code_system
+  const grouped = rows.reduce(
+    (acc, row) => {
+      const groupKey = `${row.author}:${row.code_system}`;
```
fzhao99, line 92: could we do some sort of check that `row.author` and `row.code_system` are defined / properly formed here? Looks like we don't have any type guarantees since this is coming from the db, but in the event that the query result is malformed, might want to protect against that here.

Could imagine something like `${row?.author}:${row?.code_system}` and then skipping assignment / logging a message about a malformed query working, but if you can think of a better fallback case in the event that those are missing, feel free to remix!

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -37,7 +38,11 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
 }) => {
   const [activeTab, setActiveTab] = useState("labs");
 
-  const [valueSetState, setValueSetState] = useState<ValueSet>({
+  const [groupedValueSetState, setGroupedValueSetState] = useState<{
+    labs: { author: string; system: string; items: ValueSetItem[] }[];
```
fzhao99, line 42: Not sure what the semantic term would be here, but think this may be a good chance to define some types here that we could reuse. Something like

```
type DefinedValueSetCollection = { // or whatever the right name is
  author: string;
  system: string;
  items: ValueSetItem[];
};

type GroupedValueSet = {
  labs: DefinedValueSetCollection[];
  medications: DefinedValueSetCollection[];
  conditions: DefinedValueSetCollection[];
};
```

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -50,48 +55,85 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
...
 
-  // Handles the toggle of the 'include' state for individual items
-  const toggleInclude = (index: number) => {
-    const updatedItems = [...valueSetState[activeTab as keyof ValueSet]];
-    updatedItems[index].include = !updatedItems[index].include;
-    setValueSetState((prevState) => ({
+  // Allows all items to be selected within the entire active tab
+  const handleSelectAllForTab = (checked: boolean) => {
+    const updatedGroups = groupedValueSetState[
+      activeTab as keyof typeof groupedValueSetState
```
fzhao99, line 107: We're using this `keyof typeof groupedValueSetState` in enough places for me to think we should define it as a type. Thoughts?

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
robertandremitchell: good call; updated!

containers/tefca-viewer/src/app/database-service.ts
robertandremitchell: yeah, that's a good question. outside of checking whether they exist, I am not entirely sure. I don't know for example what authors we would expect or as noted above, we could get stuck monitoring slight name change variations in the code system URLs. 

added, curious your thoughts!

fzhao99: I think this looks good! Good point on the name change issue.

This log message might be a good thing to add some instrumentation against to understand when / where there might be queries that are requesting missing valueset data against the defined eRSD defaults. That error rate might be an interesting product metric that may expose opportunities for us to augment eRSD if we see a particular set of author / code system that shows up over and over again. 

cc: @jankimo @DanielPaseltiner

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -146,224 +186,57 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
       topCheckbox.indeterminate =
         selectedCount > 0 && selectedCount < items.length;
     }
-  }, [valueSetState, activeTab]);
-
-  const accordionItems: AccordionItemProps[] = useMemo(() => {
```
fzhao99, line 151: Moved this block of code into `CustomizeQueryAccordionHeader.tsx`

containers/tefca-viewer/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -146,224 +186,57 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
...
-                  ) : (
-                    <Icon.ExpandMore size={4} />
-                  )}
-                </div>
-              </div>
-            ),
-            id: items[0].author + ":" + items[0].system,
-            className: "accordion-item",
-            content: (
-              <div className="padding-bottom-3">
```
fzhao99, line 239: Moved this block of code into `CustomizeQueryAccordionBody.tsx`

containers/tefca-viewer/src/styles/styles.scss
```diff
@@ -5,5 +5,3 @@
 @forward "layout";
 
 @forward "custom-styles";
-
-@import "~@trussworks/react-uswds/lib/index.css";
```
fzhao99, line 9: This import was breaking some of the accordion stylings but pulling it out doesn't seem to be breaking anything? If folks can find something with the react components that is breaking, happy to add this back, but couldn't get the colors to match for the life of me with this file included

containers/tefca-viewer/src/app/page.tsx
```diff
@@ -33,7 +34,10 @@ export default function LandingPage() {
                 TEFCA, giving you access to more complete and timely data.
               </h2>
             </div>
-            <img src="/tefca-viewer/tefca-graphic.svg" />
+            <Image
```
fzhao99, line 37: This was getting caught in the linter so I went ahead and fixed it

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -274,7 +272,7 @@ export async function createBundle(
     entry: [],
   };
 
-  Object.entries(queryResponse).forEach(([key, resources]) => {
+  Object.entries(queryResponse).forEach(([_, resources]) => {
```
fzhao99, line 275: This was getting caught in the linter so I went ahead and fixed it

containers/tefca-viewer/src/app/query/components/Demographics.tsx
```diff
@@ -32,7 +32,7 @@ const Demographics: React.FC<DemographicsProps> = ({ patient }) => {
 
   return (
     <div>
-      {demographicData.map((item, index) => (
```
fzhao99, line 35: This was getting caught in the linter as an unused var so I went ahead and fixed it

m-goggins:
The frontend looks really good! I am noticing that we do not have the right grouping of value sets and codes. For example, Chlamydia medications look incorrect. I would have assumed that the value set title would be something like 'missing medications' and contain levofloxacin and azithromycin 1000MG as items in the the accordion. 

I'm 100% understanding that this could come from the way that we are inserting the hard coded data, but the name of the medication is not the name of the valueset as far as I know

containers/tefca-viewer/flyway/sql/V01_03__insert_hardcoded_values.sql
```diff
@@ -12,20 +12,20 @@ INSERT INTO valuesets VALUES('2_20240909','2','20240909','Cancer (Leukemia)','DI
...
 
 -- Missing Gonorhea valueset(s)
-INSERT INTO valuesets VALUES('7_20240909','7','20240909','Missing Gonorhea mrtc','DIBBs','mrtc');
-INSERT INTO valuesets VALUES('8_20240909','8','20240909','Missing Gonorhea dxtc','DIBBs','dxtc');
-INSERT INTO valuesets VALUES('9_20240909','9','20240909','Missing Gonorhea lotc','DIBBs','lotc');
-INSERT INTO valuesets VALUES('10_20240909','10','20240909','Missing Gonorhea lrtc','DIBBs','lrtc');
+INSERT INTO valuesets VALUES('7_20240909','7','20240909','Gonorhea Medication','DIBBs','mrtc');
+INSERT INTO valuesets VALUES('8_20240909','8','20240909','Gonorhea Diagnosis Problem','DIBBs','dxtc');
+INSERT INTO valuesets VALUES('9_20240909','9','20240909','Gonorhea Lab Organism','DIBBs','lotc');
+INSERT INTO valuesets VALUES('10_20240909','10','20240909','Gonorhea Lab Result','DIBBs','lrtc');
```
m-goggins, line 23:
```suggestion
INSERT INTO valuesets VALUES('7_20240909','7','20240909','Gonorrhea Medication','DIBBs','mrtc');
INSERT INTO valuesets VALUES('8_20240909','8','20240909','Gonorrhea Diagnosis Problem','DIBBs','dxtc');
INSERT INTO valuesets VALUES('9_20240909','9','20240909','Gonorrhea Lab Organism','DIBBs','lotc');
INSERT INTO valuesets VALUES('10_20240909','10','20240909','Gonorrhea Lab Result','DIBBs','lrtc');
```

this is my bad but I always misspell these

m-goggins (approved):

fzhao99:
closing in favor of the work started here: https://github.com/CDCgov/phdi/pull/2564

PR #2547, opened 2024-09-12
Remove modal, change default state

### LINKED ISSUE #2525: Remove Query Modal
As part of our work to collapse the `demo` and `test` user experiences into a single one, users should be able to click on "Go to demo" on the landing page and it will route them directly to the query page.

bamader:
# PULL REQUEST

## Summary
This PR removes the modal from the demo user journey, instead using the `Go to the demo` button as a direct link to the SearchForm page. It also removes the useParams search state initialization of the search form, so we just always default to the first option of our demo use cases.

## Related Issue
Fixes #2525

### DIFF
--- a/containers/tefca-viewer/src/app/query/components/SearchForm.tsx
+++ b/containers/tefca-viewer/src/app/query/components/SearchForm.tsx
@@ -25,7 +25,6 @@ import {
 } from "../../query-service";
 
 import { FormatPhoneAsDigits } from "@/app/format-service";
-import { useSearchParams } from "next/navigation";
 
 interface SearchFormProps {
   setOriginalRequest: (originalRequest: UseCaseQueryRequest) => void;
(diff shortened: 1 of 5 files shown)


containers/tefca-viewer/src/app/query/components/SearchForm.tsx
```diff
@@ -54,12 +53,8 @@ const SearchForm: React.FC<SearchFormProps> = ({
   setQueryType,
   userJourney,
 }) => {
-  const params = useSearchParams();
-
   // Get the demoOption (initial selection) selected from modal via the URL
-  const [useCase, setUseCase] = useState<USE_CASES>(
-    (params.get("useCase") as USE_CASES) || "cancer",
-  );
+  const [useCase, setUseCase] = useState<USE_CASES>("cancer");
```
fzhao99, line 57: a minor quality of life suggestion: it seems to me that the chlamydia dataset has a more rich combo of labs/meds. maybe worth using that one as the default all else equal?

don't feel super strongly about it either way.

fzhao99 (approved):
🧹

robertandremitchell (approved):
looks good, agreed with Bob that I wasn't entirely sure on what the expected workflow would look like for the connectathon. I remember we talked about adding an Advanced button, but I'm not sure if this falls under "permanent change" v. "change to make things sleeker for connectathon"

bamader:
To clarify, Marcelle and I synced yesterday on what the correct workflow is for the connectathon. We do in fact want everything merged into one user journey, which means we're going to be slowly phasing out the test workflow over a series of PRs. The first step in that chain is eliminating the modal and just going straight to the search form. At that point, Rob, you're correct that we'll eventually want an "Advanced" section for choosing a FHIR server, but this is a separate ticket to be completed later after we remove field auto population.

bamader:
CDC folks actually need the testing workflow to validate their synthetic, which doesn't use the modal at all and will be accounted for in the release Marcelle just cut.

PR #2552, opened 2024-09-13
Save changes to valuesets when customizing query

### LINKED ISSUE #2512: Write helper function to format data to execute query
We can return data from the TEFCA database to retrieve all concepts for a given condition, but the data needs to be reformatted so that it can be used to actually execute a query, i.e., used the code_systems to organize the results of the TEFCA db query to look like the CustomQuery class.

The data that we get back from the TEFCA database looks like this:
```
[
{
    query_name: 'Chlamydia trachomatis infection (disorder)',
    id: 'eba3cb5f-c400-4222-b04c-7938ac2d9396',
    valueset_id: '2.16.840.1.113762.1.4.1146.609_20191227',
    concept_id: '2.16.840.1.113762.1.4.1146.609_29976007',
    include: true,
    code: '29976007',
    code_system: 'http://snomed.info/sct',
    display: 'Trachoma, initial stage (disorder)'
  },
  {
    query_name: 'Chlamydia trachomatis infection (disorder)',
    id: 'eba3cb5f-c400-4222-b04c-7938ac2d9396',
    valueset_id: '2.16.840.1.113762.1.4.1146.609_20191227',
    concept_id: '2.16.840.1.113762.1.4.1146.609_52812002',
    include: true,
    code: '52812002',
    code_system: 'http://snomed.info/sct',
    display: 'Trachoma, active stage (disorder)'
  }]
```

And we would like the data to be organized to look like the `CustomQuery` class
```
labCodes: string[] = [];
  snomedCodes: string[] = [];
  rxnormCodes: string[] = [];
  classTypeCodes: string[] = [];
```

We may consider updating the `Custom Query` class or return additional data from the TEFCA DB to make this possible.

bamader:
# Persist valueset changes in state

## Summary
This PR implements all of the work up to the point of executing a custom query. When the user presses apply on the `CustomizeQuery` page/component, inclusions/exclusions they've made to particular value sets are saved into the overall state structure maintained by the `page` workflow. Further, this PR moves some of the lower level state into the top level page component to stop re-renders when moving between modes (like search, customize, etc.). Finally, this PR implements a formatting function that transforms the statefully held collection of value set items into a `querySpec` that can eventually be used to execute a custom query.

## Related Issue
Fixes #2512

### DIFF
--- a/containers/tefca-viewer/src/app/query/components/SearchForm.tsx
+++ b/containers/tefca-viewer/src/app/query/components/SearchForm.tsx
@@ -72,8 +72,6 @@ const SearchForm: React.FC<SearchFormProps> = ({
 
   const [autofilled, setAutofilled] = useState(false); // boolean indicating if the form was autofilled, changes color if true
 
-  console.log("use case: " + useCase);
-
   // Fills fields with sample data based on the selected patientOption
   const fillFields = useCallback(
     (patientOption: PatientType, highlightAutofilled = true) => {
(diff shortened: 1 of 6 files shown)


containers/tefca-viewer/src/app/query/components/SearchForm.tsx
```diff
@@ -81,10 +84,10 @@ const SearchForm: React.FC<SearchFormProps> = ({
...
         setFhirServer(data.FhirServer as FHIR_SERVERS);
         setUseCase(data.UseCase as USE_CASES);
-        setQueryType(
-          demoQueryOptions.find((option) => option.value === data.UseCase)
-            ?.label || "",
-        );
+        // setQueryType(
+        //     demoQueryOptions.find((option) => option.value === data.UseCase)
+        //         ?.label || "",
+        // );
```
fzhao99, line 90:
```suggestion
```
Guessing this might be a leftover comment?

bamader: Actually this one is just commented out because it relates to some work Marcelle is doing removing auto-population of fields. I theoretically could take that out of my PR, but since I didn't want to cross into what she's doing I just needed to take out the part of the callback hook that modifies the important state I'm using. So it's just commented out until we see how her code chooses to resolve the auto population stuff.

fzhao99 (approved):

PR #2561, opened 2024-09-16
update database query to support multiple accordion use case

fzhao99:
# PULL REQUEST

## Summary
Updates some sql / methods in the database service to support the frontend work as described in #2523 

## Related Issue
Fixes [#2522](https://github.com/CDCgov/phdi/issues/2522) in a way that's compatible with the changes that got put in for apply changes on Friday. Adds in information about the valueset name to be used in the final display for the customize query pages

## Acceptance Criteria
Now that we have a DB client set up and can successfully retrieve rows from the postgres instance in TEFCA, we want to be able to filter and group these results according to the coding scheme to which they belong. To do this, we need a back end server function in the database-service.ts file that separates results into different groups based on their coding scheme. This function should be called after splitting up the rows into their respective clinical service types (labs, conditions, and meds).

## Additional Information
The acceptance criteria mentions splitting the data into the respective clinical service types before grouping, but after some work for apply changes that got put in on Friday, we elected to have the backend just return the list of `ValueSetItems` to the frontend and have the relevant pages do the transformation / grouping needed for each individual use case. This way, we don't couple what the database interface returns to the data structure needs of any one flow.

One question we weren't sure about in this section of code:
```
  const labCodes: string[] = vsItems
    .filter((vs) => vs.system === "http://loinc.org")
    .map((vs) => vs.code);
  const snomedCodes: string[] = vsItems
    .filter((vs) => vs.system === "http://snomed.info/sct")
    .map((vs) => vs.code);
  const rxnormCodes: string[] = vsItems
    .filter((vs) => vs.system === "http://www.nlm.nih.gov/research/umls/rxnorm")
    .map((vs) => vs.code);
```

what the impact would be in not having ICD9/10 represented / if we want to represent them?

### DIFF
--- a/containers/ingestion/description.md
+++ b/containers/ingestion/description.md
@@ -44,3 +44,56 @@ To build the Docker image for the Ingestion service from source code instead of 
 ### The API 
 
 When viewing these docs from the `/redoc` endpoint on a running instance of the ingestion service or the PHDI website, detailed documentation on the API will be available below. 
+
+## Diagrams
+
+###
+
+```mermaid
+flowchart LR
+
+  subgraph requests["Requests"]
+    direction TB
+    subgraph GET["fas:fa-download <code>GET</code>"]
+      hc["<code>/</code>\n(health check)"]
+
+    end
+    subgraph POST-standardization["fas:fa-upload <code>POST Standardization</code>"]
+      ecr["<code>/fhir/harmonization/<br>standardization/standardize_names<//code>"]
+      phones["<code>/fhir/harmonization/<br>standardization/standardize_phones<//code>"]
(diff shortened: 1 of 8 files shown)


m-goggins:
Pretty sure I already took care of this in #2545

fzhao99:
> Pretty sure I already took care of this in #2545. Do we need a separate PR?

Good question! We decided to pull out this PR from 2545 to make sure there wasn't issues for any parallel streams, since the previous PR changed the db interface pretty drastically. It also solves some merge conflict issues (which got a little gnarly in the larger PR) integrating some of the work from Friday, as well as asking the above question about the ICD-9/10 codes.

m-goggins (approved):

PR #2564, opened 2024-09-16
customize query multiple accordion implementation pt 1

fzhao99:
# PULL REQUEST

## Summary
First half of the fix for the multiple accordion work, broken out from the actual functional work for PR size reasons. Should be a no op since this just introduces styling / refactoring / tweaking code to the larger work.

## Related Issue
Part one of #2523 

## Additional Information

### DIFF
--- a/containers/tefca-viewer/src/app/database-service.ts
+++ b/containers/tefca-viewer/src/app/database-service.ts
@@ -62,11 +62,11 @@ export const getSavedQueryByName = async (name: string) => {
  */
 export const filterValueSets = async (
   vsItems: ValueSetItem[],
-  type: "labs" | "medications" | "conditions",
+  type: "labs" | "medications" | "conditions"
 ) => {
   // Assign clinical code type based on desired filter
   // Mapping is established in TCR, so follow that convention
-  let valuesetFilters;
+  let valuesetFilters: string[];
   if (type == "labs") {
     valuesetFilters = ["ostc", "lotc", "lrtc"];
   } else if (type == "medications") {
@@ -75,16 +75,16 @@ export const filterValueSets = async (
     valuesetFilters = ["dxtc", "sdtc"];
   }
   const results = vsItems.filter((vs) =>
-    valuesetFilters.includes(vs.clinicalServiceType),
+    valuesetFilters.includes(vs.clinicalServiceType)
   );
   return results;
 };
 
 /**
- * Helper function that transforms a set of database rows into a list of
- * ValueSet item structs for display on the CustomizeQuery page.
+ * Helper function that transforms and groups a set of database rows into a list of
+ * ValueSet items grouped by author and code_system for display on the CustomizeQuery page.
  * @param rows The rows returned from the DB.
- * @returns A list of ValueSetItems constructed from the DB rows.
+ * @returns A list of ValueSetItems grouped by author and system.
  */
 export const mapQueryRowsToValueSetItems = async (rows: QueryResultRow[]) => {
   const vsItems = rows.map((r) => {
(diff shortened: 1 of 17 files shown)


containers/tefca-viewer/src/app/database-service.ts
```diff
@@ -66,7 +66,7 @@ export const filterValueSets = async (
 ) => {
   // Assign clinical code type based on desired filter
   // Mapping is established in TCR, so follow that convention
-  let valuesetFilters;
+  let valuesetFilters: string[];
```
fzhao99, line 69: This was getting caught in the type system so I went ahead and fixed it

containers/tefca-viewer/src/app/page.tsx
```diff
@@ -33,7 +34,12 @@ export default function LandingPage() {
                 TEFCA, giving you access to more complete and timely data.
               </h2>
             </div>
-            <img src="/tefca-viewer/tefca-graphic.svg" />
+            <Image
```
fzhao99, line 37: This was getting caught in the linter so I went ahead and fixed it

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -274,7 +272,7 @@ export async function createBundle(
     entry: [],
   };
 
-  Object.entries(queryResponse).forEach(([key, resources]) => {
+  Object.entries(queryResponse).forEach(([_, resources]) => {
```
fzhao99, line 275: This was getting caught in the linter so I went ahead and fixed it

containers/tefca-viewer/src/app/query/components/Demographics.tsx
```diff
@@ -32,7 +32,7 @@ const Demographics: React.FC<DemographicsProps> = ({ patient }) => {
 
   return (
     <div>
-      {demographicData.map((item, index) => (
+      {demographicData.map((item) => (
```
fzhao99, line 35: This was getting caught in the linter as an unused var so I went ahead and fixed it

containers/tefca-viewer/src/styles/styles.scss
```diff
@@ -5,5 +5,3 @@
 @forward "layout";
 
 @forward "custom-styles";
-
-@import "~@trussworks/react-uswds/lib/index.css";
```
fzhao99, line 9: This import was breaking some of the accordion stylings, but we don't seem to need it for anything? Happy to add this back if folks can find a place where it's needed, but getting the stylings right on the customize query page was a pain in the neck with it included

containers/tefca-viewer/src/app/query/components/customizeQuery/CustomizeQueryAccordionBody.tsx
```diff
@@ -0,0 +1,64 @@
+import { Icon } from "@trussworks/react-uswds";
```
fzhao99, line 1: These are componetizations of UI code that are currently inlined in `CustomizeQuery`, but the actual work implementing them is upcoming in the next PR. Adding them here since their atomic implementations should be reviewable independently so I can make the next PR more reviewable, but lmk if we prefer to include these new files in the PR that they're actually getting called in

robertandremitchell (approved):
looks good. to confirm, this is just refactoring existing work, correct?

So we expect the next PR to:
* Fix accordion colors
* Separate accordions by valueset_name

fzhao99:
> looks good. to confirm, this is just refactoring existing work, correct?
> 
> So we expect the next PR to:
> 
> * Fix accordion colors
> * Separate accordions by valueset_name

Yes!

PR #2566, opened 2024-09-16
Execute custom queries

### LINKED ISSUE #2524: Execute custom query

bamader:
# Execute Custom Queries

## Summary
This PR adds the ability to execute queries that have been customized with the `CustomQuery` workflow. Base querying (i.e. for unaltered default valuesets of the use cases) is left unchanged. It also deletes the `test` page component of the query workflow completely, now that Marcelle's PR will be fusing test and demo.

## Related Issue
Fixes #2524


containers/tefca-viewer/src/app/format-service.tsx
```diff
@@ -257,3 +259,39 @@ export async function GetPhoneQueryFormats(phone: string) {
...
+ * @param useCase The base use case being queried for.
+ * @param vsItems The list of value set items the user wants included.
+ * @returns A structured specification of a query that can be executed.
+ */
+export const formatValueSetItemsAsQuerySpec = async (
+  useCase: string,
+  vsItems: ValueSetItem[],
+) => {
+  let secondEncounter: boolean = false;
+  if (["cancer", "chlamydia", "gonorrhea", "syphilis"].includes(useCase)) {
```
m-goggins, line 275: Just for my own understanding, why these 4 use cases and not the others?

containers/tefca-viewer/src/app/query-service.ts
```diff
@@ -174,13 +178,30 @@ export async function UseCaseQuery(
   return queryResponse;
 }
 
+/**
+ * Performs a generalized eCR-based query for collections of patients matching
```
m-goggins, line 182: eCR-based?

containers/tefca-viewer/src/app/query/test/page.tsx
m-goggins, line 1: I kinda think we should keep these page until the Connectathon is over, if only to avoid confusion because we already sent the link out to a variety of folks and it shouldn't be any different than the page formerly known as "demo".

containers/tefca-viewer/src/app/format-service.tsx
m-goggins: oh nvm, I see these are the only ones where we have to query for encounters twice. A future ticket/tech debt item could be pulling out these queries into `constants` or at least some other location where it's easier to identify and modify this list.

containers/tefca-viewer/src/app/format-service.tsx
```diff
@@ -257,3 +259,39 @@ export async function GetPhoneQueryFormats(phone: string) {
...
+  }
+  const labCodes: string[] = vsItems
+    .filter((vs) => vs.system === "http://loinc.org")
+    .map((vs) => vs.code);
+  const snomedCodes: string[] = vsItems
+    .filter((vs) => vs.system === "http://snomed.info/sct")
+    .map((vs) => vs.code);
+  const rxnormCodes: string[] = vsItems
+    .filter((vs) => vs.system === "http://www.nlm.nih.gov/research/umls/rxnorm")
+    .map((vs) => vs.code);
```
m-goggins, line 286: This set up is nice and allows us to expand to other systems, like ICD-10, and easily add to the QueryStruct when we sort out the classTypeCodes!

m-goggins (commented):
Looks good to me! I think we should consider leaving the `test` page until after the connectathon since that specific link was shared with folks last week, and then permanently remove it as part of tech debt clean up. 

The other thing I noticed, which I think has to do with Rob + Bob's PR not yet being fully merged is the headings on the customize query page do not reflect the Value Set names.

containers/tefca-viewer/src/app/format-service.tsx
bamader: Yeah I like that idea. These were just things that in our original demo use case formulation had follow up encounters we needed to query but pulling them into a constant would be reader-helpful for sure. Will add to my list of expanding tech debt items.

containers/tefca-viewer/src/app/query-service.ts
bamader: More like developer fugue based since I have no memory of writing this...

containers/tefca-viewer/src/app/query/test/page.tsx
bamader: That's fair. I can put it back in, but since we don't actually have any rendering for the test page components there will be no way to pass in the requisite `queryValueSets` and `setQueryValueSets` variables to the components defined here. We'll have to just make up vacuous unknowns for them. That _should_ be okay because I don't think we're directly rendering any of this stuff, but just flagging it.

fzhao99 (approved):
Thanks so much for your hard work on this Brandon and for fighting the tests!

PR #2621, opened 2024-09-24
configure local prettier to align with CI config

fzhao99:
# PULL REQUEST

Bob got really annoyed that there were formatter misalignments between CI prettier and local prettier, so he attempted to fix it. Think this one setting is the main misalignment, but also did [an exact install of prettier against the CI version like they suggest](https://prettier.io/docs/en/install.html#:~:text=This%20makes%20sure%20that%20everyone,other's%20changes%20back%20and%20forth.)

Now hitting save on our TSX files (🤞) shouldn't generate a bunch of comma-related changes that will just get reverted by the CI linter. Can test by running `npx prettier --write src/app` and notice that there aren't changes to any of the typescript files

### DIFF
--- a/containers/tefca-viewer/.prettierrc
+++ b/containers/tefca-viewer/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "trailingComma": "all"
+}
--- a/containers/tefca-viewer/package.json
+++ b/containers/tefca-viewer/package.json
@@ -73,7 +73,7 @@
     "jest-extended": "^4.0.2",
     "mocha": "^10.2.0",
     "postcss": "^8",
-    "prettier": "3.1.1",
+    "prettier": "4.0.0-alpha.8",
     "sass": "^1.69.5",
     "supertest": "^7.0.0",
     "tailwindcss": "^3.3.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -941,7 +941,7 @@
         "jest-extended": "^4.0.2",
         "mocha": "^10.2.0",
         "postcss": "^8",
-        "prettier": "3.1.1",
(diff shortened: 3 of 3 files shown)


robertandremitchell (commented):
```bash
rob@Roberts-MacBook-Pro tefca-viewer % git status
On branch bob/test-lint
Your branch is up to date with 'origin/bob/test-lint'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/app/api/query/parsing-service.ts
	modified:   src/app/assets/TEFCA_Query_Connector_API.postman_collection.json
	modified:   src/app/tests/assets/BundleClinicalInfo.json
	modified:   src/app/tests/assets/BundleEcrMetadata.json
	modified:   src/app/tests/assets/BundleHAPIServer.json
	modified:   src/app/tests/assets/BundleLabInfo.json
	modified:   src/app/tests/assets/BundleMiscNotes.json
	modified:   src/app/tests/assets/BundlePatient.json
	modified:   src/app/tests/assets/BundleSexualOrientation.json
	modified:   src/app/tests/assets/BundleTravelHistory.json
```

this generated a few additional files when I ran the the `rob@Roberts-MacBook-Pro tefca-viewer % npx prettier --write src/app` command. nbd, but wanted to flag

fzhao99:
Thanks for testing! Realized I assumed that those files getting caught was fine since I assumed they were prettierignored, but realizing now that's probs not a good assumption.

Zooming out here, wondering if folks think it's worth spending more time / merging this in the first place? Think it's fine for me just to leave as a local config if the papercut isn't annoying enough / if we think would be too high.

PR #2641, opened 2024-09-30
Enable Pool Client SSL Configuration for tefca

robertandremitchell:
# PULL REQUEST

## Summary
Does 2 things:
1. Enables SSL to remove it as a potential confound for AWS connection per @alismx recommendation
2. Restores `tefca.env` and amends `npm run dev` to work correctly for local development

## Related Issue
Fixes SSL issue that appears to be impact 

## Acceptance Criteria

## Additional Information
I would test both docker compose up and npm run dev to confirm both are still able to connect.

## Checklist


containers/tefca-viewer/docker-compose.yaml
```diff
@@ -35,7 +35,7 @@ services:
       - "3000:3000"
     environment:
       - NODE_ENV=production
-      - DATABASE_URL=postgres://postgres:pw@db:5432/tefca_db
```
DanPaseltiner, line 38: No change should be needed here since this only matters when we run on our locals and this is working just fine

containers/tefca-viewer/src/app/database-service.ts
```diff
@@ -1,6 +1,6 @@
 "use server";
 import { Pool, PoolConfig, QueryResultRow } from "pg";
-// import dotenv from "dotenv";
```
DanPaseltiner, line 3: I don't believe we should be using dotenv here as we aren't going to supply a dotenv file for any deployment beyond running locally.

containers/tefca-viewer/src/app/database-service.ts
```diff
@@ -14,14 +14,18 @@ select q.query_name, q.id, qtv.valueset_id, vs.name as valueset_name, vs.author
 `;
 
 // Load environment variables from tefca.env and establish a Pool configuration
-// dotenv.config({ path: "tefca.env" });
+console.log("Database Configuration:");
+console.log("Connection String:", process.env.DATABASE_URL);
+dotenv.config({ path: "tefca.env" });
 const dbConfig: PoolConfig = {
   connectionString: process.env.DATABASE_URL,
+  ssl: false, // Disable SSL explicitly
```
DanPaseltiner, line 20: Disable SSL seems highly suspect to me.

containers/tefca-viewer/package.json
```diff
@@ -4,13 +4,13 @@
   "version": "1.0.1",
   "private": true,
   "scripts": {
-    "dev": "docker-compose -f docker-compose-dev.yaml up && next dev",
+    "dev": "docker-compose -f docker-compose-dev.yaml up -d && npx next dev",
```
DanPaseltiner, line 7: I don't believe `npx` is required here. `...&& next dev` is sufficient on my machine

containers/tefca-viewer/src/app/database-service.ts
```diff
@@ -14,14 +14,18 @@ select q.query_name, q.id, qtv.valueset_id, vs.name as valueset_name, vs.author
 `;
 
 // Load environment variables from tefca.env and establish a Pool configuration
-// dotenv.config({ path: "tefca.env" });
+console.log("Database Configuration:");
+console.log("Connection String:", process.env.DATABASE_URL);
```
fzhao99, line 18: would there be any world where we'd log out secrets here?

fzhao99 (approved):
left one q but otherwise lgtm! local dev spins up as expected

containers/tefca-viewer/src/app/database-service.ts
robertandremitchell: oops, no, that's a leftover from when I was testing what was happening with local. removing now!

DanPaseltiner (changes requested):
Getting the `dev` script running again is a good thing, but I'm concerned about taking steps within the app to disable SSL. This hasn't been and issue previously and creates a security issue for us.

containers/tefca-viewer/src/app/database-service.ts
robertandremitchell: this was added back to support running locally. the dev compose file doesn't have a URL, so we need some way of supplying it in the event that it's not in a docker file, so I just restored the dotenv that had been used previously.

containers/tefca-viewer/package.json
robertandremitchell: weird, I had needed it at one point, but yeah, seems fine.

containers/tefca-viewer/src/app/database-service.ts
robertandremitchell: swapped to enable, which seems to be working fine locally.

containers/tefca-viewer/src/app/database-service.ts
robertandremitchell: I think I'm okay getting rid of it if we want to test this solution. We will just have to backburner a better way to get local development working, because it was a real headache not having working local development, but we can remove tefca.env again to get AWS working..

DanPaseltiner:
@robertandremitchell why have we checked in an entire Postgres DB? I could very well be missing something here.

robertandremitchell:
> @robertandremitchell why have we checked in an entire Postgres DB? I could very well be missing something here.

I've been having trouble getting the playwright tests to pass, so was just trying to chase down that error.

I went from:
```
[WebServer] db-1            | postgres: could not access the server configuration file "/var/lib/postgresql/data/postgresql.conf": No such file or directory
```

to:

```
[WebServer] db-1            | 2024-10-01 14:50:01.851 UTC [11] FATAL:  data directory "/var/lib/postgresql/data" has wrong ownership
db-1            | 2024-10-01 14:50:01.851 UTC [11] HINT:  The server must be started by the user that owns the data directory.
```

with this change, so hoping that once I figure that out, I can then remove pgdata/ or limit it to just pgdata.conf.

PR #2652, opened 2024-10-01
Add Flyway to tefca-viewer container

### LINKED ISSUE #2627: Move Flyway to run within tefca-viewer app
Flyway does not need to spin up its own service, as it seems to be creating issues for (or at least complicating) deployment. Instead, we should run flyway from within the `tefca-viewer` app by modifying the docker-compose file.

Acceptance Criteria:

- [x] The Dockerfile and dock-compose files are modified to run flyway at startup from within the tefca-viewer service
- [x] Flyway does not spin up its own service
- [x] Flyway migrations run successfully upon start up
- [x] Flyway-specific files, e.g., DockerFile.flyway, are removed from the codebase

m-goggins:
# PULL REQUEST

## Summary
This PR folds Flyway into the TEFCA Viewer app, removing Flyway as a separate service. It also re-works how we run Flyway tests locally by running the docker compose and build commands outside of playwright (previously these commands were coupled with the webServer start up) and how we run the tests in the TEFCA Viewer github workflow (breaks out into separate steps). You can run playwright tests locally using `npm run test:playwright`. There is also a fair amount of clean up to remove Flyway-specific files. 

Note: When you run `npm run dev`, the flyway service should still spin up separately so that migrations can run on the test db while also running the app using `next dev` to allow for hot reloads. 

## Related Issue
Fixes #2627 

## Acceptance Criteria

## Additional Information

## Checklist


containers/tefca-viewer/package.json
```diff
@@ -4,9 +4,9 @@
   "version": "1.0.1",
   "private": true,
   "scripts": {
-    "dev": "docker-compose -f docker-compose-dev.yaml up && next dev",
-    "dev-win": "start docker-compose -f docker-compose-dev.yaml up && next dev",
-    "dev:db": "docker-compose -f docker-compose-dev.yaml up",
+    "dev": "docker compose -f docker-compose-dev.yaml up && next dev",
```
fzhao99, line 7: Two things here:
- This is in a couple of different PR's that somehow all haven't made their way to main but I think we need this to be `docker-compose-dev.yaml up -d` so that the db/flyway containers run in the background / don't block the node thread from spinning up `next dev`. Otherwise, next won't ever start.
- Once adding that flag, the app spins up fine, it seems like the value of `process.env.DATABASE_URL` that the db service is expecting isn't getting set anywhere. Sticking a debug log ahead of the config object says that the env value is undefined

m-goggins: Thanks, @fzhao99! I add the `-d` to the `next dev` command. I'm on windows so not able to test locally in the same way, but dev-win is working for me.

For the DATABASE_URL issue, having a `.env` file with the DATABASE_URL set should take care of the issue. I added a note about it in the docker-compose-dev.yaml file as well. Let me know if you run into any issues!

DanPaseltiner:
@m-goggins I'm looking at the last successful test run and it doesn't [appear](https://github.com/CDCgov/phdi/actions/runs/11217494218/job/31179186097) that the e2e tests actually ran.

DanPaseltiner (approved):
LGTM. Whenever we can get all of the tests passing and feel confident that the tests are truly run consistently and reliably in the GH runners I think this is good to merge. Thank you for your work here @m-goggins this turned into a bit of beast to get sorted.

PR #2655, opened 2024-10-02
refactor select patient

### LINKED ISSUE #2585: Build out SelectPatient page
This can likely be built off existing work on MultiplePatients /No Patients. Rather than only direct to that page with 2+ patients, we could direct to this page and have the select patient functionality.

https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=3576-41989&node-type=instance&m=dev

fzhao99:
# PULL REQUEST

## Summary
Builds out the patient select page by

-  Renaming "multiple patients" to "patient search results"
-  Folding in the "no patients" screen into this view
-  Moving state setting around so that children components are only responsible for state as is scoped to what they're displaying
- A couple misc spacing / styling tweaks

## Related Issue
Fixes #2585 

## Acceptance Criteria
This can likely be built off existing work on MultiplePatients /No Patients. Rather than only direct to that page with 2+ patients, we could direct to this page and have the select patient functionality.

https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=3576-41989&node-type=instance&m=dev

## Additional Information
We haven't fixed the redirect yet since doing so would break query selection, so when a single result is returned, things still get redirected to the results view. You can get to the patient select page that this PR implements by 

1. Selecting the Chlamydia use case
2. Deleting the first name that autofills and leaving the last name "JMC"
3. Deleting the medical record number and the birth date
4. Hitting search for patient

https://github.com/user-attachments/assets/2dd6df81-6a14-4957-a192-8a2c848cc759

You can similarly get to the "No records found" empty state by putting some gibberish in the first name field
https://github.com/user-attachments/assets/719e6aee-c29a-4493-afae-da907c2a2881


containers/tefca-viewer/src/app/query/components/searchForm/SearchForm.tsx
```diff
@@ -126,12 +126,11 @@ const SearchForm: React.FC<SearchFormProps> = ({
     setOriginalRequest(originalRequest);
     const queryResponse = await UseCaseQuery(originalRequest, queryValueSets);
     setUseCaseQueryResponse(queryResponse);
-    if (!queryResponse.Patient || queryResponse.Patient.length === 0) {
-      setMode("no-patients");
-    } else if (queryResponse.Patient.length === 1) {
+
+    if (queryResponse.Patient && queryResponse.Patient.length === 1) {
```
fzhao99, line 130: will eventually delete this check in the final wiring PR

containers/tefca-viewer/e2e/query_workflow.spec.ts
```diff
@@ -283,7 +286,7 @@ test.describe("Test the user journey of a 'tester'", () => {
 
     await page.getByRole("link", { name: "Return to search results" }).click();
     await expect(
-      page.getByRole("heading", { name: "Multiple Records Found" }),
+      page.getByRole("heading", { name: "Search for a Patient", exact: true }),
```
fzhao99, line 289: This test reflects a slight navigational change where the backlink will now go back to the search field rather than the results view. Reasoning here is 

1) based on the way we're centralizing state in the parent view, maintaining a "history" of the query responses needed to get this to work correctly gets super complicated. The correct way to solve this would be to rewrite our SPA string-based routing system with something like [next-router](https://nextjs.org/docs/pages/api-reference/functions/use-router) instead and then integrating some sort of state history setup, which isn't something I wanted to do without discussion.

2) based on Michelle's comment from this morning, this link is going to change to query selection anyway

If folks feel that this goBack link not going back to the patient search results view is too major of a regression / should block this PR, happy to chat further.

containers/tefca-viewer/src/app/query/components/ResultsView.tsx
```diff
@@ -61,11 +56,6 @@ const ResultsView: React.FC<ResultsViewProps> = ({
     <>
       <div className="results-banner">
         <div className={`${styles.resultsBannerContent}`}>
-          <Backlink
```
fzhao99, line 64: I elected in the sort term to pull the backlink to the patient search results here because 

1) based on the way we're centralizing state in the parent view, maintaining a "history" of the query responses needed to get this to work correctly gets super complicated. The correct way to solve this would be to rewrite our hand rolled string-based routing system with something like [next-router](https://nextjs.org/docs/pages/api-reference/functions/use-router) instead and then integrating some sort of state history setup, which isn't something I wanted to do without discussion.

2) based on [Michelle's comment](https://skylight-hq.slack.com/archives/C07PQ078KK2/p1727882161133179) from this morning, this link is going to change to query selection anyway

If folks feel that removing this goBack link  is too major of a regression / should block this PR, happy to chat further.

m-goggins: I think it's okay to remove for now as long as there is a ticket in the backlog to address this. At the connectathon, folks were very interested to see what happens when we return multiple patients so having this easily accessible is a nice feature.

DanPaseltiner (approved):
LGTM

robertandremitchell (changes requested):
two minor notes:

* it looks like in figma that we are switching from `View Record` to `Select patient`.
* It also looks like we are adding a blue hover highlight here? https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=3576-42184&node-type=instance&m=dev

mikang:
> two minor notes:
> 
> * it looks like in figma that we are switching from `View Record` to `Select patient`.
> * It also looks like we are adding a blue hover highlight here? https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=3576-42184&node-type=instance&m=dev

- Where are switching from "View Record" to "Select Patient"?
- Blue highlight looks good!

fzhao99:
> > two minor notes:
> > 
> > * it looks like in figma that we are switching from `View Record` to `Select patient`.
> > * It also looks like we are adding a blue hover highlight here? https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=3576-42184&node-type=instance&m=dev
> 
> * Where are switching from "View Record" to "Select Patient"?
> * Blue highlight looks good!

I think in the actions column on the "select a patient screen" 
Screenshot 2024-10-03 at 10 15 27 AM

fzhao99:
@robertandremitchell  ready for re-review!

https://github.com/user-attachments/assets/173a53d3-9505-45b0-9f6e-bd8a33869786

containers/tefca-viewer/src/app/query/components/ResultsView.tsx
fzhao99: Ahh: flagging a potential source of misalignment between [the flow Michelle put together](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/VIPER-Viewer-(Name-TBD)?node-id=4-6068&t=723PFItWMBWFEwQe-1) and what (I think?) @m-goggins just raised. As is, there's no "return to search results" link in the above flow, but it sounds like that's something that users requested at the connectathon. Is this something we want to add / update in the designs?

cc: @DanPaseltiner @mikang

DanPaseltiner: Yes that feature has come up a lot. I think it is something we should implement, but for sanity and keeping PRs small my vote is for addressing in a separate ticket and associated PR.

DanPaseltiner:
> two minor notes:
> 
> * it looks like in figma that we are switching from `View Record` to `Select patient`.
> * It also looks like we are adding a blue hover highlight here? https://www.figma.com/design/YU5hAHY1Eg7J6gtNbBxm7d/DIBBs-Site-Experiences?node-id=3576-42184&node-type=instance&m=dev


Nice catches @robertandremitchell! 🚀

m-goggins (approved):
Realized that I forgot to approve yesterday and just commented by accident 🤦‍♀️

DanPaseltiner:
And ditto on the highlighting looking nice

robertandremitchell (approved):

mikang:
Looks great, thank you! 

Can we make the entire row clickable? Keep the link as an affordance, but make the entire row clickable to give a larger click target @fzhao99

fzhao99:
> Looks great, thank you!
> 
> Can we make the entire row clickable? Keep the link as an affordance, but make the entire row clickable to give a larger click target @fzhao99

@mikang sure thing! I'll merge this and add that fix right after

PR #2568, opened 2024-09-17
implement multiple accordions on customize query

fzhao99:
# PULL REQUEST

## Summary
Implements the frontend for the customize query multiple accordion selection. Follow on from #2564 and takes care of #2521

### Inclusion switches work (stuck an extra console log into the handle apply listener)
https://github.com/user-attachments/assets/04724a60-683d-4869-8ada-654234fb7e9f

### Redirect back to main page
https://github.com/user-attachments/assets/ae703257-6445-4ad2-87bf-c33b80808ce1

## Related Issue
Fixes #2523 

## Acceptance Criteria
Once we have a back end function that divides up the returned DB rows by their coding scheme, we want to display this information correctly to the user on the front end. We'll pass in all the relevant value set data to the CustomizeQuery component, but on each tab of labs, medications, and conditions, we want to show one accordion per coding scheme present in the data.

## Additional Information
There's a minor UI issue where the accordions don't persist their collapsed state when the tab switches because I couldn't get the out-of-box `handleToggle` function from Truss to work correctly. Didn't want that to block this work, so made a [follow up ticket](https://github.com/CDCgov/phdi/issues/2569) for it.

Additionally, did want to ask the team what their thoughts are about some of the transformation functions in `customizeQueryUtils.ts`, since ideally, I'd be writing some unit tests with some known starting / end-state test data. Wondering what our larger state of frontend testing looks like and wrote up a couple questions into a ticket [here](https://github.com/CDCgov/phdi/issues/2562). Happy to close this if this is duplicated anywhere.

PR #2665, opened 2024-10-03
cleanup frontend console errors

fzhao99:
# PULL REQUEST
A bunch of frontend console error cleanup as promised in [this PR](https://github.com/CDCgov/phdi/pull/2568#pullrequestreview-2310182175).

## Summary
- Adding key props where missing to solve the [react key issue](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key)
     - Discovered during testing that the chlamydia use case sample patient has a duplicate value in the observation data set. [Added a ticket to fix](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/2666)
     Screenshot 2024-10-03 at 2 40 31 PM
- Adding alt labels to a few icons / images to solve a slew of accessibility-related warnings
- Had an annoying issue with the data disclaimer modal that was ultimately solved by upgrading Truss to the latest version. In debugging, went ahead and componentized the Modal component into something that hopefully will be more ergnomic and updated related react types and USWDS versions
- Had to rework some semantic table elements to use the native `tr`, `td` values. Had to tweak some stylings as a result to match designs.


containers/tefca-viewer/src/app/page.tsx
```diff
@@ -39,6 +39,7 @@ export default function LandingPage() {
               src="/tefca-viewer/tefca-graphic.svg"
               width={250}
               height={300}
+              priority
```
fzhao99, line 42: https://nextjs.org/docs/pages/api-reference/components/image#priority

containers/tefca-viewer/next.config.js
```diff
@@ -4,14 +4,7 @@ const path = require("path");
...
     includePaths: [
-      path.join(
-        __dirname,
-        "../..",
-        "node_modules",
-        "@uswds",
-        "uswds",
-        "packages",
-      ),
+      path.join(__dirname, "./", "node_modules", "@uswds", "uswds", "packages"),
```
fzhao99, line 7: https://skylight-hq.slack.com/archives/C06SXT5PXSQ/p1727987233941229

Worth chatting about in cross-eng sync what the value of turbo repo is? Could see a world where we share out the frontend components, but esp if we're moving to a separate repo soon this might just be all headache without benefit

robertandremitchell (approved):
looks good to me

containers/tefca-viewer/next.config.js
DanPaseltiner: The history with the turbo repo is that for a while we had wanted a way to ensure consistency between the eCR Viewer and the Query Connector. However, we have yet to actually use any shared components and the two products are quickly diverging. I think it is worth removing turbo at this point given that we aren't getting any value from it currently and it is adding a lot of unnecessary complexity.

PR #2672, opened 2024-10-04
Add optional chaining for facility contact address

angelathe:
# PULL REQUEST

## Summary
Add optional chaining to Encounter Info Facility Contact Address info, so that the eCR Viewer doesn't crash when we load an eCR.

Example: http://localhost:3000/view-data?id=016accab-d6bf-11ec-81e1-8465ad080d41#eicr-custodian-details

## Related Issue
N/A - Fixes bug that happened during live demo :(

## Checklist


containers/ecr-viewer/src/app/services/evaluateFhirDataService.ts
```diff
@@ -373,7 +373,7 @@ export const evaluateFacilityData = (
     referenceString = facilityContactAddressRef[0].reference;
   }
   const facilityContactAddress = referenceString
-    ? evaluateReference(fhirBundle, mappings, referenceString).address[0]
+    ? evaluateReference(fhirBundle, mappings, referenceString)?.address?.[0]
     : "";
```
BobanL, line 377: Can you also change this to be `undefined` rather than `""` so the this variable type isn't as confusing?

BobanL (approved):
Great find! Just one tiny nit.

containers/ecr-viewer/src/app/services/evaluateFhirDataService.ts
angelathe: Done!

PR #2738, opened 2024-10-10
Add function to retrieve eRSD data

### LINKED ISSUE #2736: write function to get eRSD from APHL
Write a function that makes a GET request to APHL to download the eRSD and JSON FHIR bundle.

GIVEN this function 
WHEN it is executed
THEN the eRSD as a JSON FHIR bundle is returned.

This function should 

1. be asynchronous 
2. use the classes from the `fhir` library
3. live in `database-service`

m-goggins:
# PULL REQUEST

## Summary
This PR adds the asynchronous `getERSD` function to `database-service.ts` that, given an API key, will return the specified version of the eRSD from the APHL API. I updated the README to include directions on obtaining the free API key and changed the name from TEFCA Viewer to Query Connector where possible. 

We don't currently have a great way to test this function in our testing framework, but I added the following code block to query/page.tsx, set up my API key in a local `.env` file, and was able to get a successful response using both `npm run dev-win` and `docker-compose`, as well as an Operation Outcome if I provided a bogus key.

```
useEffect(() => {
    console.log("Fetching eRSD data... here");
    const fetchErsdData = async () => {
      console.log("Fetching eRSD data...");
      const result = await getERSD();
      console.log("Received result:", result);
    };

    // Call the function here
    fetchErsdData().catch((error) =>
      console.error("Error fetching eRSD data:", error)
    );
  }, []);
```

Note for @shanice-skylight: When we deploy this to jurisdictions, they will need to sign up for their own API key and make the key available wherever it is deployed.

## Related Issue
Fixes #2736 

## Acceptance Criteria
GIVEN this function
WHEN it is executed
THEN the eRSD as a JSON FHIR bundle is returned.

## Additional Information
We pull v2 of the eRSD by default (like we did for the TCR) but v3 of the eRSD is now available. We'll have the ability to test out how to update the data when we get to that point in future development 🙂 

## Checklist


containers/tefca-viewer/src/app/database-service.ts
```diff
@@ -87,3 +87,51 @@ export const mapQueryRowsToValueSetItems = async (rows: QueryResultRow[]) => {
...
+          code: "exception",
+          diagnostics:
+            error instanceof Error
+              ? error.message
+              : "Error retrieving eRSD data. Please confirm that your eRSD API key is valid.",
+        },
+      ],
+    } as OperationOutcome;
+  }
+}
```
DanPaseltiner, line 137: Wouldn't we expect to get a 401 or 403 from APHL if the API key is no good? In this case why do we need this second error handling mechanism. Maybe I'm missing something.

m-goggins: Oh yup, you're totally right! I removed the try/catch

DanPaseltiner:
Given that the scope of this function is so small and limited to just calling and external API using the well tested `fetch` library I'm okay skipping a test here. We could do some mocking to avoid coupling the tests to the APHL API, but I'm not sure it would be terribly meaningful.

DanPaseltiner (approved):
We may need to revisit how we are handling eRSD versions, but I think what we are doing now is okay to start

robertandremitchell (approved):

PR #2697, opened 2024-10-08
Add Date of Death, Extended Race, Extended Ethnicity 

lina-roth:
# PULL REQUEST

### Date of Death 
Death Date Ecr - 1.2.840.114350.1.13.363.2.7.8.688883.861749262
Screen Shot 2024-10-07 at 10 07 56 PM

### Additional Race 
Additional Race Ecr - daf7f4ae-bff3-11ec-81e1-8465ad080d41
Screen Shot 2024-10-07 at 10 45 14 PM

### Additional Ethnicity 
Additional Ethnicity Ecr - e2c69f15-aaea-11ec-81e0-9e7eef92c510
Screen Shot 2024-10-07 at 10 46 29 PM

### Age calculated from date of death if exists 
Screen Shot 2024-10-09 at 6 10 36 PM

### Age calculated from current date if DOD doesnt exist 
Screen Shot 2024-10-09 at 6 10 14 PM

## Summary
What changes are being proposed?
- Add DOD to fhir converter 
- Update Race to have a new line 
- Add Age at Death 
- Add additional ethnicity and split with a new line 

## Related Issue
Fixes #2518 

## Acceptance Criteria
- [ ] If the patient is deceased display a "Date of Death" field
- [ ]  Ensure we're FHIR converting "Additional Race" and display it on a new line in the Race field if it exists
- [ ] Ensure we're FHIR converting "Detailed Ethnicity" and display it on a new line in the Ethnicity field if it exists
- [ ]  If Date of Death exists, Current Age should be undefined and instead show Age at Death (the reverse is also true)
- [ ] update tests

## Additional Information
- Removing     console.log(releaseVersion);

## Checklist

- [ ] If this code affects the other scrum team, have they been notified? (In Slack, as reviewers, etc.)

[//]: # (PR title: Remember to name your PR descriptively!)

PR #2786, opened 2024-10-17
Add date of death, age at death, detailed ethnicity, and detailed race

lina-roth:
# PULL REQUEST

## Summary
New PR because other one was before git history rewrite and could not reconcile changes 

https://github.com/CDCgov/phdi/pull/2697


BobanL (approved):
👍

PR #2806, opened 2024-10-23
fix seed db config to generate the bundle.json

BobanL:
# PULL REQUEST

## Summary
- Add outputs in the seed-ecr-viewer-config to ensure that bundle would be returned
- Remove unused print statement


JNygaard-Skylight (approved):
LGTM

PR #2833, opened 2024-10-30
build: use release of fhir converter instead of main

mcmcgrath13:
# PULL REQUEST

## Summary
Instead of following along on `main` of the `skylight-hq` fork of the fhir converter, start cutting releases on that fork and point our docker file to the releases.

I need to make an update to the fhir converter that will need coordination on this repo to not break things, so starting to version allows this to be managed safely.


containers/fhir-converter/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM mcr.microsoft.com/dotnet/sdk:6.0 as build
 
 # Download FHIR-Converter
-RUN git clone https://github.com/skylight-hq/FHIR-Converter.git --branch main --single-branch /build/FHIR-Converter
+RUN git clone https://github.com/skylight-hq/FHIR-Converter.git --branch v7.0-skylight-1 --single-branch /build/FHIR-Converter
 COPY /CustomFhir/CustomFilters.cs /build/FHIR-Converter/src/Microsoft.Health.Fhir.Liquid.Converter/Filters/
```
BobanL, line 5: ooo - does this mean we could just move all of our custom filters into that fhir converter repo rather than maintain them here?

BobanL (approved):
🎉

containers/fhir-converter/Dockerfile
mcmcgrath13: I don't see why not. It would change the dev flow somewhat, but probably make it somewhat easier/more straightforward in the long run. @austin-hall-skylight what do you think?

mcmcgrath13: implications I can think of:
- more cross repo coordination/release cutting
- functionally giving up on ever getting off the fork (though it sounds like maybe we'd already done that? cc @emmastephenson )
- easier dev setup (currently getting the fhir repo working is super clugey)
- easier to mess up picking up changes in fhir repo

mcmcgrath13: to recap IRL discussion, we'll move all customizations into this fork and rely on this fork for development and release for phdi repo (pending @austin-hall-skylight making sure this won't break anyone)

PR #2877, opened 2024-11-08
Update getDataId to use ID attribute instead of data-id comment

### LINKED ISSUE #2871: BUG: Lab results missing information
## Description

Lab results are no longer pulling data from the html table so some fields are empty

Potential problem:
- The ID attributes of HTML elements in the tables are no longer being converted into comments with data-id tags so the values cannot be looked up.

Potential solution:
- Figure out how to get the observation divs processed correctly again..

Relevant files/functions: 
    - `Templates/eCR/Resource/_Observation.liquid` lines 136-153
    - `CustomFhir/CustomFIlters.cs`

## Steps to reproduce

Example eCRs:
http://localhost:3000/view-data?id=1.2.840.114350.1.13.297.3.7.8.688883.567479

## Priority

#### Priority
- [x] **High priority** _High priority bugs must be addressed immediately. They often affect the app’s functionality and impact customers and their user experience. They must take priority._

austin-hall-skylight:
# PULL REQUEST

## Summary
Because the xhtml tables in Compositions are now being saved as their raw string representation rather than being converted to html, we are no longer adding element ID's as data-id in comment tags. This updates the getDataId function to look at an element's ID rather than looking for comment tags with data-id in them.

## Related Issue
Fixes #2871 

## Checklist


containers/ecr-viewer/src/app/services/formatService.tsx
```diff
@@ -415,25 +415,18 @@ export function getFirstNonCommentChild(
...
-      const match = commentText.match(/data-id:\s*([^\s]+)/);
-      if (match) {
-        return match[1];
-      }
-    }
+  if (elem.getAttribute("data-id")) {
+    return elem.getAttribute("data-id");
+  } else if (elem.id) {
+    return elem.id;
+  } else {
```
mcmcgrath13, line 427: should we still look for the comment data-id in case some other place is relying on it that we don't know?

containers/ecr-viewer/src/app/tests/services/formatService.test.tsx
```diff
@@ -567,56 +567,29 @@ describe("getDataId", () => {
     expect(result).toEqual("attribute123");
   });
 
-  it("should return the correct data-id from a comment node", () => {
+  it("should return the id if there is one", () => {
```
mcmcgrath13, line 570: could you add a test that takes something like `<tag ID="hi" />` and parses and sanitizes it and then gets the data id and checks it's `"hi"`?

containers/ecr-viewer/src/app/services/formatService.tsx
austin-hall-skylight: So the only parts of the code that call formatTablesToJSON (which is the only function that calls getDataId) are related to building plan of treatment and lab results, neither of which should be getting data-id comments since they don't get passed to ToHtmlString/ToHtmlStringJoinBr anymore. I'm happy to pair on it if you want to look into it some more.

mcmcgrath13: works for me!

mcmcgrath13 (approved):
thanks!

PR #2910, opened 2024-11-15
ci: add data conversion github action

mcmcgrath13:
# PULL REQUEST

## Summary
Add seed data conversion as a CI job to make sure changes don't inadvertently affect its success.

There are currently a few failing conversions 😢. These are due to genuinely duplicate IDs in eCRs. We should figure out how we want to address those

Also simplified the running steps for conversion by using the `include` option.


BobanL (approved):
👏 👏 👏 👏

containers/ecr-viewer/seed-scripts/docker-compose-seed.yml
```diff
@@ -1,3 +1,5 @@
+include:
+  - ../docker-compose.yml
```
BobanL, line 2: Actually - I think this causes issues in the `npm run local-dev` script. After running `convert-seed-data`, the DB is is empty for `local-dev` because the DB volume is created for a different service. 

If we want to keep this pattern, I _think_ a new seed profile will need to be added this whole file and add that to `convert-seed-data`. `local-dev` and `clear-local` should point this docker-compose file.

BobanL (changes requested):
Thought of a change that needs to be made to the npm scripts!

containers/ecr-viewer/seed-scripts/docker-compose-seed.yml
mcmcgrath13: nice catch! I was able to fix this by having both composes work in the same `name` space, so the volume is named consistently during conversion and use

containers/ecr-viewer/README.md
```diff
@@ -71,8 +71,8 @@ The eCR Viewer is primailly deveoped on Mac silicon machines, See this [integrea
 
 Sample eICRs are included in `containers/ecr-viewer/seed-scripts/baseECR/`. If you ever need to update the eCRs or add new eCRs you can regenerate the data by:
 
-1. Run `npm run convert-seed-data` to re-run the FHIR conversion of the seed eCRs
-2. Delete the current volume used by your DB: `docker compose -f ./docker-compose.yml --profile "*" down -v`
+1. Delete the current volume used by your DB: `docker compose -f ./seed-scripts/docker-compose-seed.yml --profile "*" down -v`
```
mcmcgrath13, line 74:
```suggestion
1. Delete the current volume used by your DB: `docker compose -f ./docker-compose.yml --profile "*" down -v`
```

containers/ecr-viewer/seed-scripts/docker-compose-seed.yml
BobanL: Never knew about the `name` property! Nice find!

BobanL (approved):
👏 👏 👏 👏 👏  for real this time

PR #2928, opened 2024-11-18
build: use standalone executable to avoid dotnet runtime

mcmcgrath13:
# PULL REQUEST

## Summary
Instead of using a `dotnet` runtime, use a `python-slim` base. This is enabled by using the `publish` command instead of `build` in the build phase of the `fhir-converter-service` and asking for a standalone linux executable. 

Also:
* build only the tool instead of the whole repo.
* add cachebusting of only the python service part of the image so you can avoid a dotnet rebuild when not working on that part of the code

Depends on https://github.com/skylight-hq/FHIR-Converter/pull/12

## Additional Information

## Checklist


containers/fhir-converter/Dockerfile
```diff
@@ -1,39 +1,27 @@
 FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
 
 # Download FHIR-Converter
-RUN git clone https://github.com/skylight-hq/FHIR-Converter.git --branch v7.0-skylight-9 --single-branch /build/FHIR-Converter
+RUN git clone https://github.com/skylight-hq/FHIR-Converter.git --branch mcm/publish-standalone --single-branch /build/FHIR-Converter
```
mcmcgrath13, line 4: TODO: update with released version before merging

BobanL (approved):
Nice fix!

Issue #2751, opened 2024-10-10
FILTER: Frontend condition filtering


emmastephenson:
## User Story

As a user of the eCR Library, I want to filter by condition, so that I can only see eCRs associated with that condition.

Examples: I want to filter by only high-priority conditions, like Anthrax, so that I can quickly see if any eCRs need followup.

## Acceptance Criteria

- [x] Condition filter is implemented according to designs
- [ ] Reset button appears when filtering is applied
- [x] Condition list is populated by a custom backend query (see #2749)
- [x] When user checks boxes for conditions and searches, only eCRs with those conditions in the RR are shown 

## Scenarios

_Scenario 1_
- I click on the "Reportable Condition" button
- A dropdown appears with all reportable conditions for my eCRs listed
- I click on 2 conditions
- I click "Apply Filter"
- Only eCRs with those reportable conditions are shown

_Scenario 2_
- I click on the "Reportable Condition" button
- From the dropdown, I choose "select all"
- When I filter, all eCRs are still in the Library

## Design Notes

[Link to Figma design in Dev mode](https://www.figma.com/design/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-Dec-2024)?node-id=4804-35705&node-type=instance&m=dev)
[Link to Figma prototype](https://www.figma.com/proto/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-Dec-2024)?node-id=4804-35695&node-type=instance&m=dev&scaling=min-zoom&content-scaling=fixed&page-id=4032%3A9311&starting-point-node-id=4806%3A39674)

_Screenshots below should not be used for working on this ticket - always refer to Figma design for most up-to-date designs:_
<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN004Qnc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--516ae9aa8dab1bbe8db303b42204b4d43ac7cd7e/Screenshot%202024-10-10%20at%207.47.24%E2%80%AFPM.png" alt="Screenshot 2024-10-10 at 7.47.24 PM.png" />

## Additional context

This will use the backend tickets #2749 and #2750.

emmastephenson:
Open question for design: should "Select All" become "Deselect All" if any conditions are selected?

PR #2938, opened 2024-11-20
Add backend filtering on conditions

### LINKED ISSUE #2750: FILTER: Backend condition filtering
## What needs to be done

Implement condition filtering on the backend. When given a list of conditions, this action can filter the returned eCRs for the Library to only those that match the condition.

## Why it needs to be done

This is the backend component needed to make filtering by condition functional. This feature is likely to be heavily used by our pilot partners using the non-integrated Viewer.

## Acceptance Criteria

- [ ] When passed a condition list, only eCRs with that condition in the RR are returned
- [ ] No other eCRs are visible
- [ ] If an eCR is multi-condition, any of those conditions can be filtered on

angelathe:
# PULL REQUEST

## Summary
- Modify SQL statement to include filtering on conditions
    - When both search and condition filtering are applied, the results should reflect the intersection of both filters.
    - If a user filters on condition "A", they should see all reports that reference condition "A" OR "A" and other conditions (and all other conditions should also be listed)
- Modify and add unit tests

## Related Issue
Fixes #2750 

## Acceptance Criteria
- [ ] When passed a condition list, only eCRs with that condition in the RR are returned
- [ ] No other eCRs are visible
- [ ] If an eCR is multi-condition, any of those conditions can be filtered on

## Additional Information
1. Right now, eCR library table only pulls from Postgres (I don't think we ended up actually implementing reading the ecr Library data from SQL Server?). Writing this ticket separately to pull data from SQL Server.
2. Placeholder for conditions to filter on has been added until #2751 is worked on (see below for testing).

## Checklist (for testing)
- [ ] To check that filtering on conditions works, look for the following code in `containers/ecr-viewer/src/app/page.tsx`
```
// Placeholder for given array of conditions to filter on, remove in #2751
  const filterConditions = undefined;
```
- [ ] Change `filterConditions` to the array of condition(s) you'd like to filter on, for example:
```
const filterConditions = ["Anthrax (disorder)", "Measles (disorder)"]
```
- [ ] Cross-reference eCRs listed in the library with the (number of) results you were expecting

Other things to check:
- [ ] Please also check that using search PLUS filtering on conditions results in the expected eCR rows. Example: searching "Cutla" plus adding filtering on "Anthrax (disorder)" should result in 5 rows of eCRs.
- [ ] Check that pagination values reflect an search and/or filtering criteria applied
- [ ] Check that all conditions for eCRs that match the filtering criteria are displayed, not just the condition(s) filtered on.


containers/ecr-viewer/src/app/api/services/listEcrDataService.ts
```diff
@@ -128,6 +152,32 @@ export const generateSearchStatement = (searchTerm?: string) => ({
...
+    }
+
+    const whereStatement = filterConditions
+      .map((condition) => {
+        return pgPromise.as.format("erc_sub.condition ILIKE $[condition]", {
+          condition: `%${condition}%`,
+        });
+      })
+      .join(" OR ");
+    const subQuery = `SELECT DISTINCT ed_sub.eICR_ID FROM ecr_data ed_sub LEFT JOIN ecr_rr_conditions erc_sub ON ed_sub.eICR_ID = erc_sub.eICR_ID WHERE erc_sub.condition IS NOT NULL AND (${whereStatement})`;
```
angelathe, line 176: Need subquery so that we filter on eCR IDs that meet the filter conditions criteria, not the ecr ID/condition pairings themselves. 

Ex. Desired behavior: If we search on condition "A", and eCR 1 has reportable conditions "A" and "B", we want eCR 1 to show up and display all its reportable conditions ("A" and "B"), not just the one filtered on ("A").

JNygaard-Skylight (dismissed):
~~Don't forget SQL Server.~~ I'm sorry bundles are never stored in SQL Server.

angelathe:
> Don't forget SQL Server.

^ Making a note that this is right. `listEcrData` only uses postgres, table also needs to be able to pull from SQL Server. Writing separate ticket for this now

BobanL (approved):

PR #2939, opened 2024-11-20
update evaluate cache to use a smaller key

### LINKED ISSUE #2679: eCRs loading times taking over 1 minute
## Description

Some eCRs take over 1 minute to load

## Steps to reproduce

_List out the steps in detail to reproduce the bug_

1. View eCR Maine 2: 1.2.840.114350.1.13.206.2.7.8.688883.470000397 http://localhost:3000/view-data?id=1.2.840.114350.1.13.206.2.7.8.688883.470000397

## Expected Result

Loading should take less than 10 - 15 seconds

## Actual result

Loading takes more than 1 minute.

## Priority

#### Priority
- [x] **Medium priority** _Typically, medium-priority bugs do not affect customers directly and therefore can be fixed in the normal course of testing and development._

## Notes

- Add a test to ensure this bug does not occur again.
- 

## Acceptance Criteria
* Bug is fixed OR a ticket is created to address the root cause.

BobanL:
# PULL REQUEST

## Summary
- update evaluate cache to a smaller key when available. 

## Related Issue
Fixes #2679 

## Additional Information


containers/ecr-viewer/src/app/view-data/utils/evaluate.ts
```diff
@@ -6,7 +6,7 @@ import {
   UserInvocationTable,
 } from "fhirpath";
 
-let cache: { [key: string]: any } = {};
+let evaluateCache: { [key: string]: any } = {};
```
mcmcgrath13, line 9: would it make sense to use a `Map` here instead of an object. Might be a bit more performant and doesn't come with any of the builtin key (deep) corners of an object

containers/ecr-viewer/src/app/view-data/utils/evaluate.ts
```diff
@@ -32,14 +32,29 @@ export const evaluate = (
   },
 ): any[] => {
   let key =
-    JSON.stringify(fhirData) +
+    (fhirData?.resourceType === "Bundle"
```
mcmcgrath13, line 35: wny just `bundle` for the bundle case? could you add a comment on this logic?

containers/fhir-converter/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
 
 # Download FHIR-Converter
-RUN git clone https://github.com/skylight-hq/FHIR-Converter.git --branch v7.0-skylight-11 --single-branch /build/FHIR-Converter
+RUN git clone https://github.com/skylight-hq/FHIR-Converter.git --branch boban/add-id-to-bundle --single-branch /build/FHIR-Converter
```
BobanL, line 4: todo - update this to latest

containers/ecr-viewer/package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "next lint",
     "test": "TZ=America/New_York jest",
     "test:watch": "TZ=America/New_York jest --watch",
-    "clear-local": "docker compose -f ./docker-compose.yml --profile \"*\" down -v",
+    "clear-local": "docker compose -f ./seed-scripts/docker-compose-seed.yml --profile \"*\" down -v",
```
mcmcgrath13, line 16: I would have expected that to be fine the old way given the namespacing and that's where the volume is declared - is that not the case?

containers/ecr-viewer/src/app/view-data/utils/evaluate.ts
```diff
@@ -6,7 +6,7 @@ import {
   UserInvocationTable,
 } from "fhirpath";
 
-let evaluateCache: { [key: string]: any } = {};
+let evaluateCache: Map<string, any> = new Map();
```
mcmcgrath13, line 9:
```suggestion
const evaluateCache: Map<string, any> = new Map();
```

containers/ecr-viewer/src/app/view-data/utils/evaluate.ts
```diff
@@ -31,30 +31,26 @@ export const evaluate = (
...
-    path,
-    context,
-    model,
-    options,
+  evaluateCache.set(
+    key,
+    fhirPathEvaluate(fhirData, path, context, model, options),
   );
-  return evaluateCache[key];
+  return evaluateCache.get(key);
```
mcmcgrath13, line 48: a really minor potential performance improvement would be to pull out the `fhirPathEvaluate(...)` to a variable assignment and then set, then return the variable to avoid two map interactions. But we're talking really minor, so feel free to ignore it

containers/ecr-viewer/package.json
BobanL: This worked fine for clearing the databases, but it wouldn't kill any of the services from the seed-scripts specifically.

containers/ecr-viewer/src/app/view-data/utils/evaluate.ts
BobanL: I think I'll stick with this approach for now..

containers/ecr-viewer/src/app/tests/components/Clinical.test.tsx
```diff
@@ -52,7 +52,7 @@ describe("Snapshot test for Procedures (Treatment Details)", () => {
         performedDateTime: "06/24/2022",
       },
       {
-        id: "b40f0081-4052-4971-3f3b-e3d9f5e1e44d",
+        id: "b40f0081-4052-4971-3f3b-e3d9f5e1e44e",
```
mcmcgrath13, line 55: why change the id here?

mcmcgrath13 (approved):
🚀

containers/ecr-viewer/src/app/tests/components/Clinical.test.tsx
BobanL: because this ID was duplicated in this example.

PR #2941, opened 2024-11-21
fix: abnormal flag display

### LINKED ISSUE #2715: BUG: Interpretation code does not always display in lab table "Value" column
## Description

For the eCR below, the interpretation code does not always display in the "Value" column of the lab component table. Please make sure that if the interpretation code exists, that it is displayed.

Relevant files/functions:
- `observationValue` path in `fhirPath.yml`
- `evaluateDiagnosticReportData` in `labsService.tsx`

## Steps to reproduce

Example eCR:
- 10e5b07b-5d3f-426f-a7e1-13ea909ff5a5: [Value interpretations (seen in the parens) does not always show up](https://skylight-hq.slack.com/files/U05J6B6RZFV/F07HPAJ4JCQ/8.png)

## Expected Result

If an interpretation code exists for a Value in the lab component table, it should be displayed.

## Actual result

Example from `10e5b07b-5d3f-426f-a7e1-13ea909ff5a5`:

## Priority

_Provide a estimate for level of impact the bug has today by placing an X in the applicable bracket_

#### Priority

## Notes

N/A

mcmcgrath13:
# PULL REQUEST

## Summary
Fix the display of interpretation codes so that they always display.

With the prior logic, due to the set nature of fhirpath, if the value and the interpretation were the same, then only the value would display along with empty parens.

## Related Issue
Fixes #2715
## Additional Information
image


containers/ecr-viewer/src/app/api/fhirPath.yml
```diff
@@ -126,7 +126,7 @@ specimenSource: "Observation.extension[0].extension.where(url = 'specimen source
 observationReferenceValue: "Observation.extension[0].extension.where(url = 'observation entry reference value').valueString"
 observationComponent: "code.coding.display.first()"
 # observationValue logic: <value> <units> (<interpretation>)
-observationValue: (valueQuantity.value.toString() | valueString | valueCodeableConcept.coding.display | iif(valueQuantity.unit.exists(), iif(valueQuantity.unit = '%', valueQuantity.unit, ' ' | valueQuantity.unit), '') | iif(interpretation.coding.display.exists(), ' (' | interpretation.coding.display | ')', '')).join('')
+observationValue: (valueQuantity.value.toString() | valueString | valueCodeableConcept.coding.display | iif(valueQuantity.unit.exists(), iif(valueQuantity.unit = '%', valueQuantity.unit, ' ' + valueQuantity.unit), '') | iif(interpretation.coding.display.exists(), ' (' + interpretation.coding.display + ')', '')).join('')
```
BobanL, line 129: instead of using the display, I think the ticket and the screenshot specifically called out using the coding.code instead. 
image

mcmcgrath13: It did use the word `code`, but I interpretted that more colloquially since the emphasis was on the missingness and other results had the long form and that wasn't called out in the ticket as an issue. @ashton-skylight which would you prefer? (I kind of prefer the short, but it does leave more room for interpretation)

image

ashton-skylight: @mcmcgrath13 From what you've seen, are the two values generally abnormal and high?

ashton-skylight: @mcmcgrath13 There are quite a few codesets. Wondering if user readability is easier with the display name?

mcmcgrath13: @ashton-skylight upon looking at the codes, I think it's probably better to be explicit/long here even though the common codes are kinda obvious. There are some in there that I would not know how to interpret without going to do the lookup

ashton-skylight: @mcmcgrath13 Agreed, it might be somewhat redundant, but would hopefully save time for users to not have to interpret codes outside of the common sightings

PR #2957, opened 2024-11-25
fix: use teal lines in ECR summary box

### LINKED ISSUE #2626: Dividing lines in eCR Summary should be teal, not gray
## Description

The horizontal dividing lines in the eCR Summary are currently gray instead of the teal in the designs.

## Steps to reproduce

1. open up an eCR in the eCR Viewer
2. observe the color of the lines in the eCR Summary

## Expected Result

Lines should be teal (#99DEEA)

## Actual result

Lines are gray

mcmcgrath13:
# PULL REQUEST

## Summary
Use teal lines in ECR summary box, per design. Achieved by adding a `themeColor` prop to `DataDisplay` and `DataTableDisplay` leaving a default of `"gray"`.

## Related Issue
Fixes #2626

## Additional Information
image


ashton-skylight:
Excellent work!

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -39,24 +39,30 @@ $main-min-width: $ecr-viewer-container-min-width + $ecr-viewer-gap + $nav-wrappe
...
   margin-left: .5rem;
   margin-right: .5rem;
 }
 
+.section__line {
+  border-top: 1px solid #99DEEA;
+  margin-top: .5rem;
+  margin-bottom: .5rem;
+}
+
```
BobanL, line 53: Would it be too far out of scope of this work to just remove this class and replace it with .section__line_teal? If you think it's too much right now that's fine.

BobanL (approved):
Looks good - just one question about if we could remove a duplicate class.

containers/ecr-viewer/src/styles/custom-styles.scss
mcmcgrath13: I think that's in scope - I'm all for clean up things while you're there (within reason)

mcmcgrath13: fixed! and a lot extra

BobanL (approved):
Looks great still!

Issue #2749, opened 2024-10-10
FILTER: Query for all unique conditions 


emmastephenson:
## What needs to be done

Write a backend query to find all unique conditions for all eCRs stored in the Viewer. 

## Why it needs to be done

This is one of the backend prerequisites to making filters work on the frontend (see ticket #2751). To populate the condition filter dropdown, we need to know all the conditions that a user could filter by.

## Acceptance Criteria

- [ ] Backend query available to gather all unique conditions across our eCR storage
    - [ ] Query works for core schema
    - [ ] Query works for extended schema
- [ ] Endpoint so that frontend can use this query should also be included

## Technical Notes

First in a series of sorting tickets - actually using this query will come later.

## Additional context

We thought about having a separate "conditions" table that would be the master list of all conditions available, but there are two issues with this approach:
1. It's another table for our partner STLTs to maintain
2. Each eCR store won't actually have the full contingent of potential conditions - it depends on what they've authored in RCKMS.

PR #2981, opened 2024-12-03
Filter by conditions in eCR Library (frontend)

### LINKED ISSUE #2751: FILTER: Frontend condition filtering

angelathe:
# PULL REQUEST

## Summary

**Main story:**
- Create `Filters` and `FilterReportableConditions` component

- Clicking the Filter button should open the Filter by Reportable Conditions combo box, and vice versa.
    - Combo box should be populated with the available conditions, populated by the `/conditions` endpoint query
- Scrollable section should be available for the condition checkboxes, while the `Select All` checkbox and `Apply filter` button should remain static.
- When a user filters for conditions and clicks `Apply filter`:
    - URL query should populate with the `=condition` param that reflects the conditions selected to filter on.
    - The eCRs displayed should update to reflect the new filters. They should have at least one of the reportable conditions selected to filter on.
    - The tag displaying number of conditions to filter on updates.
- When a user checks boxes for conditions but closes button before clicking 'Apply fliter', the filter conditions and the # conditions tag should reset to their most recent "Applied" state.
- Opening and closing the filter combo box when filters are applied should not change the conditions filtered on, and should maintain the state of which conditions are checkmarked.
- `Select all`/`Deselect all` - By default, all conditions are selected, the button displays "Deselect all," and the URL search query does not have the condition param. When no conditions are selected, the button changes to "Select all," and only eCRs with no conditions are displayed (though none should exist).
- Query persists over a page reload

**Misc:**
- CSS styling to reflect the Figma Designs
- Add snapshot and unit tests
- Update backend `listEcrDataService` & tests so that if no conditions are given, then only eCRs with no reportable conditions are returned. If `filterConditions` is undefined, it defaults to including all conditions.

## Related Issue
Fixes #2751

## Design Review

[Link to Figma prototype](https://www.figma.com/proto/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-Dec-2024)?node-id=4804-35695&node-type=instance&m=dev&scaling=min-zoom&content-scaling=fixed&page-id=4032%3A9311&starting-point-node-id=4806%3A39674)

## Acceptance Criteria

- [ ]  Condition filter is implemented according to designs
- [ ]  Condition list is populated by a custom backend query (see [FILTER: Query for all unique conditions #2749](https://github.com/CDCgov/phdi/issues/2749))
- [ ]  When user checks boxes for conditions and searches, only eCRs with those conditions in the RR are shown
- [ ]  ~~Reset button appears when filtering is applied~~ Wrote a separate ticket for this

## Additional Information
This ticket was getting big, so I'm breaking out a few other related tickets from this issue:

~~- The filter by conditions combo box should default to all conditions being selected. If no conditions are selected, only eCRs with no reportable conditions should appear.~~ Added in this ticket
~~- Modify button styling when in regular/hover/applied filter states~~ Added in this ticket
- Adding a No Conditions Reported option
- Adding a filter reset button
- Clicking anywhere outside of the button should close the button
- Pre-existing bug: Empty space below the eCR Library?

## Checklist

1. Pull down branch and open the eCR Library
2. Test that the interactions with the  filter conditions button and combo box follow the scenarios above (and any others you may think of)


angelathe:
@ashton-skylight Just thought of another question -- should the filter conditions combo box default to all conditions selected? Right now, it defaults to having selected no conditions (which acts the same as all conditions being selected), which may not make as much sense to users.

ashton-skylight:
@angelathe Yes, that's correct. If that's the interaction, I'm making revised designs for the initial state and how users continue through the dropdown. This will probably be a new ticket.

angelathe:
> @angelathe Yes, that's correct. If that's the interaction, I'm making revised designs for the initial state and how users continue through the dropdown. This will probably be a new ticket.
> 
> 

Great, thanks so much @ashton-skylight! I'll add this to the list of tickets that I'll write 👌

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,267 @@
...
+              aria-hidden="true"
+              focusable="false"
+              role="img"
+            >
+              <use
+                xlinkHref={`${basePath}/assets/img/sprite.svg#coronavirus`}
+              ></use>
+            </svg>
+          </span>
+          Reportable condition
```
JNygaard-Skylight, line 174: "Reportable condition", "Filter by Reportable Condition", and "Apply filter" seem to have inconsistent capitalizations. I'm not the designer/grammar expert, but something to consider.

angelathe: Thanks Josh! @ashton-skylight Thoughts?

mcmcgrath13:
I haven't done any code review yet, but a couple design notes. 

1. As a keyboard user, I have to navigate through every item in the drop down in order to hit the apply filter button. This is quite tedious. Can we set this up as a form so that when I hit enter, it "submit"'s?
2. After hitting apply filter, I had expected the drop down to disappear and the focused element to be the outer button again, but it stayed as is (maybe more of an @ashton-skylight question)
3. There isn't any real feedback that the filter "worked" unless something in view happens to change. Maybe a brief loading overlay would help? Or make sure the footer with the counts is always on screen? (the latter wouldn't need to be for this PR)

mcmcgrath13:
> > 1. As a keyboard user, I have to navigate through every item in the drop down in order to hit the apply filter button. This is quite tedious. Can we set this up as a form so that when I hit enter, it "submit"'s?
> 
> Hi @mcmcgrath13! Can you elaborate on # 1? The `Apply Filter` button should always be static (you shouldn't need to scroll through all the conditions in order to hit the button).
> 
> I'll defer to @ashton-skylight on these questions!

@angelathe when a user isn't using a mouse, but keyboard (or paddle, or similar) only. They currently have to scroll through every option to get the `Apply Filter` button in focus so it's selectable. If we make the whole filter a `<form>` and the `Apply Filter` button its `submit` then if they hit `enter` then it will work even if they haven't tabbed through all the way

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,267 @@
...
+ * Functional component for filtering eCRs in the Library based on reportable conditions.
+ * @returns The rendered FilterReportableConditions component.
+ * - Fetches conditions from the `/api/conditions` endpoint.
+ * - Users can select specific conditions or select all conditions.
+ * - Updates the browser's query string when the filter is applied.
+ */
+const FilterReportableConditions = () => {
+  const router = useRouter();
+  const pathname = usePathname();
+  console.log(pathname);
```
mcmcgrath13, line 37:
```suggestion

```

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,267 @@
...
+  const [appliedConditions, setAppliedConditions] = useState<{
+    [key: string]: boolean;
+  }>({});
+
+  // Fetch list of conditions
+  useEffect(() => {
+    const fetchConditions = async () => {
+      try {
+        // Fetch condition list and set default state to filtering on all conditions
+        const response = await fetch(`${basePath}/ecr-viewer/api/conditions`);
```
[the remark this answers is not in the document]
angelathe: We need the base path because `/ecr-viewer` basepath exists in prod but not in local dev. 

(Before, I had left it as `/api/conditions`, and it wasn't fetching when running in prod)

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+        }
+        const allConditions = await response.json();
+        setConditions(allConditions);
+
+        const initialFilterConditions = allConditions.reduce(
+          (dict: { [key: string]: boolean }, condition: string) => {
+            dict[condition] = true;
+            return dict;
+          },
+          {} as { [key: string]: boolean },
```
mcmcgrath13, line 65: would it make more sense to use a `Set` here given we're just tracking membership?

mcmcgrath13: though on second thought, that might be tedious with mutability rules

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const handleSelectAll = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { checked } = event.target;
+
+    // Loop through each condition checkbox and set the checked value
+    const checkboxes = document.querySelectorAll(
+      'input[id^="condition-"]',
+    ) as NodeListOf<HTMLInputElement>;
+    checkboxes.forEach((checkbox) => {
+      (checkbox as HTMLInputElement).checked = checked;
+    });
```
mcmcgrath13, line 109: this strikes me as "going around" react. Should we instead set the filter Conditions and let react handle the updated rendering?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+        >
+          <span className="width-205 usa-icon">
+            <svg
+              className="usa-icon"
+              aria-hidden="true"
+              focusable="false"
+              role="img"
+            >
+              <use
+                xlinkHref={`${basePath}/assets/img/sprite.svg#coronavirus`}
```
mcmcgrath13, line 169: same question on base path here

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: though a plain array with filtering/destructure add value might be simpler? then you could just set `filterConditions` to `conditions` on select all

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+                  key={"all"}
+                >
+                  <input
+                    id={"condition-all"}
+                    className="usa-checkbox__input"
+                    type="checkbox"
+                    value={"all"}
+                    onChange={handleSelectAll}
+                    checked={
+                      Object.keys(filterConditions).length === conditions.length
```
mcmcgrath13, line 203:
```suggestion
                      isAllSelected
```

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Yup, same as the answer above for this

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: I went back to `Joy of React` cheatsheet on inputs, and after looking at that, I'd think we could have one piece of state for conditions that is a dict with keys => bools.

Then when it's applied we update the query string, so we don't need to track that separately.

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+            className="usa-tag padding-05 bg-base-darker radius-md"
+            data-testid="filter-conditions-tag"
+          >
+            {Object.keys(filterConditions).length}
+          </span>
+        </Button>
+
+        {isFilterBoxOpen && (
+          <div className="usa-combo-box top-full left-0">
+            <div className="usa-combo-box margin-top-1 bg-white position-absolute radius-md shadow-2 z-top maxh-6205 width-4305">
```
mcmcgrath13, line 184: should this be a `fieldset` instead of a `div`?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+                      className="checkbox-color usa-checkbox"
+                      key={condition}
+                    >
+                      <input
+                        id={`condition-${condition}`}
+                        className="usa-checkbox__input"
+                        type="checkbox"
+                        value={condition}
+                        onChange={handleCheckboxChange}
+                        checked={!!filterConditions[condition]}
```
mcmcgrath13, line 228: why do we need the `!!` here if we're always setting a boolean?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+                <div className="border-top-1px border-base-lighter margin-x-neg-105"></div>
+                <Button
+                  id="button-filter-conditions-apply"
+                  className="margin-y-1 margin-x-0 padding-y-1 padding-x-205 flex-fill"
+                  aria-label="Apply filter"
+                  onClick={(e) => {
+                    e.preventDefault();
+                    setAppliedConditions(filterConditions);
+                    updateFilterConditionsQuery(filterConditions);
+                  }}
```
mcmcgrath13, line 252: when you add the `<form>` move this to the form's `onSubmit`

containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -352,8 +352,13 @@ export const generateFilterConditionsStatement = (
 ) => ({
   rawType: true,
   toPostgres: () => {
-    if (!filterConditions) {
-      return pgPromise.as.format("NULL IS NULL");
+    if (
+      !filterConditions ||
+      filterConditions.length === 0 ||
```
mcmcgrath13, line 357:
```suggestion
      !Array.isArray(filterConditions)
```

containers/ecr-viewer/src/app/components/Filters.tsx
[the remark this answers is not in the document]
mcmcgrath13: ahhhh - that'd do it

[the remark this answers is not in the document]
angelathe: Lol though I just realized I forgot to take out `/ecr-viewer` from when I was manually testing this, it should be `${basePath}/api/conditions`

containers/ecr-viewer/public/assets/img/sprite.svg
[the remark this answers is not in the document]
BobanL: I tested locally and I think just adding `<Icon.Coronavirus />` should work in place of this sprite.

ashton-skylight:
@mcmcgrath13 @angelathe Thanks for being our eyes on interaction patterns! We haven't been able to get the design script to run properly, and this a great help. 
2. Good call out! Agree, after the user hits "Apply Filter", the menu should automatically close, and the button would be the active state of the Filters Applied buttons. 
3. We have a skeleton loading state that could be applied after the user hits "Apply Filter" or the idea about the footer persisting works as well. @angelathe, that could be pushed to a different ticket!

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Thanks for all the comments @mcmcgrath13! Just reworked the state management and made the keyboard nav easier by using `form`. Let me know if you have any thoughts!

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  // Build list of conditions to filter on
+  const handleCheckboxChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value, checked } = event.target;
+    setFilterConditions((prev) => {
+      if (checked) {
+        return { ...prev, [value]: true };
+      } else {
+        return { ...prev, [value]: false };
+      }
```
mcmcgrath13, line 79:
```suggestion
      return { ...prev, [value]: checked };
```

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+        const allConditions = await response.json();
+        const initialFilterConditions = allConditions.reduce(
+          (dict: { [key: string]: boolean }, condition: string) => {
+            dict[condition] = true;
+            return dict;
+          },
+          {} as { [key: string]: boolean },
+        );
+        setFilterConditions(initialFilterConditions);
+        updateFilterConditionsQuery(initialFilterConditions);
```
mcmcgrath13, line 62: I think this logic is a bit backwards. We need to initialize the filter conditions based on the initial search param filters, but not update the filter conditions query.

Right now, if you update the filters, hit apply, then do a hard refresh, the filters will reset to all selected. (Could you add a test that covers this case?)

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+      } catch (error) {
+        console.error("Error fetching conditions:", error);
```
mcmcgrath13, line 64: should this manifest in a UI error as well?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+      .filter((key) => filterConditions[key] === true)
+      .join("|");
+    if (searchParams.get("condition") !== filterConditionsSearch) {
+      router.push(
+        pathname + "?" + createQueryString("condition", filterConditionsSearch),
+      );
+    }
+  };
+
+  const resetFilterConditions = () => {
```
mcmcgrath13, line 125: I think this is always used in conjunction with the setter, so maybe move that part into the method?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+                </legend>
+
+                {/* Select All checkbox */}
+                <div className="display-flex flex-column">
+                  <div
+                    className="checkbox-color usa-checkbox padding-bottom-1 padding-x-105"
+                    key={"all"}
+                  >
+                    <input
+                      id={"condition-all"}
```
mcmcgrath13, line 203:
```suggestion
                      id="condition-all"
```

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+                      className="usa-checkbox__input"
+                      type="checkbox"
+                      value={"all"}
```
mcmcgrath13, line 206:
```suggestion
                      value="all"
```

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Hm, makes sense that the query should persist over a reload, but struggling a bit with how to distinguish a page load between being the page's first load (no filters have been able to be applied yet, so we want to add all conditions to the query), and a reload from which a user deselected all conditions from the filter. 

Would you recommend explicitly naming the "deselect all" as a named value for the conditions param instead of just `condition=`?

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Could building the error UI for a specific filter box be made a different ticket? Don't think there are designs for this yet @ashton-skylight

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: do we put the param there with empty values if all is deselected? If so, I think that in itself is distinguishable from the case where the condition param has never been set to any value

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -0,0 +1,266 @@
+"use client";
+
+import React, { useCallback, useEffect, useState } from "react";
+import { Button } from "@trussworks/react-uswds";
+import { useRouter, usePathname, useSearchParams } from "next/navigation";
+
+const basePath =
+  process.env.NODE_ENV === "production" ? process.env.NEXT_PUBLIC_BASEPATH : "";
```
BobanL, line 8: `NEXT_PUBLIC_BASEPATH` was replaced with `BASE_PATH` in #2996 

When you merge in main you'll need to change this:
```suggestion
  process.env.NODE_ENV === "production" ? process.env.BASEPATH : "";
```

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -124,16 +114,30 @@ const FilterReportableConditions = () => {
     [searchParams],
   );
 
+  const deleteQueryParam = useCallback(
```
angelathe, line 117: @mcmcgrath13: I made some changes that seem to work, but I'm not really confident about the approach. 

By default, when a user has been modifying the filters and "Select all", the condition parameter would be filled with all the conditions. Is it a step backward to check if all conditions are selected and if so, delete the condition parameter entirely?

mcmcgrath13: deleting the param when all are selected seems good to me

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -109,16 +114,30 @@ const FilterReportableConditions = () => {
...
+    [searchParams],
+  );
+
   const updateFilterConditionsQuery = (filterConditions: {
     [key: string]: boolean;
   }) => {
     const filterConditionsSearch = Object.keys(filterConditions)
       .filter((key) => filterConditions[key] === true)
       .join("|");
     if (searchParams.get("condition") !== filterConditionsSearch) {
```
mcmcgrath13, line 133: this if won't catch the all selected un-changed case. Maybe it makes more sense to construct the path then check if it's the same as the current path before pushing?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -127,12 +146,13 @@ const FilterReportableConditions = () => {
       searchParams.get("condition")?.split("|") || [],
     );
 
-    return Object.fromEntries(
+    const prevConditions = Object.fromEntries(
       Object.entries(filterConditions).map(([key]) => [
         key,
         conditionsToTrue.has(key),
       ]),
     );
```
mcmcgrath13: this uses a different pattern than others - should it be switched over to the reduce as well?

It also doesn't (at least obviously) handle the all conditions case. 

Should we maybe use this in the initial `useEffect` as I think that it should be the same in both cases?

mcmcgrath13:
@angelathe it looks like the extra space the filters take up is offsettin the footer placement

angelathe:
> @angelathe it looks like the extra space the filters take up is offsetting the footer placement

Good call, just updated `height-ecrLibrary`

mcmcgrath13 (approved):
looking good from the technical side to me!

ashton-skylight:
Reviewed w/ Angela! Noting for future tickets - we'lll add a ticket for the "Apply Filter" to be inactive if no condition is selected

PR #2991, opened 2024-12-05
Style ecr-library sort

### LINKED ISSUE #2860: Styling sorting function of ECR Library
## What needs to be done
- Complete styling updates on the ECR Library around sorting

## Why it needs to be done
- This needs to be done in order for the styling to be consistent through the app. 

## Acceptance Criteria

- [ ] Move sort arrows to the far right of the columns
- [x]  The default sort is that the table is sorted by Received Date, with the most recent eCRs at the top. The arrow next to Received Date should be pointing up by default always to indicate this sort, but the column should not be highlighted blue until somebody clicks the arrow again.  Received date is already the default sort, just turn off highlighting
- [x]  Change highlighted column color to  #DFE1E2. Check with design
- [x] Fix outline around SortButton when clicked to be a more proper size

## Technical Notes
- There is no figma design, consult with Ashton for any specifics 
- Based on USWDS sort table
- Sorting was added in #2233

lina-roth:
# PULL REQUEST
Screenshot 2024-12-05 at 10 35 08 AM

## Summary
- Remove outline on arrow 
- Remove highlight color 
- Move arrows to the far right ( sometimes appears as if not at the end because of the changing width of the rckms rule summary column)

## Related Issue
Fixes #2860 

## Acceptance Criteria
- [ ] Move sort arrows to the far right of the columns 
- [ ] ~~The default sort is that the table is sorted by Received Date, with the most recent eCRs at the top. The arrow next to Received Date should be pointing up by default always to indicate this sort, but the column should not be highlighted blue until somebody clicks the arrow again. Received date is already the default sort, just turn off highlighting~~
- [ ] Change highlighted column color to #DFE1E2. Check with design
- [ ] Fix outline around SortButton when clicked to be a more proper size

## Checklist


ashton-skylight:
Reviewed! Lina and I chatted about moving the filtering arrow to the far right for received date and encounter date. As the ticket calls out, due to the dynamic nature of the RCKMS summary, we had two options: (1) remove the responsive columns from RCKMS or (2) push the filtering arrows to the far right. We chose to keep the responsive columns in favor of user readability for the RCKMS summary.

mcmcgrath13 (changes requested):
Removing the highlight on the arrows is an accessibility issue. If you keyboard navigate through the viewer there is now no visual indication you are on a sort button that is interactive (Just kinda goes into the void until you end up on the names in the list)

mcmcgrath13:
I think you can get the no outline for pointer users, but still outline with keyboard navigators by re-adding the styles with the `:focus-visible` selector

```
button:focus-visible {
  outline: 5px auto -webkit-focus-ring-color;
}
```

ashton-skylight:
@mcmcgrath13 @lina-roth Great callout, thank you, Mary! Looking at the component now. The default state of the table with sorting does not highlight the column. If we keep that as the default state that should work, and then if the user clicks on the column, the header could change to base-lighter, #A9AEB1.

Screenshot 2024-12-05 at 11 57 08 AM

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -184,39 +184,68 @@ h4 {
...
   }
   &:focus {
-    outline: none;
     background-color: inherit; /* Removes fill after click */
   }
 }
 
 .sortable-column{
+  &.usa-button:focus{
+    //outline: none;
```
mcmcgrath13:
```suggestion
```
do we still need this sub-rule at all? the `margin-left: auto` is on the base button as well

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -184,39 +184,68 @@ h4 {
...
+    margin-left: auto;
+    padding: .65rem .65rem .65rem .65rem;
+  }
   background-color: transparent;
   @include add-background-svg("usa-icons/sort_arrow");
 }
 
 .sortable-desc-column{
+  &.usa-button:focus{
+    //outline: none;
```
mcmcgrath13: same question here

though also thinking that the `.sortable-asc-column` `.sortable-column` and `.sortable-desc-column` have a lot of overlap for one changing thing. Can we dry this out? maybe `.sortable-column` and then `.sort-choose` `.sort-asc` and `.sort-desc` as modifier classes or similar?

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -230,6 +259,15 @@ h4 {
...
   font-family: "Source Sans Pro Web";
 
+  &.usa-table th{
+    &[aria-sort=descending] {
+      background-color: #DFE1E2;
+    }
+    &[aria-sort=ascending] {
+      background-color: #DFE1E2;
+    }
+  }
```
mcmcgrath13, line 266: I'm not seeing this kick in when sorting

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: removed sub rule

mcmcgrath13:
when encounter date is sorted, the arrow on patient name is cut off
image

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: the only thing is, the selectors are coded to how the css is, so re-doing this is going to be a big lift

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: its matching the highlight to the background per design request to not show the highlighted column

lina-roth:
@mcmcgrath13 arrow isnt cutoff for me when sorting by encounter date

containers/ecr-viewer/src/styles/custom-styles.scss
mcmcgrath13: ooooh - could you add a comment to that effect?

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: see containers/ecr-viewer/src/app/components/SortButton.tsx

mcmcgrath13: hmm... I think it could still be simplified by doing something like

```
.sortable-column, .sortable-column-desc, .sortable-coloumn-asc {
   &.usa-button {
      ...
   }
}
```

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: done

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: how would i differentiate between what arrow to show if they are all lumped together?

mcmcgrath13:
@lina-roth I think the difference has to do with what the `rem` is set to. I use a smaller font than standard (14), so I think that's what's causing the difference. We should probably account for that. I don't see the icon cut off on the main branch.

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -184,39 +184,56 @@ h4 {
...
-    outline: none;
     background-color: inherit; /* Removes fill after click */
   }
 }
 
 .sortable-column{
+  &.usa-button{
+    justify-content: flex-end;
+    margin-left: auto;
+    padding: .65rem .65rem .65rem .65rem;
```
mcmcgrath13, line 206: since these are fit to an icon that is sized in px - we should probably use px here instead of rem. Inspector says the icons are 24x24

lina-roth: done

containers/ecr-viewer/src/styles/custom-styles.scss
lina-roth: done

containers/ecr-viewer/src/styles/custom-styles.scss
```diff
@@ -184,39 +184,47 @@ h4 {
...
 }
 
-.sortable-column{
-  background-color: transparent;
-  @include add-background-svg("usa-icons/sort_arrow");
+.sortable-column, .sortable-desc-column, .sortable-asc-column{
+    &.usa-button{
+      margin-left: auto;
+      justify-content: flex-end;
+      padding: 24px;
```
mcmcgrath13, line 208:
```suggestion
      padding: 12px;
```
1/2 image height/width

mcmcgrath13 (approved):
LGTM (assuming acceptance of 12px change)

thank you!!

Issue #2536, opened 2024-09-10
Add basepath support to eCR Viewr


emmastephenson:
## What needs to be done

Add basepath support to the eCR Viewer. 

## Why it needs to be done

Without basepath support, the eCR Viewer runs into gateway errors for our CSS/JS files when deployed to a cloud environment.

## Acceptance Criteria

- [ ] Basepath support added to the application
- [ ] New release cut
- [ ] AWS/Azure are running new code and working as intended (the Viewer can load)

## Technical Notes

The TEFCA Viewer already has this basepath support. We should be able to use this example to get things working.

https://github.com/CDCgov/phdi/blob/37b19b04329c72baa7770db92c7d8e184281826a/containers/tefca-viewer/next.config.js#L27

PR #2996, opened 2024-12-05
fix build without pg promise variables

BobanL:
# PULL REQUEST

## Summary
- Currently builds/images fail without any environment variables because PG-promise db connection is being made on the top level.
- Remove base .env since it just has local configuration
- Remove NEXT_PUBLIC_BASEPATH in favor of hard coding the value. [NEXT_PUBLIC](https://nextjs.org/docs/app/building-your-application/deploying#environment-variables:~:text=runtime%20environment%20variables.-,By%20default%2C%20environment%20variables%20are%20only%20available%20on%20the%20server.%20To,.,-You%20safely%20read) values are only modifiable during build time anyways so they aren't needed in this case.
- Add .dockerignore to improve local build speed

## Related Issue
Fixes #

## Acceptance Criteria

## Additional Information

## Checklist


containers/ecr-viewer/next.config.js
```diff
@@ -20,10 +20,7 @@ const nextConfig = {
     ];
   },
   output: "standalone",
-  basePath:
-    process.env.NODE_ENV === "production"
-      ? process.env.NEXT_PUBLIC_BASEPATH
-      : "",
+  basePath: process.env.NODE_ENV === "production" ? "/ecr-viewer" : "",
```
austin-hall-skylight, line 23: @emmastephenson do we think that STLT's are ever going to want to configure this path? I couldn't find any info in [the original ticket](https://github.com/CDCgov/phdi/issues/2536) about whether it should be configurable or not.

emmastephenson: I think this should be fine from the STLT perspective; most of their networking changes will likely be before our custom paths. But would definitely double-check this with @rin-skylight and @alismx, they'll be most familiar with it.

austin-hall-skylight: Ok, I've reviewed everything and am ready to approve as soon as we get an answer about this from devops.

rin-skylight: The path reflected in this change is the one that we have pre-programmed in our App Gateway object, so I see no issue for use in Azure. I would defer a merge until @alismx affirmatively answers in regard to AWS.

alismx: The value for `NEXT_PUBLIC_BASEPATH` defaults to `/ecr-viewer` in the AWS repos but can be overridden. If we're removing it here, I'd like to remove it there. From an AWS perspective, I see no problem with moving forward on this.

austin-hall-skylight (approved):

PR #3005, opened 2024-12-06
Handle empty travel histories as unavailable data

### LINKED ISSUE #2978: Travel history table appears even when travel history not present
## Description

The travel history table appears in the Patient Social History section, even if there is no data populated for travel history.

## Steps to reproduce

1. Load eCR #6c211958-6972-4899-8b7a-0ce3cae8c570 in the Viewer
2. See Travel History table under Social History, but no information present 

## Expected Result

If there is no data for Travel History, it should show under "Unavailable Data".

## Priority

#### Priority
- [x] **Medium priority** _Typically, medium-priority bugs do not affect customers directly and therefore can be fixed in the normal course of testing and development._

mcmcgrath13:
# PULL REQUEST

## Summary
If a travel history observation exists, but has no useful data, treat it as unavailable data instead of rendering a table with rows full of `No Data`.

## Related Issue
Fixes #2978

## Additional Information

http://localhost:3000/view-data?id=d9039113-a9e0-4ae4-a8c6-6b46327d3af8

Before:
image

After:
image


containers/ecr-viewer/src/app/services/formatService.tsx
```diff
@@ -21,12 +21,6 @@ export interface TableJson {
   tables?: TableRow[][];
 }
 
-export interface TableJson {
```
mcmcgrath13, line 24: this is a duplicate of the definition immediately above

containers/ecr-viewer/src/app/services/evaluateFhirDataService.ts
```diff
@@ -263,6 +263,7 @@ export const evaluateSocialData = (
       value: evaluateValue(fhirBundle, mappings["patientMaritalStatus"]),
     },
   ];
+  console.log({ res: evaluateData(socialData) });
```
mcmcgrath13, line 266:
```suggestion
```

BobanL (approved):
One nit, but otherwise looks good!

containers/ecr-viewer/src/app/tests/components/EvaluateTable.test.tsx
```diff
@@ -68,22 +68,48 @@ describe("Evaluate table", () => {
...
+            infoPath: "getId",
+          },
         ]}
       />,
     );
 
-    expect(screen.getByText("Col1")).toBeInTheDocument();
-    expect(screen.getByText("No data")).toBeInTheDocument();
+    expect(getByText("Col1")).toBeInTheDocument();
+    expect(queryByText("No data")).not.toBeInTheDocument();
```
BobanL, line 112: Is there a reason why we are replacing the screen query with the one from render? The docs [recommend using screen.](https://testing-library.com/docs/qwik-testing-library/api/#render:~:text=If%20your%20baseElement%20is%20the%20default%20document.body%2C%20we%20recommend%20using%20screen%20rather%20than%20bound%20queries.)

mcmcgrath13: I can switch back. I was copying an example from their docs 😅

PR #3018, opened 2024-12-10
Refactor save service response object

BobanL:
# PULL REQUEST

## Summary
- No longer return NextResponse from service since it's more difficult to read and handle in the service layer
- Do not return error messages in response since it may leak information (see a similar [code scanning issue](https://github.com/CDCgov/phdi/security/code-scanning/6)).
- Save both fhir bundle and metadata async
- Remove pool.close() since it [isn't recommended](https://github.com/tediousjs/node-mssql#:~:text=Also%20notice%20that,of%20the%20script.).


containers/ecr-viewer/src/app/api/save-fhir-data/route.ts
```diff
@@ -43,13 +43,21 @@ export async function POST(request: NextRequest) {
...
+    const saveResp = await saveWithMetadata(
       fhirBundle,
       ecrId,
       saveSource,
       requestBody.metadata,
     );
+    return NextResponse.json(
+      { message: saveResp.message },
+      { status: saveResp.status },
+    );
```
mcmcgrath13, line 55:
```suggestion
    const {message, status} = await saveWithMetadata(
      fhirBundle,
      ecrId,
      saveSource,
      requestBody.metadata,
    );
    return NextResponse.json(
      { message },
      { status },
    );
```

minor style suggestion (feel free to take or leave) - can be applpied below as well

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -74,15 +77,17 @@ export const saveToS3 = async (fhirBundle: Bundle, ecrId: string) => {
...
+    return {
+      message: "Success. Saved FHIR Bundle to S3: " + ecrId,
+      status: 200,
+    };
   } catch (error: any) {
-    return NextResponse.json(
-      { message: "Failed to insert data to S3. " + error.message },
-      { status: 500 },
-    );
+    console.error(`Failed to save FHIR Bundle to S3: ${error.message}`);
```
mcmcgrath13, line 86: should we take advantaged of console's more structured logging here and do something like 
```
console.error({ message: "Failed to save FHIR Bundle to S3", error })
```

then any other fields in error would also be available (maybe? hopefully? at least this works well with front end logs, not as familiar with our server-side logging)

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -120,19 +127,18 @@ export const saveToAzure = async (fhirBundle: Bundle, ecrId: string) => {
...
-        message:
-          "Failed to insert FHIR bundle to Azure Blob Storage. " +
-          error.message,
-      },
-      { status: 500 },
+    console.error(
+      `Failed to save FHIR Bundle to Azure Blob Storage: ${error.message}`,
     );
+    return {
+      message: "Failed to save FHIR bundle to Azure Blob Storage.",
```
mcmcgrath13, line 139: should we include the `ecrId` in these strings? (I know this isn't new from this PR, but since we're removing context from the client reponse, but be worth adding other context in?)

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -142,27 +148,60 @@ export const saveToAzure = async (fhirBundle: Bundle, ecrId: string) => {
...
+      );
+    } else {
+      return {
+        message: "Unknown metadataSaveLocation: " + metadataSaveLocation,
+        status: 400,
+      };
+    }
+  } catch (error: any) {
+    return {
+      message: "Failed to save FHIR data with metadata. " + error.message,
```
mcmcgrath13, line 202: do we want to remove this error message interpolation as well? or do we know this is fine because it's one of ours?

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
BobanL: Great idea!

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -12,37 +11,46 @@ import { get_pool } from "../services/sqlserver_db";
...
-      },
-      { status: 500 },
-    );
+    console.error({
+      message: `Failed to save FHIR bundle to postgres.`,
+      error,
+      ecrId,
+    });
+    return {
+      message: "Failed to save FHIR bundle.",
```
mcmcgrath13, line 50: should we include the `ecrId` in the error message?

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -74,15 +81,21 @@ export const saveToS3 = async (fhirBundle: Bundle, ecrId: string) => {
...
-      { message: "Failed to insert data to S3. " + error.message },
-      { status: 500 },
-    );
+    console.error({
+      message: "Failed to save FHIR bundle to S3.",
+      error,
+      ecrId,
+    });
+    return {
+      message: "Failed to save FHIR bundle.",
```
mcmcgrath13, line 95: same here (and all where this applies)

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -505,42 +550,30 @@ export const saveMetadataToPostgres = async (
...
-        ecrId,
-      );
-    } else {
-      metadataResult = NextResponse.json({
-        message: "Unknown metadataSaveLocation: " + metadataSaveLocation,
-      });
-    }
+    [fhirDataResult, metadataResult] = await Promise.all([
+      saveFhirData(fhirBundle, ecrId, saveSource),
+      saveFhirMetadata(ecrId, metadataSaveLocation, metadata as BundleMetadata),
```
mcmcgrath13, line 568: does this mean we could now end up in a situation where we have metadata, but not a bundle? (the opposite I suppose was always possible). Is this an invariant we should allow?

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
BobanL: The eCR Id is provided to us in the body of the request so I don't think it would any value, but if you think otherwise I can add that in quickly.

mcmcgrath13: It just seems odd to me that we include it on the success message, but not the failure. The asymmetry doesn't seem to have a reason for it

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
BobanL: Yes there is a chance! I've made a tech debt ticket since this issue existed before is out of scope. #3034

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
BobanL: I've just decided to remove ecr id from response since it's the ID that's in the body of the request.

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -505,42 +550,30 @@ export const saveMetadataToPostgres = async (
...
+    ]);
   } catch (error: any) {
-    return NextResponse.json(
-      { message: "Failed to save FHIR data with metadata. " + error.message },
-      { status: 500 },
-    );
+    const message = "Failed to save FHIR data with metadata.";
+    console.error({ message, error, ecrId });
+    return {
+      message: message,
```
mcmcgrath13, line 574:
```suggestion
      message,
```

PR #3042, opened 2024-12-12
Prefer Annotating Types Over Asserting

JNygaard-Skylight:
# PULL REQUEST

## Summary
Where possible, annotate the returned value of the `evaluate` function instead of asserting it.

## Related Issue
Fixes # n/a

## Acceptance Criteria
n/a

## Additional Information
We use the `evaluate` function from the `fhirpath` library throughout. the ecr-viewer. However, it is typed to return `any`. To take advantage of type hints in IDEs, we need to either assert the expected type, e.g., `const rrDetails: Observation[] = evaluate(fhirBundle, mappings.rrDetails);` or assert the expected type, e.g., `const rrDetails  = evaluate(fhirBundle, mappings.rrDetails) as Observation[];`.  Both options more-or-less do the same thing, but there are a few reasons to prefer annotation. In brief it is the difference between saying, "I know the type of this value, and throw an error if it does not fit this type," and "Treat this value as this type no matter what."

See:
- https://stackoverflow.com/questions/47994926/detailed-differences-between-type-annotation-variable-type-and-type-assertion
- https://medium.com/@bsalwiczek/type-annotation-vs-assertion-in-typescript-one-important-difference-4f4df715b5fe


BobanL (approved):

containers/ecr-viewer/src/app/view-data/components/PatientBanner.tsx
```diff
@@ -25,8 +25,9 @@ const PatientBanner = ({ bundle, mappings }: PatientBannerProps) => {
       </span>
       <span className="patient-banner-dob">
         {bundle && mappings
-          ? formatDate((evaluate(bundle, mappings.patientDOB) as any[])[0]) ||
-            ""
+          ? formatDate(
+              (evaluate(bundle, mappings.patientDOB) as string[])[0],
+            ) || ""
           : ""}
```
mcmcgrath13, line 31:
```suggestion
          && formatDate(
              (evaluate(bundle, mappings.patientDOB) as string[])[0],
            )}
```
not related to your changes, but if you want to simplify while you're here 🙏

mcmcgrath13 (approved):
thanks! I like this change stylistically, but given these types all return `any`, will typescript ever be able to tell us that they're not returning the thing we expect? (still want to do it either way, this is a curiosity)

JNygaard-Skylight:
> thanks! I like this change stylistically, but given these types all return `any`, will typescript ever be able to tell us that they're not returning the thing we expect? (still want to do it either way, this is a curiosity)

I don't think Typescript will be able to help us if `evaluate` does not return what we expect, but it should help us when we use the variable defined.

PR #3053, opened 2024-12-12
set ecr viewer path permanently

BobanL:
# PULL REQUEST

## Summary
- Permanently set the base_path in ecr-viewer to /ecr-viewer so we can have a consistent experience between both dev and production
- Update middleware matchers to match properly on all paths
- Replace all A tags with Link tags


mcmcgrath13:
Could this prevent deployments from hosting this on a shared server/domain where they need a custom base path?

containers/ecr-viewer/src/app/components/EcrTableClient.tsx
```diff
@@ -304,9 +302,9 @@ const DataRow = ({ item }: { item: EcrDisplay }) => {
   return (
     <tr>
       <td>
-        <a href={`${basePath}/view-data?id=${item.ecrId}`}>
+        <Link href={`/view-data?id=${item.ecrId}`}>
```
mcmcgrath13, line 305: not really from this PR< but I believe `Link` should handle the basepath automagically, so this should change no matter what

BobanL: Yep - with [Link base path will be appended](https://nextjs.org/docs/app/api-reference/config/next-config-js/basePath#links)

BobanL:
> Could this prevent deployments from hosting this on a shared server/domain where they need a custom base path?

With our current deploy, the base path is already hard coded as `/ecr-viewer`. This change just makes sure local development uses that same path rather than `/`.

mcmcgrath13:
> > Could this prevent deployments from hosting this on a shared server/domain where they need a custom base path?
> 
> With our current deploy, the base path is already hard coded as `/ecr-viewer`. This change just makes sure local development uses that same path rather than `/`.

But could a deployment hypothetically override that environment variable to something else? I'm not against making the local version match the default for production, moreso worried that we're taking out flexibility that we are going to need. Maybe @alismx has intuition here?

BobanL:
> > > Could this prevent deployments from hosting this on a shared server/domain where they need a custom base path?
> > 
> > 
> > With our current deploy, the base path is already hard coded as `/ecr-viewer`. This change just makes sure local development uses that same path rather than `/`.
> 
> But could a deployment hypothetically override that environment variable to something else? I'm not against making the local version match the default for production, moreso worried that we're taking out flexibility that we are going to need. Maybe @alismx has intuition here?

I don't think this ever really worked in production anyways because the [basePath in the next js config is set at build time](https://nextjs.org/docs/app/api-reference/config/next-config-js/basePath#:~:text=Good%20to%20know%3A%20This%20value%20must%20be%20set%20at%20build%20time%20and%20cannot%20be%20changed%20without%20re%2Dbuilding%20as%20the%20value%20is%20inlined%20in%20the%20client%2Dside%20bundles.). The work arounds are [quite hacky](https://github.com/vercel/next.js/discussions/41769#discussioncomment-9329938) and I don't think that's a path we should take unless we get firm requirements. 
A [similar conversation](https://github.com/CDCgov/phdi/pull/2996#discussion_r1876233718) was had around this without any concern from Rin or Alis.

containers/ecr-viewer/src/app/view-data/retrieval-failed.tsx
```diff
@@ -31,6 +32,7 @@ const RetrievalFailed = () => (
             <br /> to troubleshoot the issue with the DIBBs team.
           </div>
         </div>
+        <BackButton className="margin-top-3 font-sans-md text-primary" />
```
mcmcgrath13, line 35: this seems unrelated to everything else in the PR - what am I missing?

BobanL: You are right - this is unrelated to the change of set ecr viewer path and I should have mentioned this in the PR body. This is the only error message that didn't have a back button so I check with Ashton and she agreed that we should add one here too.

mcmcgrath13 (approved):

PR #3059, opened 2024-12-13
fix sql server pool issues

BobanL:
# PULL REQUEST

## Summary
- Currently there are scenarios where transactions are open, but NOT closed. This can cause all connections to be open and not allowing any new connections
- Refactor condition service layer to not use transactions at all and to simply return an array of conditions

## Additional Information
Proof it still works:
image


austin-hall-skylight (approved):
🎄

PR #3072, opened 2024-12-17
Filter by Date (Part 1)

### LINKED ISSUE #2752: FILTER: Filter By Date (Part 1 - Default selections)
## User Story

As a user of the eCR Library, I want to filter by date, so that I can only see eCRs from within a given date range.

Example: I only want to see eCRs for the last week, so I can identify issues.

## Acceptance Criteria

- [x] "Filter By Received Date" implemented for:
    - [x] Last 24 hours
    - [x] Last 7 days
    - [x] Last 30 days
    - [x] Last 3 months 
    - [x] Last 6 months
    - [x] Last year
- [x] When user applies filter for any of the above, only eCRs received in that date range are included
- [x] My selection is shown as the button text (e.g. "Last 7 days")
- [x] By default, only eCRs from the last 24h are shown

## Scenarios

_Scenario 1_
- I click on the date filter button
- A menu appears with selectable presets
- I choose "Last 7 days"
- I click "Apply Filter"
- Only eCRs from the last week are shown

## Design Notes

[Link to Figma prototype](https://www.figma.com/proto/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-Dec-2024)?node-id=4806-39714&node-type=canvas&m=dev&scaling=min-zoom&content-scaling=fixed&page-id=4032%3A9311&starting-point-node-id=4806%3A39674)

## Additional context

This ticket does NOT include implementing or showing the custom date filter - that is a separate ticket.

This ticket includes both the frontend design implementation and the backend querying needed for full functionality.

angelathe:
# PULL REQUEST

## Summary
- Add filter by date range to eCR Library
- Add backend querying on date created in `listEcrDataService.ts`
- Make reusable components/hook between `FilterbyReportableCondition` and `FilterbyDate` components
- Create `FiltersService.ts` to house date-related logic (see comment I wrote below)
- Add/modify tests

## Related Issue
Fixes #2752 

## Acceptance Criteria
- [ ] "Filter By Received Date" implemented for:
    - [ ] Last 24 hours
    - [ ] Last 7 days
    - [ ] Last 30 days
    - [ ] Last 3 months 
    - [ ] Last 6 months
    - [ ] Last year
- [ ] When user applies filter for any of the above, only eCRs received in that date range are included
- [ ] My selection is shown as the button text (e.g. "Last 7 days")
- [ ] By default, only eCRs from the last 24h are shown

## Screenshot
https://github.com/user-attachments/assets/29143ed7-4d6c-4b1d-9d45-4896505f3129

## Additional Information
Left a couple questions/comments in the thread.

**Random notes:**
- For all the date range options except for "Last 24 hours", the start day time begins at midnight.
- Will implement Custom Date filtering in another ticket.

**For testing:**
- I manually changed dates in my db client for about 10 eCRs per date range option to see whether the filter by date component was working. For example:
    - [ ] Last 24 hours: 16 Dec 2024 (at time of creating this PR)
    - [ ] Last 7 days: 14 Dec 2024
    - [ ] Last 30 days: 30 Nov 2024
    - [ ] Last 3 months: 30 Sept 2024
    - [ ] Last 6 months: 30 June 2024
    - [ ] Last year: 20 Dec 2023


containers/ecr-viewer/src/app/tests/services/FiltersService.test.ts
```diff
@@ -0,0 +1,102 @@
+import {
+  DateRangeOptions,
+  convertDateOptionToDateRange,
+} from "@/app/services/FiltersService";
+
+describe("convertDateOptionToDateRange", () => {
```
angelathe, line 6: Don't know how necessary these tests really are...

containers/ecr-viewer/src/app/services/FiltersService.ts
angelathe, line 1: Any other ideas of where these things can live? I don't think this fits as a "service" file, but `convertDateOptionToDateRange` can't live in `Filters.tsx` (client-side) and be imported to `page.tsx` (server-side).

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -238,18 +204,141 @@ const FilterReportableConditions = () => {
...
+              onChange={handleDateOptionChange}
+              checked={filterDateOption === option}
+            />
+            <label
+              className="line-height-sans-6 font-sans-xs margin-y-0 usa-radio__label"
+              htmlFor={`filter-date-${option}`}
+            >
+              {dateRangeLabels[option]}
+            </label>
+          </div>
```
angelathe, line 290: Any ideas on how to make these sections reusable for both the 'Filter by Condition' and 'Filter by Date' components without needing to pass parameters for essentially every single value?

mcmcgrath13:
While the 24 hr default makes sense for real users, in development it makes for a pretty bad experience. Can we have a different default there? Ideally, I'd like to see all of them all of the time

mcmcgrath13:
I'm just noticing that when I submit the form as a keyboard user, my focus is back at the main filter button, but there's no indicator of it via the usual styling. I wonder if we need to do a bit more there to get that correctly working (I'm pretty sure it should be highlighted at least, but that's intuition more than research)

containers/ecr-viewer/src/app/components/EcrTable.tsx
```diff
@@ -21,11 +22,13 @@ const EcrTable = async ({
...
   filterConditions,
+  filterDate,
 }: {
   currentPage: number;
   itemsPerPage: number;
   sortColumn: string;
   sortDirection: string;
+  filterDate: DateRangePeriod;
   searchTerm?: string;
   filterConditions?: string[];
```
mcmcgrath13, line 33:
```suggestion
  searchTerm?: string;
  filterDate: DateRangePeriod;
  filterConditions?: string[];
```

super minor - just keeps the filters next to each other

containers/ecr-viewer/src/app/components/EcrTable.tsx
```diff
@@ -36,6 +39,7 @@ const EcrTable = async ({
     itemsPerPage,
     sortColumn,
     sortDirection,
+    filterDate,
     searchTerm,
     filterConditions,
```
mcmcgrath13, line 44:
```suggestion
    searchTerm,
    filterDate,
    filterConditions,
```

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -259,4 +348,104 @@ const FilterReportableConditions = () => {
...
+const useQueryParam = () => {
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+
+  // Set a query param with a specific value
+  const setQueryParam = useCallback(
+    (key: string, value: string) => {
+      const params = new URLSearchParams(searchParams.toString());
+      params.set("page", "1");
```
mcmcgrath13, line 366: maybe call out in the docs that this hook always resets page to 1. That may make it unsuitable for some query param management

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -238,18 +204,141 @@ const FilterReportableConditions = () => {
...
+      </div>
+    );
+  };
+
+  return (
+    <div>
+      <div className="position-relative display-flex flex-column">
+        <Button
+          className={"margin-right-0 filters-applied"}
+          id="button-filter-date"
```
mcmcgrath13, line 302: for the focus thing, I think you need to add a ref to this button and on the reset restore focus to that ref

mcmcgrath13: something like
```
const openBtnRef = React.useRef();

// in reset code
openBtnRef.current.focus()

// in button jsx
ref={openBtnRef}
```

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -238,18 +204,141 @@ const FilterReportableConditions = () => {
...
-                    type="submit"
-                    id="button-filter-conditions-apply"
-                    className="margin-y-1 margin-x-0 padding-y-1 padding-x-205 flex-fill"
-                    aria-label="Apply Filter"
-                  >
-                    Apply Filter
-                  </Button>
-                </div>
+                {ApplyFilterButton(
+                  "button-filter-conditions-apply",
```
mcmcgrath13, line 208: are these IDs really needed? they don't seem to be used anywhere and this could simplify the abstracting

containers/ecr-viewer/src/app/services/FiltersService.ts
mcmcgrath13: maybe just a generic `date-utils` type file?

containers/ecr-viewer/src/app/services/FiltersService.ts
```diff
@@ -0,0 +1,75 @@
...
+  LastYear = "last-year",
+}
+
+export const dateRangeLabels: { [key in DateRangeOptions]: string } = {
+  [DateRangeOptions.Last24Hours]: "Last 24 hours",
+  [DateRangeOptions.Last7Days]: "Last 7 days",
+  [DateRangeOptions.Last30Days]: "Last 30 days",
+  [DateRangeOptions.Last3Months]: "Last 3 months",
+  [DateRangeOptions.Last6Months]: "Last 6 months",
+  [DateRangeOptions.LastYear]: "Last year",
```
mcmcgrath13, line 21: could this be computed instead by replacing the `-` with ` ` and then applying `toSentenceCase`? (or maybe this is title case)

containers/ecr-viewer/src/app/services/FiltersService.ts
```diff
@@ -0,0 +1,75 @@
...
+      };
+    case DateRangeOptions.Last7Days:
+      const sevenDaysAgo = new Date();
+      sevenDaysAgo.setDate(currentEndDate.getDate() - 7);
+      sevenDaysAgo.setHours(0, 0, 0, 0);
+      return { startDate: sevenDaysAgo, endDate: currentEndDate };
+    case DateRangeOptions.Last30Days:
+      const thirtyDaysAgo = new Date();
+      thirtyDaysAgo.setDate(thirtyDaysAgo.getDate() - 30);
+      thirtyDaysAgo.setHours(0, 0, 0, 0);
```
mcmcgrath13, line 51: would it be worth making little helpers for `daysAgoDate(n)` and `monthsAgoDate(n)` (year could maybe use this with 12)

containers/ecr-viewer/src/app/services/FiltersService.ts
```diff
@@ -0,0 +1,75 @@
...
+  filterOption: string,
+  endDate?: Date,
+) {
+  const today = new Date();
+  const currentEndDate = endDate || today;
+
+  switch (filterOption) {
+    case DateRangeOptions.Last24Hours:
+      return {
+        startDate: new Date(currentEndDate.getTime() - 24 * 60 * 60 * 1000),
```
mcmcgrath13, line 40: can this also use the `daysAgoDate` just without the hours reset? maybe add that as an optional param so it's `daysAgoDate(n, zeroTime=true)` or similar

containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -91,14 +96,19 @@ async function listEcrDataPostgres(
...
 ): Promise<EcrDisplay[]> {
   const { database } = getDB();
   const list = await database.manyOrNone<CoreMetadataModel>(
     "SELECT ed.eICR_ID, ed.patient_name_first, ed.patient_name_last, ed.patient_birth_date, ed.date_created, ed.report_date, ed.report_date,  ARRAY_AGG(DISTINCT erc.condition) AS conditions, ARRAY_AGG(DISTINCT ers.rule_summary) AS rule_summaries FROM ecr_data ed LEFT JOIN ecr_rr_conditions erc ON ed.eICR_ID = erc.eICR_ID LEFT JOIN ecr_rr_rule_summaries ers ON erc.uuid = ers.ecr_rr_conditions_id WHERE $[whereClause] GROUP BY ed.eICR_ID, ed.patient_name_first, ed.patient_name_last, ed.patient_birth_date, ed.date_created, ed.report_date $[sortStatement] OFFSET $[startIndex] ROWS FETCH NEXT $[itemsPerPage] ROWS ONLY",
     {
-      whereClause: generateWhereStatementPostgres(searchTerm, filterConditions),
+      whereClause: generateWhereStatementPostgres(
+        filterDate,
+        searchTerm,
+        filterConditions,
```
mcmcgrath13, line 110: very minor nit, can we have it be search, filter, filter?

containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -209,48 +221,61 @@ const processExtendedMetadata = (
...
  * Retrieves the total number of eCRs stored in the ecr_data table.
+ * @param filterDate - The date (range) to filter on
  * @param searchTerm - The search term used to filter the count query
  * @param filterConditions - The array of reportable conditions used to filter the count query
  * @returns A promise resolving to the total number of eCRs.
  */
 export const getTotalEcrCount = async (
+  filterDate: DateRangePeriod,
   searchTerm?: string,
   filterConditions?: string[],
```
mcmcgrath13, line 232: how many of these do we plan to add? would it make sense to switch to an config-object style vs ordered args?

containers/ecr-viewer/src/app/tests/services/FiltersService.test.ts
mcmcgrath13: since it is just kind of reimplementing the actual code, I tend to agree. I wonder if there's a more robust way to cross-check this 🤔

mcmcgrath13 (commented):
meta-question: how does this interact with time zones?

containers/ecr-viewer/src/app/components/EcrTable.tsx
angelathe: Hm I originally did that but TS said I couldn't list a required param after an optional?

mcmcgrath13: womp womp. The things we do to appease typescript 😓

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Yup they're not really needed, I'll remove them! Is it generally fine practice to not include IDs?

mcmcgrath13: yeah, if they're not used anywhere (testing, labels, etc), then they aren't needed

containers/ecr-viewer/src/app/services/listEcrDataService.ts
angelathe: Currently, I think we only have plans for the two filters, but if we add anymore this would be a good idea!

mcmcgrath13:
A design question for @ashton-skylight 

Right now you can open both filters and they overlap - how do you want this to look/work?
image

ashton-skylight:
@mcmcgrath13 Not sure if it's in this ticket, but the final interaction will only allow the user to open one filter at a time. We can also increase spacing between the filters if needed

angelathe:
> @mcmcgrath13 Not sure if it's in this ticket, but the final interaction will only allow the user to open one filter at a time. We can also increase spacing between the filter buttons if needed

@mcmcgrath13 @ashton-skylight This will be worked on in a different [ticket](https://app.zenhub.com/workspaces/streamline-ecr-6480bf2ee530095ab41ebbe9/issues/gh/cdcgov/phdi/2995)!

angelathe:
> meta-question: how does this interact with time zones?

@mcmcgrath13 The functions should handle things in user's local TZ. Ex. When I set the date range to "Last 7 days", it will automatically set the time of the start date to midnight local time.

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -256,4 +213,235 @@ const FilterReportableConditions = () => {
...
+  ) => {
+    // Set option
+    const { value } = event.target;
+    setFilterDateOption(value);
+  };
+
+  // Local dev default = Last year ; Prod default = Last 24 hours
+  const isFilterDateDefault =
+    filterDateOption ===
+    (isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear);
```
BobanL, line 251: I read the ticket #2752 and I'm not sure why we need to have a different default for local & production. I think it just adds unnecessary complexity where we can just manually click 24 hours.

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -256,4 +213,235 @@ const FilterReportableConditions = () => {
...
+
+  const resetFilterDate = () => {
+    const queryDateRange = searchParams.get("dateRange");
+    if (!queryDateRange) {
+      setFilterDateOption(
+        isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear,
+      );
+    } else {
+      if (queryDateRange !== filterDateOption) {
```
BobanL, line 260: Cane we combine this into one elseif?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -256,4 +213,235 @@ const FilterReportableConditions = () => {
...
+        <div className="border-top-1px border-base-lighter margin-x-105"></div>
+      </div>
+    );
+  };
+
+  return (
+    <div>
+      <div className="position-relative display-flex flex-column">
+        <Button
+          className={"margin-right-0 filters-applied"}
```
BobanL, line 300: I think we should probably to set a fixed width here so that changing the filter option does not change the width


https://github.com/user-attachments/assets/dc6ac0b1-12e7-4c42-8024-6df6a51cf59e

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -3,18 +3,25 @@
...
  */
-export const Filters = () => {
+const Filters = () => {
   return (
     <div>
       <div className="border-top border-base-lighter"></div>
       <div className="margin-x-3 margin-y-105 display-flex flex-align-center gap-105">
         <span className="line-height-sans-6">FILTERS:</span>
+        <FilterByDate />
         <FilterReportableConditions />
```
BobanL, line 25: Could we move both of these components into their own file? I think it might help with readability of this `Filters.tsx` file

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -256,4 +213,235 @@ const FilterReportableConditions = () => {
...
+    const { value } = event.target;
+    setFilterDateOption(value);
+  };
+
+  // Local dev default = Last year ; Prod default = Last 24 hours
+  const isFilterDateDefault =
+    filterDateOption ===
+    (isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear);
+
+  const resetFilterDate = () => {
```
BobanL, line 253: Would `setFilterDate` be a more appropriate name for this function?

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: @ashton-skylight Do you prefer having both filter boxes be the same width? In the Figma, the `Filter by Received Date` box is skinnier.

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: `resetFilterDate` is only used on first load or when we changed the filter but didn't apply it

containers/ecr-viewer/src/app/components/Filters.tsx
ashton-skylight: @angelathe I think the Received Date filter box dynamically changing width is fine because once the custom date range is in, the filter box width will expand to include the dates.

ashton-skylight: Out of curiosity, are the filter boxes the same height? It's probably just my vision, but the `Received Date' filter box looks vaguely shorter than the `Reportable Condition` filter box

angelathe: @ashton-skylight If it's okay, could I finalize the widths response in the custom date filter ticket? And I'll check on the heights as well!

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Pushed some changes last night that refactored the file into two smaller files (did it slightly differently w/ Mary's abstraction branch)

`<BaseFilter>` - has reusable Filter component, helper funcs/components
`<Filters>` - builds out Filter by Date/Condition, and the whole Filter bar

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: This was my request. Because we don't have eCRs loading all the time in dev environments, more often than not with the last 24 hours filter, there are no ecrs available on load. My preference is a longer window for dev to make sure the data always shows up

mcmcgrath13: (I can very much see myself debugging why there are no eCRs loaded even though I definitely thought I loaded eCRs only to realize out after too much time that it's the date filter causing the issue)

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -0,0 +1,193 @@
...
+  icon: IconTag,
+  tag = "",
+  resetHandler,
+  submitHandler,
+  children,
+}: {
+  btnClass?: string;
+  type: string;
+  title?: string;
+  icon: ComponentType<{ className?: string }>;
```
mcmcgrath13, line 98: should this be `IconProps`? (or did you otherwise solve the build error you were seeing?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -139,120 +107,152 @@ const FilterReportableConditions = () => {
...
+    const queryDateRange = searchParams.get("dateRange");
+    if (!queryDateRange) {
+      setFilterDateOption(
+        isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear,
+      );
```
mcmcgrath13, line 214:
```suggestion
      setFilterDateOption(filterDateOption);
```

containers/ecr-viewer/src/app/page.tsx
```diff
@@ -19,19 +25,28 @@ const HomePage = async ({
...
   const itemsPerPage = Number(searchParams?.itemsPerPage) || 25;
   const sortColumn = (searchParams?.columnId as string) || "date_created";
   const sortDirection = (searchParams?.direction as string) || "DESC";
   const searchTerm = searchParams?.search as string | undefined;
   const filterConditions = searchParams?.condition as string | undefined;
   const filterConditionsArr = filterConditions?.split("|");
+  const filterDateRange = convertDateOptionToDateRange(
+    (searchParams?.dateRange as string) ||
+      (isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear),
+  );
```
mcmcgrath13, line 41: is it worth making a little utility in `date-utils` that returns the default date range? this logic is in quite a few spots

containers/ecr-viewer/src/app/tests/components/Filters.test.tsx
```diff
@@ -266,3 +269,155 @@ describe("Filters Component", () => {
...
+  it("Updates filter date range when selection is made", async () => {
+    render(<Filters />);
+    const toggleButton = screen.getByRole("button", {
+      name: /Filter by Received Date/i,
+    });
+    fireEvent.click(toggleButton);
+
+    // Check default state is "Last year" (local dev default)
+    await waitFor(() => screen.getByRole("radio", { name: "Last year" }));
+    console.log(screen.debug());
```
mcmcgrath13, line 320: delete log?

containers/ecr-viewer/src/app/tests/date-utils.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      DateRangeOptions.Last24Hours,
+      today,
+    );
+    startDate.setSeconds(0, 0);
+
+    const expectedStartDate = new Date(today);
+    expectedStartDate.setSeconds(0, 0);
+    expectedStartDate.setTime(
+      expectedStartDate.getTime() - 24 * 60 * 60 * 1000,
+    );
```
mcmcgrath13, line 20: if today is hard coded, can we hard code the expected date as well? (same for all the below)

containers/ecr-viewer/src/app/components/BaseFilter.tsx
angelathe: Solved it by making className optional!

containers/ecr-viewer/src/app/view-data/utils/date-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+  [DateRangeOptions.Last3Months]: toSentenceCase(
+    DateRangeOptions.Last3Months.replace(/-/g, " "),
+  ) as string,
+  [DateRangeOptions.Last6Months]: toSentenceCase(
+    DateRangeOptions.Last6Months.replace(/-/g, " "),
+  ) as string,
+  [DateRangeOptions.LastYear]: toSentenceCase(
+    DateRangeOptions.LastYear.replace(/-/g, " "),
+  ) as string,
+};
```
mcmcgrath13, line 36:
```suggestion
export const dateRangeLabels = Object.keys(DateRangeOptions).reduce((res, k) => {
  if (typeof DateRangeOptions[k] === 'number') {
    res[k] = k.toLowerCase()
  }
  return res
}, {}
)
```

(needs some more massaging to convince typescript this is fine)

containers/ecr-viewer/src/app/view-data/utils/date-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Calculates the date a specified number of days ago from the given date
+ * @param refDate - The reference date
+ * @param days - The number of days to subtract from the reference date
+ * @param resetHours - If true, resets the time to the start of the day (00:00:00.000) local time
+ * @returns The updated date
+ */
+function daysAgoDate(refDate: Date, days: number, resetHours: boolean) {
```
mcmcgrath13, line 45: I was thinking the `resetHours` would have a default true since that's what's mostly used - what do you think?

Also, `monthsAgoDate` never takes advantage of reset hours - do we need that option?

containers/ecr-viewer/src/app/page.tsx
angelathe: @mcmcgrath13 Thoughts on if this lived in <FilterbyDate>?

containers/ecr-viewer/src/app/view-data/utils/date-utils.ts
angelathe: Thank youuu for this, was struggling with this 😮‍💨

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -0,0 +1,193 @@
...
+ * @param props.title - Title text displayed on the button; defaults to `type`.
+ * @param props.icon - Icon component rendered inside the filter button.
+ * @param props.tag - Optional tag element displayed next to the title.
+ * @param props.resetHandler - Callback for resetting the filter.
+ * @param props.submitHandler - Callback for applying the filter on form submission.
+ * @param props.children - The filter form fields and content displayed in the dropdown.
+ * @returns A JSX element for the filter with a dropdown form.
+ */
+export const Filter = ({
+  btnClass = "filters-applied",
```
mcmcgrath13, line 86: in thinking about the props as an API, I wonder if it might be cleaner to name this `isActive` or similar so it's more functional than explicit. As that seems to be what it's trying to achieve in the conditions case and dates are just always active because of the nature of radio buttons?

containers/ecr-viewer/src/app/view-data/utils/date-utils.ts
angelathe: Uncommitted the change till later, need to figure out why tests are failing w/ this change

containers/ecr-viewer/src/app/page.tsx
mcmcgrath13: as long as it's importable where needed and lives in a place that makes some sense, that works for me

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Don't we need this to set the date range default if there is no `dateRange` param in the query?

I ended up changing this to:
```
    if (!queryDateRange) {
      setFilterDateOption(defaultDateRange);
```

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -0,0 +1,193 @@
+import React, { ComponentType, ReactNode, useCallback, useState } from "react";
+import { Button } from "@trussworks/react-uswds";
+import { useRouter, usePathname, useSearchParams } from "next/navigation";
+
+/**
+ * Custom hook to manage query parameters in the URL (set, delete, and update). Hook always resets page back to 1.
+ * @returns - An object containing
+ *  - searchParams: current search params from the URL and a function to update
```
mcmcgrath13, line 8:
```suggestion
 *  - searchParams: current search params from the URL
```

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -0,0 +1,193 @@
...
+  // Update a specific query param (set or delete if default)
+  const updateQueryParam = (
+    key: string,
+    value: string | { [key: string]: boolean },
+    isDefault: boolean,
+  ) => {
+    let updatedParams;
+
+    if (typeof value === "object") {
+      // Filter by Reportable Condition
```
mcmcgrath13, line 47: instead of the `// Filter by...` comments could you instead add in the docstring that if an object is passed, its keys that are set to true are concatenated with a `|` and otherwise the value is set directly?

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -0,0 +1,193 @@
...
+        .join("|");
+      updatedParams = isDefault
+        ? deleteQueryParam(key)
+        : setQueryParam(key, filteredKeys);
+    } else {
+      // Filter by Date
+      updatedParams = isDefault
+        ? deleteQueryParam(key)
+        : setQueryParam(key, value);
+    }
```
mcmcgrath13, line 59:
```suggestion
    if (typeof value === "object") {
      value = Object.keys(value)
        .filter((k) => value[k] === true)
        .join("|");
    }

    const updatedParams = isDefault
        ? deleteQueryParam(key)
        : setQueryParam(key, value);
```

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  return { searchParams, updateQueryParam };
+};
+
+/**
+ * A reusable Filter component for eCR Library. It displays a button
+ * that toggles a dropdown form for filtering data. The form includes functionality
+ * for resetting and submitting the filter.
+ * @param props - The props for the Filter component.
+ * @param props.isActive - CSS class for the filter button; default is 'filters-applied' // TODO
```
mcmcgrath13, line 75:
```suggestion
 * @param props.isActive - Whether the filter is actively filtering the eCRs
```

containers/ecr-viewer/src/app/page.tsx
```diff
@@ -19,19 +25,28 @@ const HomePage = async ({
...
   const currentPage = Number(searchParams?.page) || 1;
   const itemsPerPage = Number(searchParams?.itemsPerPage) || 25;
   const sortColumn = (searchParams?.columnId as string) || "date_created";
   const sortDirection = (searchParams?.direction as string) || "DESC";
   const searchTerm = searchParams?.search as string | undefined;
   const filterConditions = searchParams?.condition as string | undefined;
   const filterConditionsArr = filterConditions?.split("|");
+  const filterDateRange = convertDateOptionToDateRange(
+    (searchParams?.dateRange as string) ||
+      (isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear),
```
mcmcgrath13, line 40: could the `isProd ? DateRangeOptions.Last24Hours : DateRangeOptions.LastYear` be abstracted out to a `defaultDateRange` function or `DEFAULT_DATE_RANGE` constant in a place that everywhere can import it?

containers/ecr-viewer/src/app/tests/date-utils.test.ts
```diff
@@ -0,0 +1,96 @@
+import {
+  DateRangeOptions,
+  convertDateOptionToDateRange,
+} from "@/app/view-data/utils/date-utils";
+
+describe("convertDateOptionToDateRange", () => {
+  const today = new Date("2024-04-03");
```
mcmcgrath13, line 7:
```suggestion
  const today = new Date("2024-04-03T12:34");
```
to make sure the zeroing works properly (not sure if this date is fully valid, so might need tweaking)

containers/ecr-viewer/src/app/tests/date-utils.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  it("should return last 24 hours range", () => {
+    const expectedStart = new Date("2024-04-02T00:00:00.000Z");
+    const { startDate, endDate } = convertDateOptionToDateRange(
+      DateRangeOptions.Last24Hours,
+      today,
+    );
+    startDate.setSeconds(0, 0);
```
mcmcgrath13, line 15: why is this needed?

containers/ecr-viewer/src/app/tests/date-utils.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    expect(startDate).toEqual(expectedStart);
+    expect(endDate).toEqual(today);
+  });
+
+  it("should return last 7 days range", () => {
+    const expectedStart = new Date("2024-03-26T04:00:00.000Z");
```
mcmcgrath13, line 22: do these tests rely on the user running them in a specific time zone? It looks like the expected times are for eastern time

containers/ecr-viewer/src/app/page.tsx
angelathe: Oops I made a const for this already, forgot to import it here

containers/ecr-viewer/src/app/tests/HomePage.test.tsx
```diff
@@ -38,3 +40,26 @@ describe("Home Page", () => {
     expect(screen.queryByText("Page not found")).not.toBeInTheDocument();
   });
 });
+
+describe("Reading query params on home page", () => {
+  it("should call convertDateOptionToDateRange with the correct dateRange from query params", async () => {
+    process.env.NEXT_PUBLIC_NON_INTEGRATED_VIEWER = "true";
```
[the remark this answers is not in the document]
angelathe: There was already this in the `afterEach`:

```
  afterEach(() => {
    process.env.NEXT_PUBLIC_NON_INTEGRATED_VIEWER = "false";
```

containers/ecr-viewer/src/app/tests/date-utils.test.ts
angelathe: Just noting for review, ended up changing to `2024-04-02T12:34` instead of `04-03` to effectively keep the same date (before, 2024-04-03 midnight UTC was 04-02 8pm in Jest local TZ).

containers/ecr-viewer/src/app/tests/date-utils.test.ts
```diff
@@ -4,15 +4,15 @@ import {
 } from "@/app/view-data/utils/date-utils";
 
 describe("convertDateOptionToDateRange", () => {
-  const today = new Date("2024-04-03");
+  const today = new Date("2024-04-02T12:34");
+  console.log(today);
 
   it("should return last 24 hours range", () => {
-    const expectedStart = new Date("2024-04-02T00:00:00.000Z");
+    const expectedStart = new Date("2024-04-01T16:34:00.000Z");
```
mcmcgrath13, line 11:
```suggestion
    const expectedStart = new Date("2024-04-01T16:34:01.009Z");
```
for extra certainty

angelathe: Done!

mcmcgrath13 (approved):
✨

PR #3078, opened 2024-12-18
refactor: moar abstraction

mcmcgrath13:


containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -413,34 +327,109 @@ const useQueryParam = () => {
...
+          <div className="usa-combo-box top-full left-0">
+            <form
+              onSubmit={(e) => {
+                e.preventDefault();
+                submitHandler();
+                setIsFilterBoxOpen(false);
+                openBtnRef?.current?.parentElement?.focus();
+              }}
+            >
+              <fieldset className="usa-combo-box border-0 padding-0 margin-top-1 bg-white position-absolute radius-md shadow-2 z-top maxh-6205 width-4305">
```
mcmcgrath13, line 392: the width here was different between the two - something to think about for if that's necessary or there's one width or maybe a better way to compute the width

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -413,34 +327,109 @@ const useQueryParam = () => {
...
+          type="button"
+        >
+          <span ref={openBtnRef} className="square-205 usa-icon">
+            <IconTag aria-hidden className="square-205" />
+          </span>
+          <span className="text-ink">{title || type}</span>
+          {tag && (
+            <span
+              className="usa-tag padding-05 bg-base-darker radius-md"
+              data-testid="filter-conditions-tag"
```
mcmcgrath13, line 375: maybe drop the conditions here and just make it `filter-tag`

angelathe (commented):
(Oops, thought I pressed submit) -- thank you thank you for this! Wrote a couple comments on some small changes I made in the 2nd commit. Still need to think about the different combo box widths.

angelathe (approved):

PR #3084, opened 2024-12-23
feat: only allow one filter open at a time, easy closing

mcmcgrath13:
# PULL REQUEST

## Summary

## Related Issue
Fixes #

## Acceptance Criteria

## Additional Information

## Checklist


containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -94,21 +95,32 @@ export const Filter = ({
   submitHandler: () => void;
   children: ReactNode;
 }) => {
-  const [isFilterBoxOpen, setIsFilterBoxOpen] = useState(false);
+  const { filterBoxOpen, setFilterBoxOpen, openButtonRef } =
+    React.useContext(FilterOpenContext);
   const openBtnRef = React.useRef<HTMLElement | null>(null);
 
+  const isFilterBoxOpen = filterBoxOpen == type;
```
mcmcgrath13, line 102:
```suggestion
  const isFilterBoxOpen = filterBoxOpen === type;
```

angelathe:
Quick thing I noticed when testing locally: filters should reset to their previous state if we change it but then don't apply & open a different filter. 

Ex. If I change something in filter A and then switch over to filter B before applying A, filter A doesn't visually reset until opening/closing A again

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -133,7 +153,7 @@ export const Filter = ({
               onSubmit={(e) => {
                 e.preventDefault();
                 submitHandler();
-                setIsFilterBoxOpen(false);
+                setFilterBoxOpen("__submitted__");
```
angelathe, line 156: Why do we want it to set filterBoxOpen to `"__submitted__"`? It's slightly unclear how filterBoxOpen gets used when it can equal either filter type or `"__submitted__"`

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -9,19 +9,70 @@ import {
...
+
+        // Return focus to the most recently selected open button
+        lastOpenButtonRef.current?.focus();
+        lastOpenButtonRef.current = null;
+      }
+    };
+    if (filterBoxOpen) {
+      window.addEventListener("keydown", handleEscapeFilters);
+      window.addEventListener("click", resetFilters);
+      return () => {
```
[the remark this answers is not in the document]
mcmcgrath13: the cleanup happens whenever `filterBoxOpen` changes

So if we start with `filterBoxOpen === false`, then the listeners don't get added and nothing returned. When it changes to `true`, there was no cleanup returned last time, so nothing happens there, then we add the event listeners and return the cleanup functions. When it flips back to `false`, the cleanup functions we returned are run, then we are back in the original state.

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -9,19 +9,70 @@ import {
...
+    lastOpenButtonRef,
+  };
+
+  const resetFilters = React.useCallback(() => {
+    setFilterBoxOpen(null);
+  }, []);
+
+  // When a filter is open, close it if the escape key is hit or a click happens
+  // outside the <Filter /> component (implemented by stopping click propogation on <Filter />)
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
mcmcgrath13: updated this and all the `useBlahs` to use imports instead of prefixes

containers/ecr-viewer/src/app/components/BaseFilter.tsx
mcmcgrath13: There's a tricky race condition this helps us avoid. There's a comment on line 121 about it. But the problem boils down to the resetting relying on the params being correct and on submit the params don't update until the next tic, so if we go the `null` state and reset, then we reset to the old state and not the new. It feels hacky, but I haven't thought of a better way yet 🤔

mcmcgrath13: I refactored a little to hopefully maybe improve the readability a bit, but it's still fundamentally the same approach 😞

angelathe (approved):

PR #3090, opened 2024-12-26
chore: replace pretty-format-json with prettier functionality

BobanL:
# PULL REQUEST

## Summary
- Replace pretty-format-json with prettier functionality
- Remove auto update from pre-commit because eslint and prettier are at the latest supported versions
- Remove terraform formatting because we don't have any tf in the project
- pin eslint and prettier to the latest supported versions.
  - Nextjs 14 does not support eslint 9
  - Pre-commit doesn't play nice with prettier 4+


.pre-commit-config.yaml
```diff
@@ -31,29 +26,17 @@ repos:
...
-        entry: bash -c 'if command -v terraform >/dev/null 2>&1; then terraform fmt -recursive -check operations; else exit 0; fi'
-        language: system
-        types:
-          - terraform
-        pass_filenames: false
 ci:
   autofix_commit_msg: |
     [pre-commit.ci] auto fixes from pre-commit hooks
   autofix_prs: true
-  autoupdate_branch: ""
```
JNygaard-Skylight, line 54: Is there a way to keep the autoupdate on, but have it ignore the pre-commit hooks we want to lock?

JNygaard-Skylight (changes requested):
This looks good to me, just a small question on whether we can keep the autoupdate, and have it ignore the two hooks we want to update. If not that's fine by me.

.pre-commit-config.yaml
BobanL: I actually don't think there is a way to disable auto update... The defaults are to always update once a week. 
https://pre-commit.ci/#configuration-autoupdate_branch.

I'll just put it back in.

JNygaard-Skylight (approved):
LGTM!

PR #3093, opened 2025-01-02
Correct FHIR produced by TCRS

### LINKED ISSUE #3076: Trigger Code Reference Service creates Invalid FHIR
## Description

When the TCR "stamps" relevant observations, encounters, conditions, etc. by adding the `"https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code"` extension,, the extension object does not conform to the [FHIR specifications.](https://www.hl7.org/fhir/R4/extensibility.html) In particular read the notes regarding the "value[x] element".  The extension added by the TCR in the `add_code_extension_and_human_readable_name() method in `utils.py`takes the form:
```
{
    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code",
    "coding": [{"code": code, "system": "http://snomed.info/sct"}],
}
```
Whereas it should be:
```
{
    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code",
    "valueCoding": [{"code": code, "system": "http://snomed.info/sct"}],
}
```
In addition to correcting the Python code for the TCR, the `ecrSummaryService.tsx` will need to be updated. In that file an interface is created to account for the inncorrect FHIR:
```
interface ExtensionConditionCode extends Extension {
  coding?: { code: string; system: string }[];
}
```
but with this change we can stick to the types in the `fhir/r4` library.

## Steps to reproduce

1.  in `containers/ecr-viewer` run `npm run convert-seed-data` to get the most up-t0-date versions of the converted seed data.
2.  In any of the newly created `bundle.json` files search for an `https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code` extension.

## Expected Result

The extension looks like:
```
{
    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code",
    "coding": [
        {
            "code": "409498004",
            "system": "http://snomed.info/sct"
        }
    ]
}
```

## Actual result

It should be:
```
{
    "url": "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code",
    "valueCoding": [
        {
            "code": "409498004",
            "system": "http://snomed.info/sct"
        }
    ]
}
```

## Priority

#### Priority

## Notes

JNygaard-Skylight:
# PULL REQUEST

## Summary
The TCRS is currently not producing valid FHIR when it "stamps" a resource with the snomed code of the reportable condition. This corrects that.

## Related Issue
Fixes #3076 

## Acceptance Criteria

## Additional Information

## Checklist


containers/ecr-viewer/src/app/services/ecrSummaryService.tsx
```diff
@@ -322,22 +299,19 @@ export const evaluateEcrSummaryRelevantLabResults = (
...
-    ) as ObservationStamped[]
+    evaluate(fhirBundle, fhirPathMappings["observations"]) as Observation[]
   )
     .filter((entry) =>
       entry.extension?.some(
         (ext) =>
           ext.url ===
             "https://reportstream.cdc.gov/fhir/StructureDefinition/condition-code" &&
-          ext.coding?.some((item) => item.code === snomedCode),
+          ext.valueCoding?.code === snomedCode,
```
mcmcgrath13, line 314: A slightly philosophical question - How are we thinking about backwards compatibility with changes like this? Presumably we don't expect old bundles to be updated with this fix, so in the wild, there would be fhir bundles with the old buggy version and the correct version. Should we support both? Or given the current maturity of the product/usage, we're comfortable with this potential breakage? (cc @austin-hall-skylight @emmastephenson )

JNygaard-Skylight: My 2¢: The possibility of having to support all broken versions of FHIR that we currently produce (and eventually produce) feels me with dread.  It would be nice if, as part of the ecr-viewer update process, it reruns the conversion. Perhaps we could add metadata to bundles to store which version of the ecr-viewer created it?

mcmcgrath13: tracking the version that created it seems wise, then we could add a batch job that goes through and updates the bundles if they are out of date (perhaps with some filters on other metadata like date as maybe for very old eCRs people just won't care)

emmastephenson: Agreed that tracking the versions seem like a good idea. I've also thought it'd be nice to have some kind of automatic FHIR re-conversion for the eCRs (maybe when someone tries to look at one, so we're not doing unnecessary conversions) but I always figured that would require the original XMLs, and for now we aren't storing those. Could be a feature to look into if team 4 ends up adding eCR XML storage to the application.

BobanL (approved):

PR #3095, opened 2025-01-02
fix: unique bundle cache key

mcmcgrath13:
# PULL REQUEST

## Summary
In the evaluation cache, instead of using `"bundle"` as the key for the `Bundle` resource, use the `fullUrl` of the first entry (if it exists).

Currently, if you are at an eCR, go back to the library, and click a new eCR the data still reflects the old eCR. I do see the network traffic successfully return the bundle that should display, but it doesn't update the page

## Additional Information
We really need integration tests to test this properly I think


BobanL (approved):
👍

PR #3097, opened 2025-01-03
testing: enforce no error or warning logs

### LINKED ISSUE #3111: Clean up error logs in tests
Don't allow error or warning console logs in tests. Clean up tests currently emitting logs.

mcmcgrath13:
# PULL REQUEST

## Summary
Don't allow console errors or warnings on tests. These can build up and tests will still pass with them on, but it seems generally bad that they exist. This PR proposes enforcing that they are handled (either in the code or test explicitly).

It also cleans up tests causing these logs.

fixes #3111


containers/ecr-viewer/src/app/tests/components/EcrTable.test.tsx
```diff
@@ -66,7 +66,9 @@ describe("EcrTable", () => {
         filterDate: mockDateRange,
       }),
     );
-    expect(await axe(container)).toHaveNoViolations();
+    await act(async () => {
+      expect(await axe(container)).toHaveNoViolations();
+    });
```
BobanL, line 71: This seems likely to be caused by [Next/Link](https://www.benmvp.com/blog/avoiding-react-act-warning-when-accessibility-testing-next-link-jest-axe/).

BobanL (approved):
Thanks for finding this jest package!

containers/ecr-viewer/src/app/tests/components/EcrTable.test.tsx
mcmcgrath13: yeah. There's an intersection observer in there that causes some re-rendering

PR #3098, opened 2025-01-03
fix: always 404 on unknown eCR id

### LINKED ISSUE #2962: 500 gets returned to user when data is not found in azure or aws
## Description

- When something isn't found in aws blob storage or s3 a 500 internal server error is returned to the user
## Steps to reproduce

_List out the steps in detail to reproduce the bug_

1. Run with azure/s3
2. View any non-existent id

## Expected Result

- 404 screen should appear

## Actual result

- 500 error is displayed
image
image

## Priority

#### Priority

## Notes

mcmcgrath13:
# PULL REQUEST

## Summary
Handle 404 paths for getting fhir bundle from s3 and azure

## Related Issue
Fixes #2962


containers/ecr-viewer/seed-scripts/create-seed-data.py
```diff
@@ -30,7 +30,7 @@ def _process_files(args):
     :return: A list of fhir bundles
     """
     print("Converting files...")
-    subfolders = ["LA"]
+    subfolders = ["ME"]
```
mcmcgrath13, line 33:
```suggestion
    subfolders = ["LA"]
```

BobanL (approved):

PR #3112, opened 2025-01-08
Look for human-readable names for conditions

JNygaard-Skylight:
# PULL REQUEST

## Summary

The message-parser config the seed script uses is not getting human readable names.

## Related Issue

Fixes # n/a

## Acceptance Criteria

## Additional Information

## Checklist


containers/message-parser/app/default_schemas/extended.json
```diff
@@ -349,7 +349,7 @@
         }
       },
       "condition": {
-        "fhir_path": "Observation.valueCodeableConcept.coding.display",
+        "fhir_path": "iif(Observation.valueCodeableConcept.text.exists(), Observation.valueCodeableConcept.text, Observation.valueCodeableConcept.coding.display)",
```
mcmcgrath13, line 352: does the display need a `first` since there can be multiple?

containers/trigger-code-reference/app/utils.py
```diff
@@ -205,53 +205,89 @@ def _find_codes_by_resource_type(resource: dict) -> List[str]:
...
-            and x["code"] == "64572001"  # Condition
-        ]:
-            condition_name = _get_condition_name_from_snomed_code(code)
-            if condition_name:
-                resource["valueCodeableConcept"]["text"] = condition_name
+        ),
+        None,
+    )
+
+    if condition_code:
```
mcmcgrath13, line 259:
```suggestion
    if condition_code is not None:
```

containers/trigger-code-reference/app/utils.py
```diff
@@ -205,53 +205,89 @@ def _find_codes_by_resource_type(resource: dict) -> List[str]:
...
+        human_readable_condition_name = _get_condition_name_from_snomed_code(
+            condition_code["code"]
+        )
+
+        if human_readable_condition_name:
+            resource["valueCodeableConcept"]["text"] = human_readable_condition_name
+        elif "display" in condition_code:
+            resource["valueCodeableConcept"]["text"] = condition_code["display"]
```
mcmcgrath13, line 267: what's the thinking around including the `display` as a fallback? the display would be in the resulting fhir bundle either way, right? so it's not stamping new/outside info

containers/trigger-code-reference/seed-scripts/seed_rckms_database.py
```diff
@@ -3,174 +3,32 @@
 seed_rckms_database.py
 ===========
 
-A significantly paired down version of the original script to seed the RCKMS database. This script will extract the conditions from the RCKMS docx files and load them into the SQLite database.
-
-
-Example:
-    $ python seed_rckms_database.py <archive.zip> <database.db>
+A significantly paired down version of the original script to seed the RCKMS database. This script will extract the conditions from the RCKMS CSV in assets (downloaded from https://www.rckms.org/content-repository) files and load them into the SQLite database.
```
mcmcgrath13, line 6:
```suggestion
A significantly pared down version of the original script to seed the RCKMS database. This script will extract the conditions from the RCKMS CSV in assets (downloaded from https://www.rckms.org/content-repository) files and load them into the SQLite database.
```

containers/trigger-code-reference/seed-scripts/seed_rckms_database.py
```diff
@@ -3,174 +3,32 @@
...
-        for condition in conditions:
-            # insert the Condition into the database if it does not exist
+        next(reader)
+        for row in tqdm(reader, desc="Loading conditions", unit=" conditions"):
             cur.execute(
-                "INSERT INTO conditions (id, system, name) "
-                "VALUES (:snomed, 'http://snomed.info/sct', :name) "
-                "ON CONFLICT DO NOTHING",
-                condition.__dict__,
+                "INSERT INTO conditions (id, system, name, description) VALUES (?, ?, ?, ?)",
```
mcmcgrath13, line 31: do we ever get conflicts in practice?

containers/message-parser/app/default_schemas/extended.json
JNygaard-Skylight: No, `evaluate` always returns a list, plus the logic to pick a `display` is easier in TS.

containers/trigger-code-reference/app/utils.py
JNygaard-Skylight: I don't know why we need to test specifically for `is not None` instead of is `condition_code` truthy.

containers/trigger-code-reference/app/utils.py
JNygaard-Skylight: I was thinking that it might make the frontend logic easier if there was always a `"text"` even if it is not the preferred human-readable name. But perhaps it might be confusing if we expect `"text"` to be the preferred name and instead it's just the same as the `"display"`?

containers/trigger-code-reference/seed-scripts/seed_rckms_database.py
JNygaard-Skylight: I don't think there is any reason why we should. You can check the CSV: `containers/trigger-code-reference/assets/RCKMS Condition Codes.csv` and if we did I'm okay with it blowing up.

PR #3114, opened 2025-01-08
Use session storage to persist filter and search

### LINKED ISSUE #3063: Filters and search should persist when re-loading the Library
## What needs to be done

Update the user experience for filtering to persist through reloads, such that:
GIVEN I've -- 1) filtered to a condition and/or date in the Library AND/OR 2) entered a search term -- and then clicked into a specific eCR
WHEN I go back to the Library page
THEN my condition and/or date filters and/or search term should still be applied.

## Why it needs to be done

Currently, when you click into a specific eCR and then go back to the Library, your filters aren't persisted. This is an annoying user experience since you now have to re-apply your filters. 

Note: local session should save:
1. Conditions filtered on
2. Date range option / custom date filtered on
3. Search term

## Design review
(This behavior will need to be validated with the design team)

## Acceptance Criteria

- [ ] When a user has filters applied and navigates into a single eCR, the filters and/or search should persist on return

## Technical Notes

~Not entirely sure how this will need to be done - cookies? Or just an update to the way the Library button functions?~
Update: should be able to use session info for this.

gordonfarrell:
# PULL REQUEST

## Summary

Saves search param info into session storage and uses it to build the href for the link back to the eCR Library from  the view page.

## Related Issue

Fixes #3063

## Acceptance Criteria

When a user has filters applied and navigates into a single eCR, the filters and/or search should persist on return


containers/ecr-viewer/src/app/components/EcrTableClient.tsx
```diff
@@ -298,11 +301,14 @@ const DataRow = ({ item }: { item: EcrDisplay }) => {
...
   );
+  const saveUrl = () => {
+    saveToSessionStorage("urlParams", searchParams.toString());
+  };
 
   return (
     <tr>
       <td>
-        <Link href={`/view-data?id=${item.ecrId}`}>
+        <Link onClick={saveUrl} href={`/view-data?id=${item.ecrId}`}>
```
mcmcgrath13, line 311: so simple! :heart:

mcmcgrath13 (approved):
🧨

PR #3117, opened 2025-01-09
Filter by Date Pt. 2 - Custom date selection

### LINKED ISSUE #2753: Copy of FILTER: Filter By Date (Part 1 - Default selections)
WIP

## User Story

As a user of the eCR Library, I want to filter by date, so that I can only see eCRs from within a given date range.

Example: I only want to see eCRs for the last week, so I can identify issues.

## Acceptance Criteria

- [ ] "Filter By Received Date" implemented for:
    - [ ] Last 24 hours
    - [ ] Last 7 days
    - [ ] Last 30 days
    - [ ] Last 3 months 
    - [ ] Last 6 months
    - [ ] Last year
- [ ] When user applies filter for any of the above, only eCRs received in that date range are included
- [ ] My selection is shown as the button text (e.g. "Last 7 days")
- [ ] By default, only eCRs from the last 24h are shown

## Scenarios

_Scenario 1_
- I click on the date filter button
- A menu appears with selectable presets
- I choose "Last 7 days"
- I click "Apply Filter"
- Only eCRs from the last week are shown

## Design Notes

[Link to Figma prototype](https://www.figma.com/proto/zAPyOEYwNtx3f2Y47aoq0G/eCR-Viewer-(Aug-Dec-2024)?node-id=4806-39714&node-type=canvas&m=dev&scaling=min-zoom&content-scaling=fixed&page-id=4032%3A9311&starting-point-node-id=4806%3A39674)

## Additional context

This ticket does NOT include implementing or showing the custom date filter - that is a separate ticket.

This ticket includes both the frontend design implementation and the backend querying needed for full functionality.

angelathe:
# PULL REQUEST

## Summary

- Add custom date selection options when filtering by date
    - Max date set to `today`
    - Start date time set to `00:00:00.000`, and end date time set to `23:59:59.999`
- Add/update unit and snapshot tests

## Related Issue

Fixes #2753 

## Additional Information

**NOTE:** Still looking to clean some things up (left a few comments), but would love to start the feedback process now.

**Things that will probably get broken out into other tickets:**
- `Apply Filter` button disabled/enabled states - button should 
- ~Adding a custom error tooltip for when `startDate > endDate`. Right now, there's a placeholder `Alert` for this error~
- URL validation and syncing

## Acceptance Criteria

- [ ] "Filter By Received Date" implemented for custom date range
- [ ] When a user chooses "Custom date range", they must populate the start and end dates
- [ ] When user applies the custom filter, only eCRs received in that date range are included
- [ ] My selection is shown as the button text (e.g. "From 10/7/2024 to 10/10/2024")

## Checklist


containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -35,10 +35,12 @@ export const useQueryParam = () => {
...
   const deleteQueryParam = useCallback(
-    (name: string, value?: string) => {
-      const params = new URLSearchParams(searchParams.toString());
+    (params: URLSearchParams, name: string) => {
+      // const params = new URLSearchParams(searchParams.toString());
       params.set("page", "1");
-      params.delete(name, value);
+      params.delete(name);
+
+      router.push(pathname + "?" + params.toString());
```
angelathe, line 43: Still trying to refactor the delete/update functions

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -292,46 +301,120 @@ const FilterByDate = () => {
...
+      const [startDate, endDate] = queryCustomDates?.split("|");
+      setStartDate(startDate);
+      setEndDate(endDate);
+    }
   };
 
   // on mount, make sure the filters match the search params
   useEffect(resetFilterDate, []);
 
+  const RadioDateOption = ({
```
angelathe, line 315: Thinking this could potentially go in `BaseFilter`?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -292,46 +301,120 @@ const FilterByDate = () => {
...
+    const params = new URLSearchParams(searchParams.toString());
+    if (filterDateOption === "custom") {
+      // TODO ANGELA: Temp: Placeholder errorhandling for when Start Date > End Date
+      if (new Date(startDate) > new Date(endDate)) {
+        alert(
+          "Invalid date range. Start date must be earlier or equal to the end date.",
+        );
+        resetFilterDate();
+        return;
+      }
```
angelathe, line 358: Note: THis is a placeholder alert to safeguard against error where `Start date > end date` -- still working on how to make the custom error tooltip for this / might get brought out into a separate small ticket

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -77,7 +88,7 @@ const Filters = () => {
     }
   }, [filterBoxOpen]);
 
-  const paramKeys = ["condition", "dateRange"];
+  const paramKeys = [ParamName.Condition, ParamName.DateRange, ParamName.Dates];
```
mcmcgrath13, line 91: can this instead be computed from the enum? (is a formal enum worth it or just use a plain object?)

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -124,7 +135,8 @@ const Filters = () => {
  * - Updates the browser's query string when the filter is applied.
  */
 const FilterReportableConditions = () => {
-  const { searchParams, updateQueryParam } = useQueryParam();
+  const { searchParams, updateQueryParam, pushQueryUpdate } = useQueryParam();
+  const params = new URLSearchParams(searchParams.toString());
```
mcmcgrath13, line 139: why do we need to do this copy?

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: does seem like it could move there - maybe just abstract out the `filter-date` part to a param?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -286,52 +308,135 @@ const FilterByDate = () => {
...
+        >
+          {label}
+        </label>
+      </div>
+    );
+  };
+
+  const submitHandler = () => {
+    const params = new URLSearchParams(searchParams.toString());
+    if (filterDateOption === "custom") {
```
mcmcgrath13, line 363: what were the pros/cons/thinking on the two param approach vs a single param like `custom(start,end)`?

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Needed to pass params to updateQueryParam in the `submitHandler` (can't pass searchParams which is read-only)

Could move Line 139 into the `submitHandler` so it's more clear why that's needed?

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Just made an update!

(Maybe an enum is overkill, but thought it would make things clearer in case we expand the scope of filters?)

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -221,3 +218,70 @@ const ApplyFilterButton = ({ type }: { type: string }) => {
...
+          End date
+        </Label>
+        <input
+          key="2"
+          id="end-date"
+          data-testid="end-date"
+          type="date"
+          className="usa-input width-card margin-top-0 border-base-dark custom-date"
+          defaultValue={endDate || ""}
+          required
```
mcmcgrath13, line 276: should we require an end date? or should it be implicitly today if not selected?

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Done!

containers/ecr-viewer/src/app/components/BaseFilter.tsx
angelathe: Hm, I think this is a question for @ashton-skylight.

Do you prefer that 1) end date is always required, i.e. folks would have to manually set today as the end date, or 2) end date is not required, i.e. if users don't fill in the end date, the end date would implicitly be set as today?

angelathe: Ok! Just making a note this slightly changes the acceptance criteria / scenario of the [original ticket](https://github.com/CDCgov/phdi/issues/2753)

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: Sorry, missed this comment on Friday! My thoughts are below, and I'm definitely open to changing the approach.

I went with two URL parameters to keep things simpler. One shows that it’s a custom date, and the other holds the actual dates. Keeping these definitions separate means we don’t have to deal with extra parsing every time we just need one of them (felt cleaner and easier to work with this approach). Also, having two separate parameters helps keep the radio button option name and the URL param values in sync, which makes it simpler to match what the user picks to what’s in the query string.

Also, keeping custom dates in their own parameter seemed like it would make it easier to extend things in the future—like if we wanted to add time ranges. Saves us from having to dig through and parse a single dateRange parameter if we need to add more customizations down the line.

The main con with maintaining the two parameters is maintaining consistency between `dateRange` and `dates`, i.e. making sure that `dates` only exists and is processed when `dateRange=custom`, and that `dates` does not exist when `dateRange` is set to one of the other predefined date ranges. Using the single `dateRange` parameter also to house the custom dates would eliminate this risk.

Thoughts?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -286,52 +309,108 @@ const FilterByDate = () => {
   };
 
   const resetFilterDate = () => {
-    const queryDateRange = searchParams.get("dateRange");
+    const queryDateRange = searchParams.get(ParamName.DateRange);
```
mcmcgrath13, line 312: if a user manually deletes the dates param, it ends up looking like this

image

should we reset to the default instead at this point? cc @ashton-skylight

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -286,52 +309,108 @@ const FilterByDate = () => {
...
+        params,
+        ParamName.DateRange,
+        filterDateOption,
+        isFilterDateDefault,
+      );
+      updatedParams = updateQueryParam(
+        updatedParams,
+        ParamName.Dates,
+        datesParam,
+        false,
```
mcmcgrath13, line 352: maybe add `false` as the default on the function signature so we can skip this here?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -286,52 +309,108 @@ const FilterByDate = () => {
...
+      if (new Date(startDate) > new Date(endDate)) {
+        alert(
+          "Invalid date range. Start date must be earlier or equal to the end date.",
+        );
+        resetFilterDate();
+        return;
+      }
+      const datesParam = `${startDate}|${endDate}`;
+      let updatedParams: URLSearchParams;
+      updatedParams = updateQueryParam(
```
mcmcgrath13, line 342:
```suggestion
      let updatedParams = updateQueryParam(
```

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -221,3 +218,120 @@ const ApplyFilterButton = ({ type }: { type: string }) => {
...
+}: {
+  setStartDate: (date: string) => void;
+  setEndDate: (date: string) => void;
+  startDate: string;
+  endDate: string;
+}) => {
+  return (
+    <div className="padding-x-105">
+      <div>
+        <Label htmlFor="start-date" className="margin-top-1">
```
mcmcgrath13, line 295: is it worth abstracting out these date inputs? there's a lot of overlap between the two

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -286,52 +309,108 @@ const FilterByDate = () => {
...
+            : ""
+          : dateRangeLabels[filterDateOption as DateRangeOptions] || ""
       }
+      submitHandler={submitHandler}
     >
       <div className="display-flex flex-column">
         {Object.values(DateRangeOptions).map((option) => (
-          <div
-            className="checkbox-color usa-radio padding-bottom-1 padding-x-105"
+          <RadioDateOption
```
mcmcgrath13, line 387: maybe move this abstraction up one level to `RadioDateOptions`? there's still a bit of "guts" in here that could have a nicer API if it handles the options bit. Then it could just pass `what are all the options` `handler for option change` and `current option`, which is pretty universal to any set of radio buttons

containers/ecr-viewer/src/app/page.tsx
```diff
@@ -68,7 +80,7 @@ const HomePage = async ({
               sortDirection={sortDirection}
               searchTerm={searchTerm}
               filterConditions={filterConditionsArr}
-              filterDate={filterDateRange}
+              filterDate={filterDates}
```
mcmcgrath13, line 83: minor nit - pluralization doesn't match

containers/ecr-viewer/src/app/page.tsx
```diff
@@ -34,14 +35,25 @@ const HomePage = async ({
...
   const filterConditions = searchParams?.condition as string | undefined;
   const filterConditionsArr = filterConditions?.split("|");
-  const filterDateRange = convertDateOptionToDateRange(
-    (searchParams?.dateRange as string) || DEFAULT_DATE_RANGE,
-  );
+  const filterDateRange =
+    (searchParams?.dateRange as string) || DEFAULT_DATE_RANGE;
+  let filterDates;
+
+  if (filterDateRange === "custom") {
```
mcmcgrath13, line 42: this logic feels a bit out of place here as it's likely something we'll want to sync with the component itself. Can we think about making this a bit more centralized then using that logic both here and in the UI component?

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: That matches what I see as well. We just need to make sure we've got robust handling of the keeping in sync and edge cases there

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: @mcmcgrath13 Made these changes in the last commit if you want to take a look -- but I'm wondering if it's really necessary for this ticket? 

I think I still need to keep `RadioDateOption` because of the lone custom date radio option, but then adding `RadioDateOptions` feels redundant since it's essentially just passing the same parameters down to `RadioDateOption`. (it's just adding an extra layer....?)

My preference would be to rollback that commit, but lmk if you have any thoughts on changes I could make so `RadioDateOptions` actually simplifies the logic.

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -221,3 +218,144 @@ const ApplyFilterButton = ({ type }: { type: string }) => {
...
+  option: string;
+  label: string;
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  isChecked: boolean;
+  classNames?: string;
+}) => {
+  return (
+    <div
+      className={`checkbox-color usa-radio padding-x-105 ${classNames}`}
+      key={`${groupName}-${option}`}
```
mcmcgrath13, line 251: `key` isn't needed here
```suggestion
```

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -221,3 +218,144 @@ const ApplyFilterButton = ({ type }: { type: string }) => {
...
+  isRequired: boolean;
+}) => {
+  const today = new Date().toISOString().split("T")[0];
+  return (
+    <div>
+      <Label htmlFor={id} className="margin-top-1">
+        {label}
+      </Label>
+      <input
+        key={id}
```
mcmcgrath13, line 345: shouldn't need a key here
```suggestion
```

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -221,3 +218,144 @@ const ApplyFilterButton = ({ type }: { type: string }) => {
...
+ */
+export const CustomDateInput = ({
+  label,
+  id,
+  onDateChange,
+  defaultValue,
+  isRequired,
+}: {
+  label: string;
+  id: string;
```
mcmcgrath13, line 333: it seems like this ID is only really used internally to this component - do we need it to be passed? or could it be generated as needed?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -271,10 +292,13 @@ const FilterReportableConditions = () => {
...
 const FilterByDate = () => {
-  const { searchParams, updateQueryParam } = useQueryParam();
-
+  const today = new Date().toISOString().split("T")[0];
+  const { searchParams, deleteQueryParam, updateQueryParam, pushQueryUpdate } =
+    useQueryParam();
   const [filterDateOption, setFilterDateOption] =
     useState<string>(DEFAULT_DATE_RANGE);
+  const [startDate, setStartDate] = useState<string>("");
+  const [endDate, setEndDate] = useState<string>(today);
```
mcmcgrath13, line 301: do we want `today` to be a literal default? or implicit? since "today" changes as a link ages would we expect that to always be the today of the time the filter was completed, or more of a moving today if it was unset?

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -286,52 +310,118 @@ const FilterByDate = () => {
...
+          option="custom"
+          label="Custom date range"
+          onChange={handleDateOptionChange}
+          isChecked={filterDateOption === "custom"}
+        />
+        {filterDateOption === "custom" && (
+          <div className="padding-x-105">
+            <CustomDateInput
+              label="Start date"
+              id="start-date"
```
mcmcgrath13, line 411: do we want to kebab-case-ify `id` from `label` inside the component?

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: I think you're right -- `today` wouldn't automatically update as an actual date changes, it would need a page reload. Changing to make this an implicit default

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: I think it can be even simpler - PR: https://github.com/CDCgov/phdi/pull/3127. And I do think it's worth keeping as it's pretty likely reusable and it keeps the date filter component more readable/tractable

containers/ecr-viewer/src/app/components/Filters.tsx
angelathe: How would this solve the multiple updates in a row issue? Before, this used to be in `updateQueryParams, but this did not allow for us to make multiple updates in `FilterByDate`-- this got fixed by being able to feed the params to `updateQueryParams` in the `submitHandler`.

angelathe: Maybe I could make `params` an optional parameter in updateQueryParams?

containers/ecr-viewer/src/app/page.tsx
```diff
@@ -34,14 +35,25 @@ const HomePage = async ({
...
+    const datesParam = searchParams?.dates as string | undefined;
+    if (datesParam) {
+      filterDates = buildCustomDateRange(datesParam);
```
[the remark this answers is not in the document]
angelathe: @ashton-skylight If someone manually changes dates in the URL and it is bad content, what kind of error would you like to see? Would you want it to be a similar page error like so?


Screenshot 2025-01-14 at 10 25 40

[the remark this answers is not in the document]
ashton-skylight: @angelathe Would the eCR viewer be intelligent enough to decipher that the dates are bad content in the URL or would it only be able to bring up the generic 404?

containers/ecr-viewer/src/app/components/Filters.tsx
ashton-skylight: @mcmcgrath13 Yes, if the user deletes the dates, it should default to the Last 24 hrs.

containers/ecr-viewer/src/app/page.tsx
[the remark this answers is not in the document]
angelathe: @ashton-skylight I thinkkkk we could make it smart enough to know that the dates are bad content. If so, would you want an error page like above?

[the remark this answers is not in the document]
angelathe: @ashton-skylight We could also take the same action as when a user deletes the `dates=` param and have it default back to `Last 24 hours`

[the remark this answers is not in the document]
ashton-skylight: @angelathe Ooo yes! Ideally, would prefer if the time filter defaulted to the `Last 24 hours` instead of the error page

containers/ecr-viewer/src/app/components/Filters.tsx
mcmcgrath13: it doesn't so much solve it, but it makes it clearer the function's delineation. Right now mutates params, but also returns the params. Instead of mutating, it could do this copy step as line one, then return the new params.

containers/ecr-viewer/src/app/components/BaseFilter.tsx
```diff
@@ -221,3 +223,135 @@ const ApplyFilterButton = ({ type }: { type: string }) => {
...
+      <Label htmlFor={id} className="margin-top-1">
+        {label}
+      </Label>
+      <input
+        id={id}
+        data-testid={id}
+        type="date"
+        className="usa-input width-card margin-top-0 border-base-dark"
+        defaultValue={defaultValue}
+        max={today}
```
BobanL, line 347: Maybe if this is ready you don't have to make the change in this PR, but it would be nice to have the End Date have a minimum of start date. 

It seems like that would allow the browser to handle some of the validation for us. 
* When start date is set
image
* When end date is set first OR when it is manually entered
image

BobanL: Maybe this could replace the alert all together.

angelathe:
@mcmcgrath13 I pushed a changes, but a couple things:
1. Currently getting a console warning that's making me hesitant about the current implementation: **Cannot update a component (`Router`) while rendering a different component (`FilterByDate`)`**
2. Seems like URL validation is an overall issue we have to address (also for non-custom date ranges and filter by conditions) -> therefore could be its own ticket (worried that the scope of this ticket keeps growing)
3. I tried for a while to make `returnParamDates` be useful in both `page` and in `resetFilterDate` in the UI component, but it really felt like it wasn't useful for the UI component because A) For UI component, we only want returned dates if the query is "custom", but B) for the page, we always want dates returned. If you have thoughts about re-restructuring this

containers/ecr-viewer/src/app/components/BaseFilter.tsx
angelathe: Ooh this might fix my issue -- I originally had it where the max start date = end date, but that meant if we wanted to adjust the custom dates to where the new start date > end date, we would have to clear the whole thing in order to make those changes (this wasn't ideal).

angelathe: @BobanL Just added, lmk how it feels if you play around with the UI! It does occasionally require "clearing" in order to reach certain dates depending on its previous value, but I don't think it's too bad

mcmcgrath13:
> 1. Currently getting a console warning that's making me hesitant about the current implementation: **Cannot update a component (`Router`) while rendering a different component (`FilterByDate`)`**

I'll try to pull this down and take a look this afternoon, but I'm not shocked our hacky solution isn't appeasing all the things

> 2. Seems like URL validation is an overall issue we have to address (also for non-custom date ranges and filter by conditions) -> therefore could be its own ticket (worried that the scope of this ticket keeps growing)

Yeah, let's defer, live on the happy path and think about how this should be done for reals.

> 3. I tried for a while to make `returnParamDates` be useful in both `page` and in `resetFilterDate` in the UI component, but it really felt like it wasn't useful for the UI component because A) For UI component, we only want returned dates if the query is "custom", but B) for the page, we always want dates returned. If you have thoughts about re-restructuring this, lmk

I can take a closer look this afternoon

angelathe:
Note: We removed the code to sync the URL parameters with the actual filter functions -- for now, just assuming happy path. The URL validation and sync process will be brought out into its own ticket.

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -271,67 +286,108 @@ const FilterReportableConditions = () => {
...
 
   // on mount, make sure the filters match the search params
   useEffect(resetFilterDate, []);
 
+  const submitHandler = () => {
+    if (filterDateOption === CustomDateRangeOption) {
+      if (!endDate) {
+        setEndDate(today);
+      }
+      const datesParam = `${startDate}|${endDate || today}`;
```
mcmcgrath13, line 322:
```suggestion
      const actualEndDate = endDate || today;
      setEndDate(actualEndDate);
      const datesParam = `${startDate}|${actualEndDate}`;
```
maaaaybe this is a bit better for readability/maintainability? but it's marginal, so take or leave as you see fit

containers/ecr-viewer/src/app/components/Filters.tsx
```diff
@@ -271,67 +286,108 @@ const FilterReportableConditions = () => {
...
-              onChange={handleDateOptionChange}
-              checked={filterDateOption === option}
+        <RadioDateOptions
+          groupName="filter-date"
+          optionsMap={dateRangeLabels}
+          onChange={setFilterDateOption}
+          currentOption={filterDateOption}
+          classNames="padding-bottom-1"
+        />
+        <div className="border-top-1px border-base-lighter margin-x-105 padding-bottom-1"></div>
```
mcmcgrath13, line 366: did you figure out why the `y-1` wasn't equivalent to the two bottoms styling wise?

containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -414,31 +418,31 @@ const generateFilterConditionsStatementSqlServer = (
 
 /**
  * A custom type format for statement filtering by date range
- * @param filterDate - Date range to filter on
+ * @param filterDates - Date range to filter on
  * @returns custom type format object for use by pg-promise
  */
 export const generateFilterDateStatementPostgres = (
-  filterDate: DateRangePeriod,
+  filterDates: DateRangePeriod,
```
mcmcgrath13, line 425:
```suggestion
  {startDate, endDate}: DateRangePeriod,
```

containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -414,31 +418,31 @@ const generateFilterConditionsStatementSqlServer = (
...
     const { pgPromise } = getDB();
 
     return [
       pgPromise.as.format("ed.date_created >= $[startDate]", {
-        startDate: filterDate.startDate,
+        startDate: filterDates.startDate,
       }),
       pgPromise.as.format("ed.date_created <= $[endDate]", {
-        endDate: filterDate.endDate,
+        endDate: filterDates.endDate,
```
mcmcgrath13, line 436:
```suggestion
        endDate,
```

containers/ecr-viewer/src/app/services/listEcrDataService.ts
```diff
@@ -414,31 +418,31 @@ const generateFilterConditionsStatementSqlServer = (
...
       }),
     ].join(" AND ");
   },
 });
 
-const generateFilterDateStatementSqlServer = (filterDate: DateRangePeriod) => {
+const generateFilterDateStatementSqlServer = (filterDates: DateRangePeriod) => {
```
mcmcgrath13, line 442: consider similar ⛳ here

containers/ecr-viewer/src/app/tests/components/Filters.test.tsx
```diff
@@ -455,6 +455,147 @@ describe("Filter by Date Component", () => {
...
+      }),
+    );
+    const radio = screen.getByRole("radio", {
+      name: "Custom date range",
+    });
+    fireEvent.click(radio);
+
+    expect(container).toMatchSnapshot();
+
+    jest.restoreAllMocks();
```
mcmcgrath13, line 484: should this be in an after each hook instead?

containers/ecr-viewer/src/app/tests/components/Filters.test.tsx
```diff
@@ -455,6 +455,147 @@ describe("Filter by Date Component", () => {
...
+  });
+  it("Display start and end date fields when 'Custom date range' is selected", async () => {
+    const mockPush = jest.fn();
+    (useRouter as jest.Mock).mockReturnValue({ push: mockPush });
```
mcmcgrath13, line 488: it doesn't look like the `mockPush` is used anywhere in the test - should it be? or could we just replace it with inline `jest.fn()`?

mcmcgrath13 (approved):
a few minor questions/comments, but overall looks good to me!

PR #3127, opened 2025-01-14
fix: radio buttons

mcmcgrath13:
suggested changes to simplify radio options


angelathe:
Mary, thanks so much for this! I see what you mean now about letting the component deal with the extraction. Will work on merging this into my branch

angelathe:
also sorry Mary, had to revert the className changes because styling ended up being wonky. Will try separately to figure out a way to eliminate that need.

angelathe (approved):

PR #3128, opened 2025-01-14
feat: allow seed data directory to be specified by env var

### LINKED ISSUE #2689: Accept a folder path for the convert seed data script
## What needs to be done

- Add a new parameter to convert seed data to accept a folder path and only convert that data
- Update E2E tests to limit number of files converted

## Why it needs to be done

- In order to run convert-seed-data efficiently in a CI environment and locally for testing specific files.

## Acceptance Criteria

- Convert seed data accepts a new parameter for a folder path
- Convert seed data only runs on the folders within that parameter
- E2E tests run on a limited number of files

mcmcgrath13:
# PULL REQUEST

## Summary

Add an environment variable that controls which directories to read seed data from when set. Defaults to `LA`.

Create a folder of `e2e` ecrs that is a small subset of the `LA` directory and update the e2e tests to use this.

## Related Issue

Fixes #2689

## Acceptance Criteria

- Convert seed data accepts a new parameter for a folder path
- Convert seed data only runs on the folders within that parameter
- E2E tests run on a limited number of files


akasper (commented):
I compared this with [Boban's PR](https://github.com/CDCgov/phdi/actions/runs/12757460614/job/35590549804?pr=3125). In that case it took 9m48s to run seed data conversion, now down to 4m28s. So at least practically speaking for one spot check there was a huge speed up.

containers/ecr-viewer/e2e/ecr-viewer.spec.ts
```diff
@@ -21,21 +21,13 @@ test("fully expanded should not have any automatically detectable accessibility
...
-    .getByTestId("accordionButton_stool-pathogens-naat-parasite")
-    .all();
-  for (let accordionButton of accordionButtons) {
-    await accordionButton.scrollIntoViewIfNeeded();
-    await accordionButton.click();
-  }
-  const viewCommentButtons = await page.getByText("View comment").all();
+  const viewCommentButtons = await page.getByTestId("comment-button").all();
   for (let viewCommentButton of viewCommentButtons) {
+    console.log({ viewCommentButton });
```
austin-hall-skylight, line 30: Let's get rid of this.

containers/ecr-viewer/seed-scripts/create-seed-data.py
```diff
@@ -30,7 +30,8 @@ def _process_files(args):
     :return: A list of fhir bundles
     """
     print("Converting files...")
-    subfolders = ["LA"]
+    subfolders_raw = os.getenv("SEED_DATA_DIRECTORIES", "LA")
```
austin-hall-skylight, line 33: I think defaulting SEED_DATA_DIRECTORIES to LA in the docker compose is good enough. It gives us the data set we want if we don't care about overriding the env var but it also lets us skip LA if we want to.

austin-hall-skylight (changes requested):

containers/ecr-viewer/e2e/ecr-viewer.spec.ts
mcmcgrath13: whooops - my bad

austin-hall-skylight (approved):

akasper (approved):

Issue #3077, opened 2024-12-18
Access database directly


JNygaard-Skylight:
## What needs to be done

To retrieve the FHIR bundle to be displayed in the `view-data` page of the eCR-Viewer we make an API call. However if we separated the code around handling the API request and response from the code that queries our databases there is no reason why the app could not directly call our data retrieval call directly.  

## Why it needs to be done

This would remove the latency of making an API call, and simplify some of the code.

## Acceptance Criteria

- the `view-data` page can retrieve the FHIR bundle without making an API query.

PR #3130, opened 2025-01-15
refactor: move dispatching logic into fhir-data-service, dry out code

mcmcgrath13:
# PULL REQUEST

## Summary

Spin off/preparation PR from #3077. Refactor the fhir data service so the logic to do dispatching on source is in the service and not the route. Also simplify those db_getter functions so that they only do db getting and don't need to worry about mappings or wrangling params.


containers/ecr-viewer/src/app/api/fhir-data/fhir-data-service.ts
```diff
@@ -1,57 +1,100 @@
...
+    res = await get_s3(ecr_id);
+  } else if (process.env.SOURCE === AZURE_SOURCE) {
+    res = await get_azure(ecr_id);
+  } else if (process.env.SOURCE === POSTGRES_SOURCE) {
+    res = await get_postgres(ecr_id);
+  } else {
+    res = { payload: { message: "Invalid source" }, status: 500 };
+  }
+  const { status, payload } = res;
+  console.log({ res });
```
mcmcgrath13, line 42:
```suggestion
```

BobanL (approved):
👍

PR #3131, opened 2025-01-15
fix: remove ecr viewer metrics

mcmcgrath13:
# PULL REQUEST

## Summary

Spinoff/prep for #3077. Per discussion in standup, remove the metrics from the ecr viewer page. It isn't being used and also wasn't working correctly. This was the only place it was used, so also remove the helper and tests.


angelathe (approved):
LGTM!

PR #3137, opened 2025-01-16
chore: remove create and read fhir bundle from postgres

### LINKED ISSUE #2406: Copy of Setup S3 for local development
## What needs to be done
 - Setup S3 localstack in docker-compose.yml
 - Update create seed data to save to S3 bucket
 
## Why it needs to be done
- Since all potential partners will be using some sort of blob storage to store the fhir bundles we should be able to mock that experience locally
- Developers need a local development environment to mock an S3 bucket

## Acceptance Criteria
- [ ]   S3 bucket is running when running `npm run local-dev`
- [ ]   FHIR data gets saved to S3 bucket
- [ ]   Environment variables values are documented and set to connect to S3 bucket

## Technical Notes

- https://docs.localstack.cloud/user-guide/aws/s3/
- https://docs.localstack.cloud/user-guide/integrations/sdks/javascript/

BobanL:
# PULL REQUEST

## Summary

- Remove data_source variable from core.sql
- Remove postgres from get fhir-data and `/save-fhir-data`
- Orchestration tests save to azure now

## Related Issue

Fixes #2406


containers/ecr-viewer/sql/core.sql
```diff
@@ -3,7 +3,6 @@ CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
 CREATE TABLE ecr_data (
   eICR_ID VARCHAR(200) PRIMARY KEY,
   set_id VARCHAR(255),
-  data_source VARCHAR(2), -- S3 or DB
```
mcmcgrath13, line 6: not really a question on this PR, but more generally. When we make a schema change like this, what's the migration plan for users?

containers/ecr-viewer/src/app/api/save-fhir-data/save-fhir-data-service.ts
```diff
@@ -492,13 +451,12 @@ export const saveMetadataToPostgres = async (
...
       // Insert main ECR metadata
       const saveToEcrData = new PQ({
-        text: "INSERT INTO ecr_data (eICR_ID, patient_name_last, patient_name_first, patient_birth_date, data_source, report_date, set_id) VALUES ($1, $2, $3, $4, $5, $6, $7)",
+        text: "INSERT INTO ecr_data (eICR_ID, patient_name_last, patient_name_first, patient_birth_date, report_date, set_id) VALUES ($1, $2, $3, $4, $5, $6)",
         values: [
           ecrId,
           metadata.last_name,
           metadata.first_name,
           metadata.birth_date,
-          "DB",
```
mcmcgrath13, line 501: I'm trying to follow along on what this was doing in the first place. Was this correct? Couldn't we have saved the data to an S3 bucket, but saved the metadata to postgres?

mcmcgrath13: also, why don't I see a similar change in the sql server code?

BobanL: I believe this was supposed to define where to find the FHIR_DATA for the ecr-viewer (s3 or azure blob), but we only have ever supported fetching the data from a single source. 

For your second question, this is only affects postgres because the CORE schema only supports postgres.

@austin-hall-skylight, do you have any thoughts about keeping the `data_source` variable?

PR #3141, opened 2025-01-16
fix: update path to stamped immunizations

BobanL:
# PULL REQUEST

## Summary

- Update path to stamped immunizations to account for change to trigger code reference service #3093


mcmcgrath13 (approved):
🥇

PR #3123, opened 2025-01-09
fix: try to make ecr viewer page ssr

mcmcgrath13:
# PULL REQUEST

## Summary

What changes are being proposed?

## Related Issue

Fixes #

## Acceptance Criteria

Please copy the acceptance criteria from your ticket and paste it here for your reviewer(s)

## Additional Information

Anything else the review team should know?

## Checklist

- [ ] If this code affects the other scrum team, have they been notified? (In Slack, as reviewers, etc.)

[//]: # "PR title: Remember to name your PR descriptively!"

PR #3143, opened 2025-01-17
refactor: tables components

mcmcgrath13:
# PULL REQUEST

## Summary

Another spinoff from #3123 (something was funky with using client and `BuildHeaders` which turned into a bit of a 🧹 🐰 🕳️ ). 

Refactor/rename the `BuildTable` ecosystem to be more declarative and react-y. `BuildTable => BaseTable`. Get rid of `BuildHeaders` and just pass the column specs down. Move `tableRows` prop to `children` so the jsx is more consolidated.


BobanL (approved):
🧹

PR #3147, opened 2025-01-17
Add version id to core schema

### LINKED ISSUE #3146: Add version ID to core schema
## What needs to be done

Add version ID field to core schema. Add materialized view of ecr_version for core schema

## Why it needs to be done

Version ID is how we will group eCRs, but it isn’t included in the core schema today.

## Acceptance Criteria

- [ ] When seeding the database using the core schema, the set ID field is populated

## Technical Notes

_Owned by eng team - write out any important technical considerations such as: architecture plan/notes, data dependencies, testing plan, etc. Also link any tickets that are dependent on or related to this task_

lina-roth:
# PULL REQUEST

## Summary

Add version ID field to core schema
## Related Issue

Fixes #3146 

## Acceptance Criteria

 When seeding the database using the core schema, the set ID field is populated

## Additional Information

## Checklist


containers/ecr-viewer/sql/version_view.sql
```diff
@@ -0,0 +1,2 @@
+create view ecr_version as
```
mcmcgrath13, line 1: given all this data is in the core schema, I don't think we need the view. What do you think?

lina-roth: I'm fine to remove it. It wasn't clarified in the ticket what the purpose of creating the materialized view was for, so I was just following the AC. Seemed like it could have been for an optimized query only pulling back necessary data

mcmcgrath13: let's remove it and add the optimization if it turns out performance is an issue, but my intuition is given all the filtering this already has to play with, that would be a deeper rabbit hole

mcmcgrath13 (approved):
thanks!

